PR #3, opened 2020-02-11
Set up Travis CI

binh-dam-ibigroup:


.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+script:
+- mvn package
+after_success:
+- bash <(curl -s https://codecov.io/bash) -s target-unit-test-results -F unit_tests
+- |
+  semantic-release --prepare @conveyal/maven-semantic-release --publish @semantic-release/github,@conveyal/maven-semantic-release --verify-conditions @semantic-release/github,@conveyal/maven-semantic-release --verify-release @conveyal/maven-semantic-release --use-conveyal-workflow --dev-branch=dev --skip-maven-deploy
+  if [[ "$TRAVIS_BRANCH" = "master" ]]; then
+    bash <(curl -s https://codecov.io/bash) -C "$(git rev-parse HEAD)" -s target-unit-test-results -F unit_tests
+    bash <(curl -s https://codecov.io/bash) -C "$(git rev-parse HEAD^)" -s target-unit-test-results -F unit_tests
+  fi
```
evansiroky, line 35: This code seems to be copy-pasted from datatools-server which contains tests suites of either unit tests or e2e tests. Once this repo has e2e tests we can add this back in, but for now, I think it would be more appropriate to copy-paste the codecov scripts that gtfs-lib uses.

evansiroky (changes requested):
See comment about scripts that upload coverage reports to codecov.

.travis.yml
binh-dam-ibigroup: Yep let's remove the test stuff for right now.

landonreed (approved):

evansiroky (approved):

PR #4, opened 2020-02-13
Add persistence

landonreed:
Adds mongo driver, basic persistence classes, and generic controller class (with a basic example for the `User` model).

Note: example endpoints are available in the Postman collection.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
+++ b/src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
@@ -25,42 +25,22 @@ public class Persistence {
 
     private static MongoClient mongo;
     private static MongoDatabase mongoDatabase;
-    private static CodecRegistry pojoCodecRegistry;
-
     // One abstracted Mongo collection for each class of persisted objects
     public static TypedPersistence<User> users;
 
     public static void initialize () {
-
-//        PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder()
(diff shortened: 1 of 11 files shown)


src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static void initialize () {
+
+//        PojoCodecProvider pojoCodecProvider = PojoCodecProvider.builder()
+//            .register("com.conveyal.datatools.manager.jobs")
+//            .register("com.conveyal.datatools.manager.models")
+//            .register("com.conveyal.gtfs.loader")
+//            .register("com.conveyal.gtfs.validator")
+//            .automatic(true)
+//            .build();
```
evansiroky: Since this is all copy-pasted anyways from elsewhere, I think we should leave these comments out for now.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -0,0 +1,5 @@
+package org.opentripplanner.middleware.utils;
+
+public class JsonUtils {
+    // TODO add gson.toJson util
```
evansiroky: Can you provide a link to an issue that describes what this means?

evansiroky (approved):
Good enough, but I have some comments about the comments.

evansiroky (approved):

binh-dam-ibigroup (approved):

PR #5, opened 2020-02-20
Basic JMeter profiling

binh-dam-ibigroup:
This PR covers the following:

* Script in Java to generate OTP queries for combinations of preset locations
* Basic webpage with links to the generated queries in OTP MOD UI
* Basic generated JMeter test case with select combinations of the generated queries
* Bash scripts to install JMeter on a remote server (e.g. middleware server)


landonreed:
@binh-dam-ibigroup, could you change the base branch to `dev` to follow the typical branching pattern? I just changed the default branch in the repo's settings so that future PRs will do this automatically.

binh-dam-ibigroup:
> @binh-dam-ibigroup, could you change the base branch to `dev` to follow the typical branching pattern? I just changed the default branch in the repo's settings so that future PRs will do this automatically.

Yes... thanks for catching that!

jmeter/2-location-pairs.jmx
```diff
@@ -0,0 +1,171 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
evansiroky: This and the other file should be combined into 1.

src/main/java/org/opentripplanner/middleware/OtpJMeterRequestGenerator.java
```diff
@@ -0,0 +1,184 @@
...
+        // Just print out a list of OTP requests from the places in the const file.
+        // Combine the following modes: Walk, Transit, Bike, Bikeshare
+
+        //System.out.println("<h1>JMeter Requests</h1>");
+        //System.out.println("<p>Structure: One thread group = One row in table above, each group runs the sequence of OD pairs, each pair runs the modes in parallel.</p>");
+        //System.out.println("<pre>");
+
+
+        // JMeter stuff
+        System.out.println("<?xml version=\"1.0\" encoding=\"UTF-8\"?>");
```
evansiroky: Absolutely not! Jmeter test plans should not be programmatically generated like this.

src/main/java/org/opentripplanner/middleware/OtpJMeterRequestGenerator.java
```diff
@@ -0,0 +1,184 @@
...
+        System.out.println("<stringProp name=\"TestPlan.user_define_classpath\"></stringProp>");
+        System.out.println("</TestPlan>");
+        System.out.println("<hashTree>");
+
+        // Start thread group here - one per row.
+
+        for (int i = 0; i < locations.size(); i++) {
+            Map.Entry<String, Double[]> e1 = locations.get(i);
+
+            System.out.println("<ThreadGroup guiclass=\"ThreadGroupGui\" testclass=\"ThreadGroup\" testname=\"User " + (i+1) + " at " + e1.getKey() + "\" enabled=\"true\">");
```
evansiroky: I don't understand the point of creating ThreadGroups like this. This should all just be one threadgroup that can be executed multiple instances by Jmeter.

src/main/java/org/opentripplanner/middleware/OtpRequestGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+
+                    System.out.println("</td>");
+                    count++;
+                }
+            }
+        }
+
+        System.out.println("</table>");
+        System.out.println("<p>Total combinations: " + count + "</p>");
+        System.out.println("</body><html>");
```
evansiroky: Don't understand the point of generating this file.

evansiroky (changes requested):
This is definitely not what I was thinking for creating the Jmeter test plan. I was thinking something super simple with one Threadgroup with a loop that creates URLs that are parsed by Jmeter and then executed in a single sampler. This is way too complicated and doesn't take advantage of custom processing that can be done completely within Jmeter.

binh-dam-ibigroup:
With latest commits, we have a JMeter test case that generates origin/destination pairs and sends parallel HTTP requests for different modes for each pair.

jmeter/run-gui.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+source jmeter-version.sh
+
+apache-jmeter-$JMETER_VER/bin/jmeter.sh -t 2-location-pairs.jmx
```
evansiroky: Please update to correct jmx file.

jmeter/testbed.jmx
```diff
@@ -0,0 +1,551 @@
...
+          <stringProp name="CounterConfig.start">1</stringProp>
+          <stringProp name="CounterConfig.end"></stringProp>
+          <stringProp name="CounterConfig.incr">1</stringProp>
+          <stringProp name="CounterConfig.name">globalCounter</stringProp>
+          <stringProp name="CounterConfig.format"></stringProp>
+          <boolProp name="CounterConfig.per_user">false</boolProp>
+        </CounterConfig>
+        <hashTree/>
+        <CSVDataSet guiclass="TestBeanGUI" testclass="CSVDataSet" testname="User Origins (CSV Data Set Config)" enabled="true">
+          <stringProp name="filename">/Users/binhdam/git/otp-middleware/jmeter/locations.csv</stringProp>
```
evansiroky: This won't work except on your computer. This should be a relative import, or passed in as a CLI argument. To see an example of how to do this, check out [this JMX snippet in datatools-server](https://github.com/ibi-group/datatools-server/blob/134a3234beab9eb25895e8e143ad5484a5161344/jmeter/test-script.jmx#L42). Also, see the [run-tests.sh](https://github.com/ibi-group/datatools-server/blob/dev/jmeter/run-tests.sh) script for assisting with collecting CLI arguments and passing them to jmeter.

jmeter/testbed.jmx
```diff
@@ -0,0 +1,551 @@
...
+        <LoopController guiclass="LoopControlPanel" testclass="LoopController" testname="LoopController" enabled="true">
+          <boolProp name="LoopController.continue_forever">true</boolProp>
+          <stringProp name="LoopController.loops">21</stringProp>
+        </LoopController>
+        <hashTree>
+          <CSVDataSet guiclass="TestBeanGUI" testclass="CSVDataSet" testname="Destinations (CSV Data Set Config)" enabled="true">
+            <stringProp name="TestPlan.comments">Locations for origin/destination pairs.</stringProp>
+            <stringProp name="delimiter">,</stringProp>
+            <stringProp name="fileEncoding">UTF-8</stringProp>
+            <stringProp name="filename">/Users/binhdam/git/otp-middleware/jmeter/locations.csv</stringProp>
```
evansiroky: Same comment as [here](https://github.com/ibi-group/otp-middleware/pull/5/files#r387362436).

evansiroky (changes requested):
This is much better than the previous iteration and outputs exactly the statistics that we need. This should be good to go after updating file locations as I mentioned.

binh-dam-ibigroup:
Oh... good point on the file locations.

jmeter/testbed.jmx
binh-dam-ibigroup: I moved this parameter to be a batch file parameter. In JMeter UI it defaults to that CSV file.

jmeter/run-gui.sh
binh-dam-ibigroup: Corrected.

evansiroky (approved):

jmeter/upload-jmeter-scripts.sh
```diff
@@ -0,0 +1,28 @@
...
+ssh $1 'rm -rf ~/jmeter'
+
+ssh $1 'mkdir -p ~/jmeter'
+scp *.jmx $host:~/jmeter
+scp install-jmeter.sh $host:~/jmeter
+scp jmeter-version.sh $host:~/jmeter
+scp run-tests.sh $host:~/jmeter
+
+ssh $1 'mkdir -p ~/jmeter/fixtures'
+scp fixtures/* $host:~/jmeter/fixtures
```
landonreed, line 25: Instead of all of these individual scp commands, can we just run a recursive command to copy all of the files here (perhaps we need to exclude certain files). I think it would be helpful to have all of the source files including the README on the remote machine.

landonreed (approved):
Looks good, just one comment about combining the multiple scp commands into a single one.

jmeter/upload-jmeter-scripts.sh
binh-dam-ibigroup: Yeah we can upload all *.sh files at once.

binh-dam-ibigroup: Actually, no, `upload-jmeter-scripts.sh` is not needed remotely, and neither is `run-gui.sh`. So at this time, there is no syntax gains for writing a pattern that excludes certain files.

PR #7, opened 2020-02-21
Add tests for persistence methods

landonreed:
- Adds basic testing framework (junit)
- Adds persistence tests for user
- Adds mongo to Travis CI config
- Deletes (`.iml`) file (Intellij project file) (note: you'll need to `Build > Rebuild Project` after pulling these changes into your local repo to regenerate the file)

NOTE: This does pull in changes from #10 and updates the README to include run instructions. As we develop the application's documentation, we might want to eventually move some of that to a `docs` directory.


pom.xml
```diff
@@ -132,7 +152,13 @@
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
         </dependency>
-
+        <!-- Used for unit testing -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
```
evansiroky: Should we start out using JUnit 5? It's the latest and greatest.

evansiroky (commented):
Just one comment about JUnit 5.

binh-dam-ibigroup (dismissed):
I agree with Evan, let's move to JUnit 5

src/test/java/org.opentripplanner.middleware/OtpMiddlewareTest.java
```diff
@@ -0,0 +1,31 @@
...
+ * This abstract class contains is used to start a test instance of datatools-server that other tests can use to perform
+ * various tests.
+ *
+ * A majority of the test involves setting up the proper config files for performing tests - especially the e2e tests.
+ */
+public abstract class OtpMiddlewareTest {
+    private static final Logger LOG = LoggerFactory.getLogger(OtpMiddlewareTest.class);
+    private static boolean setUpIsDone = false;
+
+    @BeforeAll
```
binh-dam-ibigroup: Do we need this annotation in this file since it has no tests?

binh-dam-ibigroup (changes requested):
Just one comment then it should be good.

src/test/java/org.opentripplanner.middleware/OtpMiddlewareTest.java
landonreed: Yes, because we want it to run before any test that is run in a class that implements this abstract class. However, I think we should remove it from `PersistenceTest` because all it's doing is calling this setup method. I'll remove the `PersistenceTest#setup` method.

binh-dam-ibigroup (approved):

evansiroky (approved):

PR #8, opened 2020-02-21
fix(Const): Update OTP URLs.

binh-dam-ibigroup:
This PR removes TriMet, uses FDOT instead for the request test endpoints.


evansiroky (approved):

landonreed (approved):
Are these strings  used by the endpoint that Jmeter hits? This is ok, but we might want to consider only registering this endpoint if running in TEST mode (or something like that we configure in the config file).

binh-dam-ibigroup:
Eventually, we will remove these starter endpoints! 

> Are these strings used by the endpoint that Jmeter hits? This is ok, but we might want to consider only registering this endpoint if running in TEST mode (or something like that we configure in the config file).

PR #9, opened 2020-02-21
docs(swagger): Add starter swagger file for API docs.

binh-dam-ibigroup:


evansiroky (changes requested):
Can we use yaml?

binh-dam-ibigroup:
> Can we use yaml?
It is possible according to the specification.
swagger-ui takes in a json file hence the choice to do a json file at this time.

evansiroky:
This thing https://swagger.io/tools/swagger-ui/ ? It can do yaml just fine.

binh-dam-ibigroup:
> This thing https://swagger.io/tools/swagger-ui/ ? It can do yaml just fine.
The UI only handles json per https://github.com/swagger-api/swagger-ui/issues/595, so I added a Travis step to convert to json for displaying with swagger-ui.

pom.xml
```diff
@@ -96,6 +96,28 @@
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
             </plugin>
+            <plugin>
```
landonreed, line 99: Could you give this plugin a comment?

pom.xml
```diff
@@ -96,6 +96,28 @@
...
+                <groupId>org.openapitools</groupId>
+                <artifactId>openapi-generator-maven-plugin</artifactId>
+                <!-- RELEASE_VERSION -->
+                <version>4.2.3</version>
+                <!-- /RELEASE_VERSION -->
```
landonreed: I'm not sure these two comments are doing much for us.

pom.xml
```diff
@@ -172,5 +194,5 @@
                 </exclusion>
             </exclusions>
         </dependency -->
-    </dependencies>
+  </dependencies>
```
landonreed: This white space doesn't seem to line up with the opening tag.

.travis.yml
```diff
@@ -41,6 +41,7 @@ before_deploy:
 - ls target/*.jar
 - cp target/otp-middleware-*.jar deploy/
 - cp target/otp-middleware-*.jar "deploy/otp-middleware-latest-$BRANCH_CLEAN.jar"
+- cp target/generated-sources/openapi/openapi.json "deploy/otp-middleware-api-latest-$BRANCH_CLEAN.json"
```
landonreed: @binh-dam-ibigroup, I would prefer that in addition to identifying deployment artifacts by the branch name, that we also publish artifacts using the `git describe` à la datatools-server (it uses a maven plugin for this): 

https://github.com/ibi-group/datatools-server/blob/423ed4d5859f0184f4587080c83f600f4b4481de/pom.xml#L68-L71

May make more sense to just use git describe in Travis rather than using the maven plugin, so that we can re-use the variable for  other artifacts as well (e.g., these API docs). By the way, is anything done with this JSON file once deployed? Or does it just sit in the S3 bucket?

src/main/resources/api.yaml
```diff
@@ -0,0 +1,174 @@
+swagger: '2.0'
+info:
+  description: >-
+    This documents the OTP Middleware endpoints. For this sample, you can use
+    the api key `special-key` to test the authorization filters.
+  version: 0.0.0
```
landonreed, line 6: Is this file being generated manually? Either way, it would be good to have a comment at the top of this file describing its purpose (e.g., how it should be maintained, where it is used, etc.). Also, I worry about the version becoming out of sync with what semantic-release labels things. Although, perhaps the API version and the software version don't necessarily match up?

src/main/resources/api.yaml
```diff
@@ -0,0 +1,174 @@
...
+      username:
+        type: string
+      firstName:
+        type: string
+      lastName:
+        type: string
+      email:
+        type: string
+      password:
+        type: string
```
landonreed, line 172: We're definitely not going to be storing or ever touching passwords. This User def seems somewhat made up (which is OK for now), but we should add a `TODO` comment stating this.

src/main/resources/api.yaml
```diff
@@ -0,0 +1,174 @@
...
+          in: path
+          description: The name that needs to be deleted
+          required: true
+          type: string
+      responses:
+        '400':
+          description: Invalid username supplied
+        '404':
+          description: User not found
+  /user:
```
landonreed, line 93: I think there needs to be a `get` method here for `getAll`?

.travis.yml
binh-dam-ibigroup: Oh good point, I'll append the git commit hash after the branch to identify each build.
Deployment beyond the S3 bucket should be more selective, like on a merge to the dev or  master branch?

src/main/resources/api.yaml
```diff
@@ -0,0 +1,174 @@
...
+            $ref: '#/definitions/User'
+        '400':
+          description: Invalid username supplied
+        '404':
+          description: User not found
+    put:
+      tags:
+        - user
+      summary: Updates user info
+      description: This can only be done by the logged in user.
```
landonreed, line 51: A lot of these values seem more like placeholders than anything. I think we should go through and either remove much of them or add TODO comments.

src/main/resources/api.yaml
binh-dam-ibigroup: This file is generated manually for now...

src/main/resources/api.yaml
binh-dam-ibigroup: Yes, that is just a starter, copy of the example in swagger-ui to set up the documentation processes. Feel free to make corrections.

src/main/resources/api.yaml
```diff
@@ -0,0 +1,174 @@
+swagger: '2.0'
```
landonreed, line 1: It doesn't feel 100% right that this file is buried in `src/main/resources`. That seems to imply that it is somehow used by the source code. Have you found any other projects that keep swagger/OpenAPI files in the same repo as the code? I'm wondering if there's a standard way of doing this.

landonreed (changes requested):
Do you know if there's a standard for where these API docs live for various projects? In general, my comments are just about adding more comments about various things.

src/main/resources/api.yaml
binh-dam-ibigroup: Still exploring... Ideally, we'd want to generate the swagger file from the code, or generate the endpoint and model code from swagger.

binh-dam-ibigroup:
Closing. Superseded by #12 .

PR #14, opened 2020-03-30
refactor(user): add sub-types for users

landonreed:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR adds some user sub types to account for admin users, and "API users" (i.e., developer/3rd party accounts). It leaves the generic User for handling OTP users (which API user and admin users both automatically get).

PR #13, opened 2020-03-18
Add auth0

landonreed:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

Depends on #12. This PR adds authentication via Auth0 (much of this code is taken from ibi-group/datatools-server). It also adds a UserController class that uses hooks into ApiController in order to inject some validation of users and create an actual Auth0 user when creating users in the otp-middleware application. Note: there are updates to env.yml for Auth0 settings.

Works with ibi-group/otp-admin-ui#1

PR #15, opened 2020-05-04
feat(User): Basic user preferences, GET User by attribute.

binh-dam-ibigroup:
### Checklist

### Description

This PR adds the following:
* An api route to get entities by attribute, used in the new MOD UI to retrieve user data using an auth0 id. (The Mongo ID is not stored or communicated by Auth0).
* User fields and types related to creating user accounts on middleware.


src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -90,6 +89,19 @@ public void bind(final SparkSwagger restApi) {
                     this::getOne, JsonUtils::toJson
             )
 
+            // Get one entity by field.
+            .get(path(ROOT_ROUTE + FIND_PATH)
+                    .withDescription("Returns a '" + classToLowercase + "' entity whose field has the specified value, or 404 if not found.")
+                    .withPathParam().withName("attribute").withDescription("The attribute the entity to search.").and()
```
br648: "The attribute the entity to search." This doesn't make sense.

src/main/java/org/opentripplanner/middleware/models/User.java
```diff
@@ -11,16 +13,54 @@
...
+
+    /**
+     * Notification preference for this user
+     * ("none", "email", "sms").
+     * TODO: use enum (https://jira.mongodb.org/browse/JAVA-268)
+     */
+    public String notificationChannel;
+
+    /** Phone number for SMS notifications. */
+    public String phoneNumber;
```
br648: Does the User class now contain all required fields? If so, perhaps remove the TODO (on line 16).

src/main/java/org/opentripplanner/middleware/models/UserNotificationChannel.java
```diff
@@ -0,0 +1,9 @@
+package org.opentripplanner.middleware.models;
+
+/**
+ * Describes the means to notify a user:
+ * none (no notification), via email, via sms).
+ */
+public enum UserNotificationChannel {
```
br648: If we decide that enums can't be saved to Mongo this will need to be changed.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -122,6 +122,16 @@ public T getById (String id) {
         return mongoCollection.find(eq(id)).first();
     }
 
+    /**
+     * Retrieves the first document that matches an field value, or null if there is no match.
+     * @param field The field to scan.
+     * @param value The value to find.
+     * @return A document whose specified field has the specified value.
+     */
```
br648: This is the only method that has this amount of detail in the comment (@param and @return). I'm not sure what the policy is on this, but probably easier to remove to be inline with the other comments?

br648 (commented):
Some minor comments. All good otherwise.

binh-dam-ibigroup:
Note: last commit will convert `notificationChannel` into `notificationChannels` to allow notifications with both email and sms.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -122,6 +122,16 @@ public T getById (String id) {
...
+    public T getByField (String field, String value) {
```
evansiroky: I would like this to be named `getFirstByFieldValue` to be clear that it is returning a single record and getting the first result whatever that may be.

src/main/java/org/opentripplanner/middleware/models/UserLocation.java
```diff
@@ -0,0 +1,18 @@
...
+    private static final long serialVersionUID = 1L;
+
+    /** The address of this location. */
+    public String address;
+
+    /** The icon name for this location (e.g. a FontAwesome icon name). */
+    public String icon;
+
+    /** The type/purpose of this location (free form) (e.g. home, work, shopping...) */
+    public String type;
```
evansiroky, line 25: Needs `lat`, `lon` added.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -188,6 +209,22 @@ private T getObjectForId(Request req, String id) {
         return object;
     }
 
+    /**
+     * Convenience method for extracting the attribute/field param from the HTTP request.
+     */
+    private T getObjectWithField(Request req, String field, String value) {
```
evansiroky: Would like this renamed to `getFirstObjectByFieldValue` to indicate that it is getting the first record that matches a field value.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -139,6 +151,15 @@ private T getOne(Request req, Response res) {
         return getObjectForId(req, id);
     }
 
+    /**
+     * HTTP endpoint to get one entity specified by a field value.
+     */
+    private T getOneByField(Request req, Response res) {
```
evansiroky: I'd like this renamed to `getFirstByFieldValue` to indicate that this is getting the first record that matches the given field value.

evansiroky (changes requested):
See comments about renaming things to indicate that certain methods are returning the first item matching a field value. I feel this is important because the first result from a search on a field without distinct values could end up returning inconsistent records.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -90,6 +89,19 @@ public void bind(final SparkSwagger restApi) {
                     this::getOne, JsonUtils::toJson
             )
 
+            // Get one entity by field.
+            .get(path(ROOT_ROUTE + FIND_PATH)
```
evansiroky: Each of these get API endpoints don't have any security associated with them. Any requester could just send over an arbitrary Auth0 user id and it would return the user object for another user.

evansiroky (changes requested):
Just discovered a big security risk, so I'm marking this PR as "requesting changes" until all get methods have authorization checks added to ensure data is returned only when proper authorization can be confirmed.

Also, it appears that the deletion methods need proper authorization checks as well.

binh-dam-ibigroup:
Caution: On parent branch `add-sub-users`, a similar "filtered query" was introduced. Need to use it. See https://github.com/ibi-group/otp-middleware/blob/77b9e500c1f308690e41bec6cbd893567aaf3667/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java#L43

binh-dam-ibigroup:
> Just discovered a big security risk, so I'm marking this PR as "requesting changes" until all get methods have authorization checks added to ensure data is returned only when proper authorization can be confirmed.
> 
> Also, it appears that the deletion methods need proper authorization checks as well.

Suggestion: address the quoted comment in #13 or #14. Thoughts? @evansiroky and @landonreed

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
binh-dam-ibigroup: I think this method will become `getOneFiltered` once #14 is merged.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
binh-dam-ibigroup: Method is redundant and I went ahead and removed it.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
binh-dam-ibigroup: Method is redundant. I went ahead and removed it.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -197,29 +198,26 @@ private String deleteOne(Request req, Response res) {
...
-    private T getObjectWithField(Request req, String field, String value) {
-        T object = persistence.getByField(field, value);
+    private T getFirstObjectByFieldValue(Request req, String field, String value) {
+        return checkAndLogIfNull(req, persistence.getOneFiltered(eq(field, value)), field, value);
+    }
+
+    /**
+     * Log if an object was not found for a request.
+     */
+    private T checkAndLogIfNull(Request req, T object, String field, String value) {
```
evansiroky: This is too much abstraction to save just a little bit of repeated code. It makes things less readable, so I think the original way of doing this in `getObjectForId` is preferred.

evansiroky: Or maybe a hybrid approach would be better where you can change this method to be something like:

```java
private void haltIfNull(T object, Request req, String message) {
    if (object == null) {
        logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, message);
    }
}
```

evansiroky (changes requested):
See https://github.com/ibi-group/otp-middleware/pull/15/files#r426110256 and https://github.com/ibi-group/otp-middleware/pull/15/files#r426108133 and https://github.com/ibi-group/otp-middleware/pull/15/files#r428252282.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: @evansiroky This is addressed by #21 in https://github.com/ibi-group/otp-middleware/blob/b827e04b81d9b2fd31a7b558ab9f3b0cd846a473/src/main/java/org/opentripplanner/middleware/auth/Auth0Utils.java#L15 so we *will* be making those checks.

binh-dam-ibigroup: Actually, it is also handled by #13 in https://github.com/ibi-group/otp-middleware/blob/94dcefa65446530ba5854afb442aac7d416c5312/src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java#L38

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -90,6 +92,19 @@ public void bind(final SparkSwagger restApi) {
...
             )
 
+            // Get one entity by field.
+            .get(path(ROOT_ROUTE + FIND_PATH)
+                    .withDescription("Returns a '" + classToLowercase + "' entity whose field has the specified value, or 404 if not found.")
+                    .withPathParam().withName("attribute").withDescription("The name of the attribute of which to extract the value.").and()
+                    .withPathParam().withName("value").withDescription("The desired value for the specified attribute.").and()
+                    // .withResponses(...) // FIXME: not implemented (requires source change).
+                    .withResponseType(clazz),
+                    this::getOneByField, JsonUtils::toJson
```
evansiroky: `getOneByField` is undefined

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
evansiroky: Ok, I'll overlook this for now since #21 has needed checks.

evansiroky (changes requested):
Travis builds failing probably due to undefined method call. Probably good to go once that is fixed.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Thanks for pointing it out!

evansiroky (dismissed):

br648 (dismissed):
Looks good!

br648 (dismissed):
All good and unti test passes.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -274,10 +297,17 @@ private T createOrUpdate(Request req, Response res) {
...
-        String id = req.params("id");
-        if (id == null) {
-            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Must provide id");
+        return getParamFromRequest(req, "id");
+    }
+
+    /**
+     * Get entity attribute value from request.
+     */
+    private String getParamFromRequest(Request req, String paramName) {
```
evansiroky: Can we rename this to `getRequiredParamFromRequest` and add a note that this method will halt the request if the param is not provided?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -58,22 +59,44 @@ public ApiController (String apiPrefix, TypedPersistence<T> persistence, String
...
+     * Either before or after(*) supplemental methods are added, be sure to call the super method to add CRUD operations.
+     *
+     * (*) Note: spark-java will resolve methods in the order they are added to the baseEndPoint parameter.
+     * For instance, if /path and /path/subpath are added in this order, then
+     * a request with /path/subpath will be treated as /path, and the method for /path/subpath will be ignored.
+     * Conversely, if /path/subpath and /path are added in this order, then
+     * a request with /path/subpath will be handled by the method for /path/subpath.
+     * @param baseEndPoint The end point to which to add the methods.
+     */
+    protected void buildEndPoint(ApiEndpoint baseEndPoint) {
```
evansiroky: Please rename `EndPoint` to `Endpoint` to maintain consistency.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -16,10 +23,35 @@
...
 public class OtpUserController extends ApiController<OtpUser> {
+    static final String NO_USER_WITH_AUTH0_ID_MESSAGE = "No user with auth0UserID=%s found.";
+    private static final String TOKEN_PATH = "/fromtoken";
+
     public OtpUserController(String apiPrefix){
         super(apiPrefix, Persistence.otpUsers, "secure/user");
     }
 
+    @Override
+    protected void buildEndPoint(ApiEndpoint baseEndPoint) {
```
evansiroky: Please rename `EndPoint` to `Endpoint` to maintain consistency.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -49,4 +81,18 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
+    /**
+     * HTTP endpoint to get the {@link OtpUser} entity, if it exists, from an {@link Auth0UserProfile} attribute
+     * available from a {@link Request} (this is the case for '/api/secure/' endpoints).
+     */
+    private OtpUser getUserFromRequest(Request req, Response res) {
+        Auth0UserProfile profile = Auth0Connection.getUserFromRequest(req);
+        OtpUser user = profile.otpUser;
+
+        if (user == null) {
+            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, String.format(NO_USER_WITH_AUTH0_ID_MESSAGE, profile.auth0UserId), null);
```
evansiroky, line 98: This should return an HTTP 403 error since it seems that whatever request this is will fail if an authenticated user is not present in the request.

src/test/java/org/opentripplanner/middleware/auth/Auth0UsersTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Checks that token is valid soon after creation and seconds until expiration does diminish.
+     */
+    @Test
+    public void isTokenStale() {
+        // Verify that token is not expired soon after it set.
+        assertFalse(Auth0Users.getCachedToken().isStale());
+        // Verify that seconds until expiration has decreased since first set.
+        assertTrue(Auth0Users.getCachedToken().secondsUntilExpiration() < TOKEN_DURATION_SECONDS);
```
evansiroky: Since this is dependent on the time, I recommend putting in a Thead.sleep or something to be extra sure that some amount of time has elapsed.

evansiroky (changes requested):
See most recent comments.

br648 (dismissed):
Looks good.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: @evansiroky Please describe a situation where this would be the case.

For users who just created an Auth0 login and have an `Auth0UserProfile`,
but have not completed the account setup form yet, the `OtpUser` profile would naturally be 404 not found until they complete the account setup form and you wouldn't want to 403 forbid that user to try to retrieve their account.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -1,10 +1,45 @@
...
+
+    /** Whether the user has consented to terms of use. */
+    public boolean hasConsentedToTerms;
+
+    /**
+     * Whether the email address has been verified.
+     * Some identity services, such as Auth0, manage the email verification process and status,
+     * so this field is only for reference.
+     */
+    public boolean isEmailVerified;
```
landonreed: @binh-dam-ibigroup, is this field populated from the Auth0 field? We're not allowing users to create an account if their email has not already been verified, correct? This suggests to me that this field is no longer required.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -1,10 +1,45 @@
...
+
+    /** Whether the phone number has been verified. */
+    public boolean isPhoneNumberVerified;
```
landonreed, line 18: Please add TODO because we have not yet added code to verify the phone number.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -1,10 +1,45 @@
...
+
+    /**
+     * Notification preference for this user
+     * ("email", "sms", or "none").
+     * NOTE: This could become an enum, e.g. https://jira.mongodb.org/browse/JAVA-268.
```
landonreed: This comment is very cryptic. Please just bump the mongo version if that resolves the enum storage issue.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
evansiroky: Huh, I guess this one could go either way.

evansiroky (dismissed):

landonreed (changes requested):
Looks good, just a few changes.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -16,10 +23,35 @@
...
 public class OtpUserController extends ApiController<OtpUser> {
+    static final String NO_USER_WITH_AUTH0_ID_MESSAGE = "No user with auth0UserID=%s found.";
+    private static final String TOKEN_PATH = "/fromtoken";
+
     public OtpUserController(String apiPrefix){
         super(apiPrefix, Persistence.otpUsers, "secure/user");
     }
 
+    @Override
+    protected void buildEndpoint(ApiEndpoint baseEndPoint) {
```
br648: I think Evan mentioned this, lowercase 'point' so baseEndpoint.

br648 (dismissed):
One very minor item, apart from that all good!

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Fixed. Thanks.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
binh-dam-ibigroup: I will remove. We don't use this field and get that information straight from Auth0.

binh-dam-ibigroup: Same for recent locations.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
binh-dam-ibigroup: Updating the comment. The Mongo driver supports the desired functionality, however there is additional setup needed.

br648 (approved):
Looks good!

evansiroky (approved):

landonreed (approved):
Good stuff. Thanks for the changes.

PR #17, opened 2020-05-12
Trip history

br648:
### Checklist

- [X] The description lists any configuration setting(s) that differ from the default settings

### Description
Covers the process of saving to MongoDB trip requests and summaries of OTP plan requests. The response provided by OTP is sent back to the calling UI as if directly served by OTP so the process of storing trip information is seamless to the end user. Trip requests and summaries are only saved if the user id (MongoDB -> User -> id) is known and a response (good or bad) is recevied from OTP.

The following three parameters have been added to env.yml and env.yml.tmp:

OTP_SERVER: https://fdot-otp-server.ibi-transit.com
OTP_SERVER_PLAN_END_POINT: /otp/routers/default/plan

The trip request query params (which are currently the raw url query params from a request) could be replace with the request parameters returned from OTP. This would have the advantage of storing POJOs instead of raw text. This can be changed if required.


br648:
A test user has to be created in the "otp_middleware" DB and it's id included in requests before trip requests and summaries are saved.

pom.xml
```diff
@@ -124,6 +124,22 @@
         </plugins>
     </build>
     <dependencies>
+<!--        <dependency>-->
+<!--            <groupId>org.opentripplanner</groupId>-->
+<!--            <artifactId>otp</artifactId>-->
+<!--            <version>1.5.0-SNAPSHOT</version>-->
+<!--        </dependency>-->
+        <!-- https://mvnrepository.com/artifact/javax.ws.rs/javax.ws.rs-api -->
+        <dependency>
```
binh-dam-ibigroup, line 134: Please insert a comment explaining the purpose of these libraries.

src/main/java/org/opentripplanner/middleware/BasicOtpDispatcher.java
```diff
@@ -118,4 +121,8 @@ private static void recordResponse(Consumer<String> sendMessageFn, long startTim
         System.out.println(log);
         sendMessageFn.accept(log);
     }
+
+    public void processTripRequest(Request req, Response res) {
```
binh-dam-ibigroup: Is this used anywhere else? If not you can remove.

src/main/java/org/opentripplanner/middleware/otp/core/api/model/error/PlannerError.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/** This API response element represents an error in trip planning. */
+public class PlannerError {
+
+    private static final Logger LOG = LoggerFactory.getLogger(org.opentripplanner.middleware.otp.core.api.model.error.PlannerError.class);
+    private static Map<Class<? extends Exception>, Message> messages;
+//    static {
```
binh-dam-ibigroup: Let's remove this commented block if you are not going to need it soon.

src/main/java/org/opentripplanner/middleware/otp/core/api/model/error/PlannerError.java
```diff
@@ -0,0 +1,119 @@
...
+    public PlannerError(Exception e) {
+        this();
+        message = messages.get(e.getClass());
+        if (message == null) {
+            LOG.error("exception planning trip: ", e);
+            message = Message.SYSTEM_ERROR;
+        }
+        this.setMessage(message);
+//        if (e instanceof VertexNotFoundException)
+//            this.setMissing(((VertexNotFoundException)e).getMissing());
```
binh-dam-ibigroup: Remove

configurations/default/env.yml.tmp
```diff
@@ -1,11 +1,12 @@
-AUTH0_DOMAIN: your_domain.auth0.com
```
binh-dam-ibigroup, line 1: @landonreed we still need these auth0 config params, don't we?

binh-dam-ibigroup (commented):
@br648 Please see my comment-related comments so far.

br648:
Branch updated to address Binh's comments and design updates.

configurations/default/env.yml.tmp
binh-dam-ibigroup: @br648 We still need `AUTH0_DOMAIN`, `AUTH0_API_CLIENT`, and `AUTH0_API_SECRET`. Requests are not going through without.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
```diff
@@ -0,0 +1,64 @@
...
+        URI uri = URI.create(uriBuilder.toString());
+        LOG.debug("Constructed URI: {}", uri);
+        return uri;
+    }
+
+    private OtpDispatcherResponse call(URI uri) {
+        OtpDispatcherResponse otpDispatcherResponse = null;
+        HttpClient client = HttpClient.newHttpClient();
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(uri)
```
binh-dam-ibigroup, line 59: I think your IDE is doing the same indentation thing as mine. I'd prefer just having one tab instead of two.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+        OtpDispatcher otpDispatcher = new OtpDispatcherImpl(otpServer);
+        OtpDispatcherResponse otpDispatcherResponse = otpDispatcher.getPlan(request.queryString(), endPoint);
+        if (otpDispatcherResponse == null) {
+            logMessageAndHalt(request, HttpStatus.INTERNAL_SERVER_ERROR_500, "No response from OTP server.");
+            return null;
+        }
+
+        User user = null;
+        String userId = request.queryParams(USER_ID);
+        if (userId != null)
```
binh-dam-ibigroup: I would recommend to use brackets and stay away from multi-line if/else without brackets.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+            //TODO log with logging system?
+            LOG.warn("User id not provided, this will be an anonymous request");
+
+        org.opentripplanner.middleware.otp.core.api.resource.Response otpResponse = otpDispatcherResponse.getResponse();
+        // only save trip details if user is known and a response from OTP is provided
+        if (user != null && otpResponse != null) {
+            TripRequest tripRequest = new TripRequest(userId, batchId, request.queryParams(FROM_PLACE), request.queryParams(TO_PLACE), request.queryString());
+            TripPlan tripPlan = otpResponse.getPlan();
+            TripSummary tripSummary;
+            if (tripPlan != null)
```
binh-dam-ibigroup: I would use brackets here.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+                tripSummary = new TripSummary(otpResponse.getPlan().from, otpResponse.getPlan().to, otpResponse.getError(), otpResponse.getPlan().itinerary, tripRequest.id);
+            else
+                tripSummary = new TripSummary(otpResponse.getError(), tripRequest.id);
+
+            // only save trip summary if the trip request was saved
+            if (saveTripRequest(tripRequest))
```
binh-dam-ibigroup: Brackets

src/main/java/org/opentripplanner/middleware/otp/core/util/Properties.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.text.MessageFormat;
+import java.util.Locale;
+import java.util.ResourceBundle;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The purpose of Properties is to easily read a ResourceBundel (set of localized .properties files), and get the named contents.
```
binh-dam-ibigroup: Typo: ResourceBundle (I know it probably comes from OTP source)

binh-dam-ibigroup (changes requested):
We need to reinstate the auth0 env variables. Besides that I have some stylistic changes outlined in the comments, but otherwise it looks pretty good!

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
```diff
@@ -0,0 +1,179 @@
...
+import org.slf4j.LoggerFactory;
+
+public class IntUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(IntUtils.class);
+    public  static final String POINT_PREFIX = "POINT("; 
+    
+    /** does this string appear to be a coordinate of some sort */
+    public static boolean looksLikeCoordinate(String str)
+    {
+        if(str != null && (str.contains(POINT_PREFIX) || str.matches("[\\s]*[0-9\\-.]+[,\\s]+[0-9\\-.]+[\\s]*")))
```
binh-dam-ibigroup: This might also be from the OTP source, but I would add brackets to multi-line ifs throughout the file.

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -0,0 +1,58 @@
+package org.opentripplanner.middleware.models;
+
+/**
+ * A trip request represents an OTP UI trip request (initiated by a user) destine for an OpenTripPlanner instance.
```
binh-dam-ibigroup: Typo: 'destined'.

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -0,0 +1,58 @@
...
+ * otp-middleware stores these trip requests for trip monitoring purposes.
```
binh-dam-ibigroup: Not exactly. We store trip requests for reporting purposes.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public static String getParamFromRequest(Request req, String paramName) {
+        String paramValue = req.queryParams(paramName);
+        if (paramValue == null) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Must provide parameter name.");
+        }
+        return paramValue;
+    }
+
+    public static LocalDate getDateFromRequestParam(Request req, DateTimeFormatter expectDateFormat, String expectedDatePattern, String paramName) {
```
binh-dam-ibigroup: Add comments for this method.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,72 @@
...
+        String param = null;
+        try {
+            param = HttpUtils.getParamFromRequest(request, LIMIT_PARAM_NAME);
+            limit = Integer.parseInt(param);
+        } catch (NumberFormatException e) {
+            LOG.error("Unable to parse {} : {}. Using default limit: {}", LIMIT_PARAM_NAME, param, DEFAULT_LIMIT, e);
+        }
+
+        LocalDate fromDate = HttpUtils.getDateFromRequestParam(request, expectDateFormat, expectedDatePattern, FROM_DATE_PARAM_NAME);
+        LocalDate toDate = HttpUtils.getDateFromRequestParam(request, expectDateFormat, expectedDatePattern, TO_DATE_PARAM_NAME);
```
binh-dam-ibigroup: There should be checks whether `fromDate` or `toDate` are null. If a date is null, you want to not pass that date's comparison filter to Mongo.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -0,0 +1,80 @@
...
+        System.out.println("Saved trip summary:" + tripSummary.toString());
+        return tripSummary;
+    }
+
+    public static TripSummary createTripSummaryWithError() {
+        OtpDispatcherResponse response = getPlanErrorFromOtp();
+        Response otpResponse = response.getResponse();
+        TripPlan tripPlan = otpResponse.getPlan();
+        TripSummary tripSummary;
+        if (tripPlan != null)
```
binh-dam-ibigroup: Brackets, please.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -0,0 +1,80 @@
...
+        int i = 0;
+        while (i < amount) {
+            tripRequests.add(createTripRequest(userId));
+            i++;
+        }
+        return tripRequests;
+    }
+
+    public static void deleteTripRequests(List<TripRequest> tripRequests) {
+        for(TripRequest tripRequest: tripRequests) {
```
binh-dam-ibigroup: Space after `for`

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,72 @@
...
+            .parseDefaulting(ChronoField.NANO_OF_DAY, 0)
+            .toFormatter()
+            .withZone(ZoneId.systemDefault());
+
+        String userId = HttpUtils.getParamFromRequest(request, USER_ID_PARAM_NAME);
+        int limit = DEFAULT_LIMIT;
+
+        String param = null;
+        try {
+            param = HttpUtils.getParamFromRequest(request, LIMIT_PARAM_NAME);
```
binh-dam-ibigroup: Besides `userId`, I don't think the other query parameters should be required.

binh-dam-ibigroup (changes requested):
Please see comments.

br648:
Updated to handle optional parameters and address other issues highlighted.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -27,46 +34,96 @@
...
+            }
         } catch (NumberFormatException e) {
-            LOG.error("Unable to parse {} : {}. Using default limit: {}", LIMIT_PARAM_NAME, param, DEFAULT_LIMIT, e);
+            LOG.error("Unable to parse {} value of {}. Using default limit: {}", LIMIT_PARAM_NAME, paramLimit, DEFAULT_LIMIT, e);
         }
 
-        LocalDate fromDate = HttpUtils.getDateFromRequestParam(request, expectDateFormat, expectedDatePattern, FROM_DATE_PARAM_NAME);
-        LocalDate toDate = HttpUtils.getDateFromRequestParam(request, expectDateFormat, expectedDatePattern, TO_DATE_PARAM_NAME);
+        String paramFromDate = HttpUtils.getParamFromRequest(request, FROM_DATE_PARAM_NAME, true);
+        Date fromDate = getDate(request, FROM_DATE_PARAM_NAME, paramFromDate, LocalTime.MIN);
```
binh-dam-ibigroup: Suggestion: rename `LocalTime.MIN` to `LocalTime.MIDNIGHT`

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -27,46 +34,96 @@
...
 
-        LocalDateTime fromStartOfDay = fromDate.atTime(LocalTime.MIN);
-        LocalDateTime toEndOfDay = toDate.atTime(LocalTime.MAX);
-
-        Bson filter = Filters.and(gte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(fromStartOfDay.atZone(ZoneId.systemDefault()).toInstant())),
-            lte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(toEndOfDay.atZone(ZoneId.systemDefault()).toInstant())),
-            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, userId));
+        String paramToDate = HttpUtils.getParamFromRequest(request, TO_DATE_PARAM_NAME, true);
+        Date toDate = getDate(request, TO_DATE_PARAM_NAME, paramToDate, LocalTime.MAX);
```
binh-dam-ibigroup, line 74: Suggestion: rename `LocalTime.MAX` to `LocalTime._23H59M`

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final String LIMIT_PARAM_NAME = "limit";
+    private static final int DEFAULT_LIMIT = 10;
+
+    /**
+     * Return JSON representation of a user's trip request history based on provided parameters.
+     * Only the user id is required.
+     */
+    public static String getTripRequests(Request request, Response response, TypedPersistence<TripRequest> tripRequest) {
+
+        Auth0UserProfile requestingUser = Auth0Connection.getUserFromRequest(request);
```
binh-dam-ibigroup: Note that this can be null if a user with the requested id does not exist. (I don't think that would be a problem under most circumstances.)

configurations/test/env.yml
```diff
@@ -1,3 +1,4 @@
 #AUTH0_XYZ: TEST
 #AUTH0_SECRET: TEST
-MONGO_DB_NAME: otp_middleware_test
\ No newline at end of file
+MONGO_DB_NAME: otp_middleware_test
+EXPECTED_DATE_PATTERN: yyyy-MM-dd
```
binh-dam-ibigroup: We don't need `EXPECTED_DATE_PATTERN` anymore.

binh-dam-ibigroup (changes requested):
Nice. Just a few more minute changes, then I'll approve.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: @binh-dam-ibigroup I don't understand your suggestion LocalTime._23H59M? This is not an option under java.time.LocalTime. I think LocalTime.MAX which provides 23:59:59 is ideal here.

binh-dam-ibigroup: It is fine to keep it that way... I actually thought you created those variables!

binh-dam-ibigroup (approved):
Thanks for the changes!

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,146 @@
...
+        if (requestingUser == null) {
+            logMessageAndHalt(request, HttpStatus.FORBIDDEN_403, "Authorization required.");
+        }
+
+        final String userId = HttpUtils.getParamFromRequest(request, USER_ID_PARAM_NAME, false);
+        OtpUser user = Persistence.otpUsers.getById(userId);
+        if (user == null) {
+            logMessageAndHalt(request, HttpStatus.FORBIDDEN_403, "Unknown user.");
```
binh-dam-ibigroup: Should this be 404 instead?

br648: I've replace this section of code with isValidUser. Will update to 404 within that method, but under https://github.com/ibi-group/otp-middleware/pull/21

binh-dam-ibigroup (changes requested):
Sorry @br648, one more thing, please resolve merge conflics, then we should be good to go.

src/main/java/org/opentripplanner/middleware/BasicOtpDispatcher.java
```diff
@@ -1,5 +1,8 @@
 package org.opentripplanner.middleware;
 
+import spark.Request;
+import spark.Response;
+
```
landonreed, line 5: Is this file even needed any longer, @binh-dam-ibigroup?

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+        OtpDispatcherResponse otpDispatcherResponse = otpDispatcher.getPlan(request.queryString(), endPoint);
+        if (otpDispatcherResponse == null) {
+            logMessageAndHalt(request, HttpStatus.INTERNAL_SERVER_ERROR_500, "No response from OTP server.");
+            return null;
+        }
+
+        OtpUser user = null;
+        String userId = request.queryParams(USER_ID);
+        if (userId != null) {
+            user = Persistence.otpUsers.getById(userId);
```
landonreed: After further thought, I think we should `checkUser` like we do on `api/secure` paths. The endpoint can still be unsecured, but now I think that instead of passing the `userId` as a query param we should just path the Auth token in the request. This method can then check the user. We don't need to return 403 if the user is not found, but if they are found (and the user object tells us to store trips) we should store it.

src/main/java/org/opentripplanner/middleware/spark/Main.java
```diff
@@ -76,17 +87,23 @@ private static void initializeHttpEndpoints() throws IOException {
         // available at http://localhost:4567/async
         spark.get("/async", (req, res) -> BasicOtpDispatcher.executeRequestsAsync());
 
-        spark.before(API_PREFIX + "secure/*", ((request, response) -> {
+        // available at http://localhost:4567/plan
+        spark.get(UNSECURE_PLAN, (request, response) -> OtpRequestProcessor.planning(request, response, getConfigPropertyAsText("OTP_SERVER"), getConfigPropertyAsText("OTP_SERVER_PLAN_END_POINT")));
```
landonreed: I wonder if instead of just serving as a proxy for the `/plan` endpoint, we also want to serve as proxy for all OTP endpoints. We would then only post-process the plan responses (for storing trip history/summary), but that could simplify the configuration for the UI (we would have only one "OTP" server rather than splitting up things between "plan" and other endpoints). @evansiroky or @binh-dam-ibigroup, thoughts?

src/main/java/org/opentripplanner/middleware/spark/Main.java
```diff
@@ -76,17 +87,23 @@ private static void initializeHttpEndpoints() throws IOException {
...
+
+        // available at http://localhost:4567/api/secure/triprequests
+        spark.get(API_SECURE_TRIP_REQUESTS, (request, response) -> TripHistoryController.getTripRequests(request, response, tripRequest));
```
landonreed: Generally, I'm not crazy about putting strings into constants far away from their usage, especially if the constant is only invoked once.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    private static Bson buildFilter(String userId, Date fromDate, Date toDate) {
+
+        final String createdFieldName = "dateCreated";
+        final String userIdFieldName = "userId";
+
+        // user id is required, so as a minimum return all trip requests for user
+        Bson filter = Filters.and(eq(userIdFieldName, userId));
+
+        if (fromDate != null && toDate != null) { // get all trips between start and end dates
```
landonreed: Minor style thing: we generally put comments like this on their own line at the start of the conditional block (i.e., just below where it is located now).

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        String paramFromDate = HttpUtils.getParamFromRequest(request, FROM_DATE_PARAM_NAME, true);
+        Date fromDate = getDate(request, FROM_DATE_PARAM_NAME, paramFromDate, LocalTime.MIDNIGHT);
+
+        String paramToDate = HttpUtils.getParamFromRequest(request, TO_DATE_PARAM_NAME, true);
+        Date toDate = getDate(request, TO_DATE_PARAM_NAME, paramToDate, LocalTime.MAX);
+
+        if (fromDate != null && toDate != null && toDate.before(fromDate)) {
+            logMessageAndHalt(request, HttpStatus.BAD_REQUEST_400, TO_DATE_PARAM_NAME + " (" + paramToDate + ") before " + FROM_DATE_PARAM_NAME + " (" + paramFromDate + ")");
```
landonreed: Could you place these args on multiple lines so that the line width is shorter. Also, I think it might help legibility if you used `String.format` to construct the message.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+        if (fromDate != null && toDate != null) { // get all trips between start and end dates
+            filter = Filters.and(
+                gte(createdFieldName, fromDate),
+                lte(createdFieldName, toDate),
+                eq(userIdFieldName, userId));
+        } else if (fromDate == null && toDate != null) { // get all trip requests to end date
+            filter = Filters.and(
+                lte(createdFieldName, toDate),
+                eq(userIdFieldName, userId));
+        } else if (fromDate != null && toDate == null) { // get all trip requests from start date
```
landonreed: I think these comments could be improved. As is, they make me think longer than a couple of seconds to process. E.g.:

```
// Get all trip requests that occurred from the first recorded until the supplied end date.
```

and

```
// Get all trip requests that occurred from supplied start date until now.
```

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Build the filter which is passed to Mongo based on available parameters
+     */
+    private static Bson buildFilter(String userId, Date fromDate, Date toDate) {
+
+        final String createdFieldName = "dateCreated";
+        final String userIdFieldName = "userId";
+
+        // user id is required, so as a minimum return all trip requests for user
+        Bson filter = Filters.and(eq(userIdFieldName, userId));
```
landonreed: In `buildFilter` there's a lot of duplicated code that makes this hard to follow. Recommendation: start building a set of filters here and add to the set in the conditional blocks. E.g.:

```java
Set<Bson> clauses = new HashSet<>();
// User id is required, so as a minimum return all trip requests for user.
clauses.add(eq(userIdFieldName, userId));
if (fromDate != null && toDate != null) { 
  // Get all trips between start and end dates.
  clauses.add(gte(createdFieldName, fromDate));
  clauses.add(lte(createdFieldName, toDate));
} 
// etc.....
// Return ANDed filters.
return Filters.and(clauses);
```

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+            localDate = DateUtils.getDateFromParam(paramName, paramValue, expectedDatePattern);
+        } catch (DateTimeParseException e) {
+            logMessageAndHalt(request, HttpStatus.BAD_REQUEST_400, paramName + " value: " + paramValue + " is not a valid date. Must be in the format: " + expectedDatePattern);
+        }
+
+        if (localDate == null) {
+            return null;
+        }
+
+        return Date.from(localDate.atTime(timeOfDay).atZone(ZoneId.systemDefault()).toInstant());
```
landonreed: I find this somewhat hard to read. Maybe separate the dot notation with new lines to help?

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A trip request represents an OTP UI trip request (initiated by a user) destined for an OpenTripPlanner instance.
+ * otp-middleware stores these trip requests for reporting purposes.
+ */
+public class TripRequest extends Model {
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * User Id. Id of user making trip request.
```
landonreed: Might be good to add an `{@link}` javadoc style to the actual field it references.

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public String userId;
+
+    /**
+     * Batch Id. Id for trip plans of different modes.
```
landonreed: Maybe: `Batch Id. Id for trip requests planned together but representing different modes.`

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String FROM_DATE_PARAM_NAME = "fromDate";
+    private static final String TO_DATE_PARAM_NAME = "toDate";
+    private static final String LIMIT_PARAM_NAME = "limit";
+    private static final int DEFAULT_LIMIT = 10;
+
+    /**
+     * Return JSON representation of a user's trip request history based on provided parameters.
+     * An authorized user (Auth0) and user id are required.
+     */
+    public static String getTripRequests(Request request, Response response, TypedPersistence<TripRequest> tripRequest) {
```
landonreed: Not sure if long term we want this method to take `TypedPersistence<TripRequest>` as an arg. Maybe just add a TODO about possibly refactoring for now?

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -0,0 +1,5 @@
+package org.opentripplanner.middleware.otp;
+
+public interface OtpDispatcher {
```
landonreed, line 6: Needs javadoc

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
```diff
@@ -0,0 +1,25 @@
+package org.opentripplanner.middleware;
+
+import org.opentripplanner.middleware.otp.OtpDispatcherImpl;
+import org.opentripplanner.middleware.otp.OtpDispatcherResponse;
+
+public class OtpDispatcherHelper {
```
landonreed: Needs javadoc

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+
+/**
+ * Responsible for constructing requests to elected OTP server end point using original query parameters provided by MOD UI.
+ * To provide the response from the OTP server in the form of status code and body so the correct response can be provided to the calling MOD UI.
+ */
+public class OtpDispatcherImpl implements OtpDispatcher {
```
landonreed, line 18: Methods in this class need comments. It's hard to tell what's going on.

src/main/resources/Message.properties
```diff
@@ -0,0 +1,31 @@
+#
+# Strings used by the Messages.java enum (i.e., if that enum's name changes, the properties files' names must change)
```
landonreed: Typo: Message.java

src/main/java/org/opentripplanner/middleware/otp/core/api/common/Message.java
```diff
@@ -0,0 +1,106 @@
+package org.opentripplanner.middleware.otp.core.api.common;
```
landonreed: Anywhere code is derived from OpenTripPlanner, we should explicitly state this in the class. Probably also should add a package info file at the root package.

src/main/java/org/opentripplanner/middleware/otp/core/api/model/error/PlannerError.java
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+
+/** This API response element represents an error in trip planning. */
+public class PlannerError {
+
+    private static final Logger LOG = LoggerFactory.getLogger(org.opentripplanner.middleware.otp.core.api.model.error.PlannerError.class);
```
landonreed: No need to write package name, is there?

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -151,4 +151,38 @@ private static JWTVerifier getVerifier(Request req, String token) {
...
+            logMessageAndHalt(request, HttpStatus.NOT_FOUND_404, "Unknown user.");
+        }
+
+        return profile;
+    }
+
+    /**
+     * Confirm that the user's actions are on their items if not admin.
+     */
+    public static void isAuthorized(String userId, Request request) {
```
landonreed, line 178: Let's make sure this isn't duplicating what's been added by #14.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+import org.opentripplanner.middleware.models.TripSummary;
+import org.opentripplanner.middleware.models.OtpUser;
+import org.opentripplanner.middleware.otp.core.api.model.TripPlan;
+import org.opentripplanner.middleware.persistence.Persistence;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.Request;
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+public class OtpRequestProcessor {
```
landonreed, line 26: Needs javadoc

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(OtpRequestProcessor.class);
+    private static final String USER_ID = "userId";
+    private static final String BATCH_ID = "batchId";
+    private static final String FROM_PLACE = "fromPlace";
+    private static final String TO_PLACE = "toPlace";
+
+    public static String planning(Request request, spark.Response response, String otpServer, String endPoint) {
```
landonreed: Would prefer that args map to Spark.get and that we set the otpServer and endPoint another way.

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return retVal;
+    }
+
+    public static Integer getIntegerFromString(String input) {
+        try {
+            return Integer.parseInt(input);
+        } catch (Exception e) {
+            try {
+                String in = input.replaceAll("\\D", "");
```
landonreed: regex should have comments.

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
```diff
@@ -0,0 +1,46 @@
+package org.opentripplanner.middleware.otp.core.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IntUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(IntUtils.class);
+
+    public static int getIntFromString(String input) {
```
landonreed: Methods in this class should have javadoc.

src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
```diff
@@ -72,6 +72,8 @@ public static void initialize () {
         otpUsers = new TypedPersistence(mongoDatabase, OtpUser.class);
         adminUsers = new TypedPersistence(mongoDatabase, AdminUser.class);
         apiUsers = new TypedPersistence(mongoDatabase, ApiUser.class);
+        tripRequest = new TypedPersistence(mongoDatabase, TripRequest.class);
+        tripSummary = new TypedPersistence(mongoDatabase, TripSummary.class);
```
landonreed: These should be pluralized like the other collections.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -139,6 +148,7 @@ public T getById (String id) {
         return mongoCollection.find(filter).into(new ArrayList<>());
     }
 
+
```
landonreed: Unneeded whitespace.

src/main/java/org/opentripplanner/middleware/utils/DateUtils.java
```diff
@@ -0,0 +1,44 @@
...
+     * Get {@Link java.time.LocalDate} from provided value base on expected date format.
+     */
+    public static LocalDate getDateFromParam(String paramName, String paramValue, String expectedDatePattern) throws DateTimeParseException {
+
+        // no date value to work with
+        if (paramValue == null) {
+            return null;
+        }
+
+        DateTimeFormatter expectDateFormat = new DateTimeFormatterBuilder().appendPattern(expectedDatePattern)
```
landonreed: Please move `appendPattern` to new line like the other chained method calls. Also, `expectedDateFormat` might be what you meant (typo).

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String OTP_SERVER_PLAN_END_POINT = "/otp/routers/default/plan";
+
+    public static OtpDispatcherResponse getPlanFromOtp() {
+        OtpDispatcherImpl otpDispatcher = new OtpDispatcherImpl(OTP_SERVER);
+        OtpDispatcherResponse response = otpDispatcher.getPlan("plan?userId=123456&fromPlace=28.54894%2C%20-81.38971%3A%3A28.548944048426772%2C-81.38970606029034&toPlace=28.53989%2C%20-81.37728%3A%3A28.539893820446867%2C-81.37727737426759&date=2020-05-05&time=12%3A04&arriveBy=false&mode=WALK%2CBUS%2CRAIL&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&companies=", OTP_SERVER_PLAN_END_POINT);
+        System.out.println("OTP Plan response:" + response.toString());
+        return response;
+    }
+
+    // fromPalce instead of fromPlace to produce error
```
landonreed: Please change to javadoc format.

landonreed (changes requested):
Looking great, but I do have a few comments. 

One big thing I'm struggling with is whether we should generalize OtpDispatcher to send any request to OTP (not just plan requests) and if there is a good way to pass along the request without parsing it as much before it gets handed off (are there some native proxy classes we can make use of?).

Another thing I'm wondering about is the inclusion of OTP classes. We should discuss as a team if that's the approach we want to take or if we want to pare down those classes more.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/spark/Main.java
br648: Updated and will check for other occurrences.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: This is much cleaner now, thanks.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: Added todo for now, but would rather refactor. What do you have in mind? Could update Persistence to return required tripRequest obj?

src/main/java/org/opentripplanner/middleware/otp/core/api/common/Message.java
br648: For now I have added the package info file. Once a decision on the use of OTP classes has been made I'll either add a comment to all the class or refactor and make unique.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: Pretty sure this is ok. https://github.com/ibi-group/otp-middleware/pull/14 appears to be checking if the user is admin. This method authorizes admins and users if they originally created the item they are requesting.

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
br648: Will update if we are keeping OTP classes

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
br648: Will update if we are keeping OTP classes

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
br648: Binh provided the param to update in Intellij

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: Ignore my last comment. Have updated for review.

src/main/java/org/opentripplanner/middleware/BasicOtpDispatcher.java
binh-dam-ibigroup: That was a remnant from when we were experimenting. You can probably remove it in #13.

src/main/java/org/opentripplanner/middleware/spark/Main.java
binh-dam-ibigroup: The only difference in the UI config is the path for the OTP router. No path changes have been needed elsewhere in the UI.

binh-dam-ibigroup (approved):
LGTM

binh-dam-ibigroup (changes requested):
Sorry, reverting to Request Changes. The UI currently forwards two plan requests for each trip plan, and I don't think we should record that request twice. The previously used `userId` query parameter was sent only once to accommodate that.

binh-dam-ibigroup (approved):
Approve after agreeing to defer.

src/main/java/org/opentripplanner/middleware/otp/core/api/common/Message.java
br648: I have now removed the package info file as no classes are now derived (copied directly) from OTP.

src/main/java/org/opentripplanner/middleware/otp/core/util/IntUtils.java
br648: We are not keeping OTP classes. This class no longer exists.

binh-dam-ibigroup (changes requested):
Just discovered that anonymous searches don't work anymore (`Halting with status code 401.  Error message: Authorization header is missing.` is shown in the log).

br648:
@landonreed I have now updated to include bespoke POJOs for handling the OTP plan response. A couple of things, my testing is limited and I certainly haven’t covered all permutations. I’ve gone down a bit of a rabbit hole in order to get fare and leg information. The amount of information is probably overkill, but I can remove parameters you don’t want. Just let me know.

src/main/java/org/opentripplanner/middleware/spark/Main.java
br648: @landonreed once a decision has been made on this please let me know and I'll implement... or do nothing.

br648:
> Just discovered that anonymous searches don't work anymore (`Halting with status code 401. Error message: Authorization header is missing.` is shown in the log).

This is by design following this comment from @landonreed :

"After further thought, I think we should checkUser like we do on api/secure paths. The endpoint can still be unsecured, but now I think that instead of passing the userId as a query param we should just path the Auth token in the request. This method can then check the user. We don't need to return 403 if the user is not found, but if they are found (and the user object tells us to store trips) we should store it."

binh-dam-ibigroup:
@br648 the plan endpoint should still return itineraries even if the authorization header is missing.

br648:
> @br648 the plan endpoint should still return itineraries even if the authorization header is missing.

ok, will look into it first thing Monday. Will have to create something new as checkUser isn't sutiable.

br648:
> 
> 
> @br648 the plan endpoint should still return itineraries even if the authorization header is missing.

@binh-dam-ibigroup - This has been addressed and the update pushed.

src/main/java/org/opentripplanner/middleware/spark/Main.java
binh-dam-ibigroup: @landonreed I see what you mean. The change in this PR breaks the route, stop, and trip viewer. We can either do it on this PR or in a separate PR. I am fine either way.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -151,4 +157,38 @@ private static JWTVerifier getVerifier(Request req, String token) {
...
+
+        Auth0UserProfile profile = isValidUser(request);
+
+        // let admin do anything
+        if (profile.adminUser != null) {
+            return;
+        }
+
+        if (userId == null ||
+            (profile.otpUser != null && !profile.otpUser.id.equalsIgnoreCase(userId)) ||
```
binh-dam-ibigroup: Here also, this should be `.equals(...)`

br648: yeah, will be in the next commit.

binh-dam-ibigroup (approved):
One more tiny change then I'm inclined to approve for the Trip Request/Trip Summary aspect. I defer to @landonreed for the handling of other OTP endpoints.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -64,6 +64,12 @@ public static void checkUser(Request req) {
         }
     }
 
+    public static boolean isAuthHeaderPresent(Request req) {
+        final String authHeader = req.headers("Authorization");
+        return (authHeader != null) ? true : false;
+    }
```
landonreed, line 70: This method should also be used in `getTokenFromRequest`. Also, the return statement can be simplified to 
```java
return authHeader != null;
```

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
```diff
@@ -42,6 +51,11 @@ private URI buildUri(String params, String endPoint) {
         return uri;
     }
 
+    /**
+     * Makes a call to the OTP server plan end point and converts, if possible, the response into POJOs. The POJOs,
+     * original response and status are wrapped in a single object and returned. It will fail if a connection is not
+     * made or is unable to parse the plan response.
+     */
```
landonreed, line 55: I'm just not sure if we want to really fail to respond just because we couldn't parse the response. To me, it's acceptable if failure to parse ends up meaning we cannot store some results, but failing to respond to the client on parse failure seems like something we want to avoid.

landonreed: In other words, it seems like we should move the parsing logic outside of this dispatcher and into the OtpRequestProcessor.

src/main/java/org/opentripplanner/middleware/otp/response/Leg.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+/**
+ * Plan response, itinerary leg information. Produced using http://www.jsonschema2pojo.org/
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Leg {
+
+    @JsonProperty("startTime")
```
landonreed: Are all of these `@JsonProperty` annotations needed?

src/main/java/org/opentripplanner/middleware/otp/response/Place.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Set;
+
+/**
+ * Plan response, place information. Produced using http://www.jsonschema2pojo.org/
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Place {
+
+    @JsonProperty("name")
```
landonreed: Are all of these `@JsonProperty` annotations needed?

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Duration of the trip on this itinerary, in seconds.
+     */
+    public Long duration = 0L;
+
+    /**
+     * Time that the trip departs.
+     */
+    public Date startTime = null;
```
binh-dam-ibigroup, line 22: Using `Date` for `startTime` and `endTime` is causing an issue when saving a trip under #21, I need to investigate more.

landonreed (commented):
Looks really good. See #24 for some other changes. I still have lingering questions about the POJOs needed to store the request summaries. Curious to hear what @evansiroky's reaction is to this approach.

src/main/java/org/opentripplanner/middleware/otp/response/Leg.java
br648: It doesn't look like it. Have removed.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
br648: For completeness, I have merge #25 into #21

src/main/java/org/opentripplanner/middleware/otp/response/Place.java
br648: It doesn't look like it. Have removed.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
br648: Have implemented this.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: Updated.

binh-dam-ibigroup: Not a big deal, but notice that the header is retrieved again two lines after this function call...

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,137 @@
...
+        String paramToDate = HttpUtils.getParamFromRequest(request, TO_DATE_PARAM_NAME, true);
+        Date toDate = getDate(request, TO_DATE_PARAM_NAME, paramToDate, LocalTime.MAX);
+
+        if (fromDate != null && toDate != null && toDate.before(fromDate)) {
+            logMessageAndHalt(request, HttpStatus.BAD_REQUEST_400,
+                String.format("%s (%s) before %s (%s)", TO_DATE_PARAM_NAME , paramToDate, FROM_DATE_PARAM_NAME,
+                    paramFromDate));
+        }
+
+        Bson filter = buildFilter(userId,fromDate, toDate);
```
binh-dam-ibigroup, line 82: Need space before `fromDate`.

binh-dam-ibigroup: Need space before `fromDate`.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -42,13 +60,161 @@ public void canDeleteUser() {
...
+
+        LocalDateTime fromStartOfDay = LocalDate.now().atTime(LocalTime.MIN);
+        LocalDateTime toEndOfDay = LocalDate.now().atTime(LocalTime.MAX);
+
+        Bson filter = Filters.and(gte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(fromStartOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            lte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(toEndOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, user.id));
+
+        List<TripRequest> result = Persistence.tripRequests.getFilteredWithLimit(filter, limit);
+        assertEquals(result.size(),tripRequests.size());
```
binh-dam-ibigroup: Need space after comma.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -42,13 +60,161 @@ public void canDeleteUser() {
...
+
+        List<TripRequest> tripRequests = createTripRequests(limit, user.id);
+
+        LocalDateTime fromStartOfDay = LocalDate.now().atTime(LocalTime.MIN);
+
+        Bson filter = Filters.and(gte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(fromStartOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, user.id));
+
+        List<TripRequest> result = Persistence.tripRequests.getFilteredWithLimit(filter, limit);
+        assertEquals(result.size(),tripRequests.size());
```
binh-dam-ibigroup: Need space after comma.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -42,13 +60,161 @@ public void canDeleteUser() {
...
+        List<TripRequest> tripRequests = createTripRequests(limit, user.id);
+
+        LocalDateTime toEndOfDay = LocalDate.now().atTime(LocalTime.MAX);
+
+        Bson filter = Filters.and(
+            lte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(toEndOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, user.id));
+
+        List<TripRequest> result = Persistence.tripRequests.getFilteredWithLimit(filter, limit);
+        assertEquals(result.size(),tripRequests.size());
```
binh-dam-ibigroup: Need space after comma.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -42,13 +60,161 @@ public void canDeleteUser() {
...
+
+        OtpUser user = createUser(TEST_EMAIL);
+
+        List<TripRequest> tripRequests = createTripRequests(limit, user.id);
+
+        Bson filter = Filters.and(
+            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, user.id));
+
+        List<TripRequest> result = Persistence.tripRequests.getFilteredWithLimit(filter, limit);
+        assertEquals(result.size(),tripRequests.size());
```
binh-dam-ibigroup: Need space after comma.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -154,19 +163,20 @@ public T getById (String id) {
...
-    public T getOneFiltered (Bson filter) {
+    public T getOneFiltered(Bson filter) {
         return getOneFiltered(filter, null);
     }
 
-    public boolean removeById (String id) {
+    public boolean removeById(String id) {
         DeleteResult result = mongoCollection.deleteOne(eq(id));
         if (result.getDeletedCount() == 1) {
             LOG.info("Deleted object id={} type={}", id, collectionName);
```
binh-dam-ibigroup, line 182: L185: Need space after comma.

binh-dam-ibigroup (changes requested):
Merge conflicts and some cosmetic changes.

binh-dam-ibigroup (approved):
LGTM

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -0,0 +1,137 @@
...
+                String.format("%s value: %s is not a valid date. Must be in the format: %s", paramName, paramValue,
+                    expectedDatePattern));
+        }
+
+        if (localDate == null) {
+            return null;
+        }
+
+        return Date.from(localDate.atTime(timeOfDay)
+                .atZone(ZoneId.systemDefault())
```
evansiroky, line 135: This method's JavaDoc should note the use of the system timezone for parsing a date.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
```diff
@@ -0,0 +1,81 @@
...
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+
+/**
+ * Responsible for constructing requests to elected OTP server end point using original query parameters provided by MOD UI.
+ * To provide the response from the OTP server in the form of status code and body so the correct response can be provided to the calling MOD UI.
+ */
+public class OtpDispatcherImpl implements OtpDispatcher {
```
evansiroky, line 18: I'm not a Java expert, so I don't quite understand the point of an Interface if there is just one implementation. Is there a good article I can read up on why this might be a good practice?

src/main/java/org/opentripplanner/middleware/utils/DateUtils.java
```diff
@@ -0,0 +1,49 @@
...
+        // no date value to work with
+        if (paramValue == null) {
+            return null;
+        }
+
+        DateTimeFormatter expectedDateFormat = new DateTimeFormatterBuilder()
+            .appendPattern(expectedDatePattern)
+            .parseDefaulting(ChronoField.NANO_OF_DAY, 0)
+            .toFormatter()
+            .withZone(ZoneId.systemDefault());
```
evansiroky, line 36: The JavaDoc should note use of system timezone.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import spark.Request;
+
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+public class HttpUtils {
+
+    /**
+     * Get entity attribute value from request.
+     */
+    public static String getParamFromRequest(Request req, String paramName, boolean allowNull) {
```
evansiroky: Can we rename this to `getRequiredParamFromRequest` and add a note that this method will halt the request if the param is not provided?

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
```diff
@@ -0,0 +1,40 @@
...
+import org.opentripplanner.middleware.otp.response.Response;
+import org.opentripplanner.middleware.utils.JsonUtils;
+
+/**
+ * Used to provide real plan responses from OTP
+ */
+public class OtpDispatcherHelper {
+
+    private static final String OTP_SERVER = "https://fdot-otp-server.ibi-transit.com";
+    private static final String OTP_SERVER_PLAN_END_POINT = "/otp/routers/default/plan";
```
evansiroky: These should probably be obtained from application config.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -42,13 +60,161 @@ public void canDeleteUser() {
...
+
+        Bson filter = Filters.and(gte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(fromStartOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            lte(TRIP_REQUEST_DATE_CREATED_FIELD_NAME, Date.from(toEndOfDay.atZone(ZoneId.systemDefault()).toInstant())),
+            eq(TRIP_REQUEST_USER_ID_FIELD_NAME, user.id));
+
+        List<TripRequest> result = Persistence.tripRequests.getFilteredWithLimit(filter, limit);
+        assertEquals(result.size(), tripRequests.size());
+
+        // tidy up
+        deleteTripRequests(tripRequests);
```
evansiroky: Is there a way to put this in an afterEach method that clears tripRequests after each test case?

evansiroky (changes requested):
Just a few comments.

evansiroky:
I'm adding the blocked tag to note dependency on #15.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
br648: There probably isn't much point if there is only one implementation. I always (rightly or wrongly) code to interfaces. If later on down the line there is a second implementation the impact will be minimal if any. Of course you already know this! Articles focus on interface best practises and not on single implementation. If you feel an interface isn't warranted, I can remove it.

@landonreed @binh-dam-ibigroup Thoughts?

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/DateUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
br648: Updated

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
br648: Moved the trip history tests into a separate class and made use of after each.

br648:
Various updates made to address @evansiroky review.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
```diff
@@ -0,0 +1,81 @@
+package org.opentripplanner.middleware.otp;
+
+import org.opentripplanner.middleware.otp.response.Response;
+import org.opentripplanner.middleware.utils.JsonUtils;
```
binh-dam-ibigroup: These two imports appear unused.

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
```diff
@@ -5,20 +5,19 @@
 import org.opentripplanner.middleware.otp.response.Response;
 import org.opentripplanner.middleware.utils.JsonUtils;
 
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
```
binh-dam-ibigroup: Does not appear to be used.

binh-dam-ibigroup (approved):
There are some unused imports here and there, but overall nice job @br648.

src/test/java/org/opentripplanner/middleware/OtpDispatcherHelper.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherImpl.java
br648: Updated.

binh-dam-ibigroup (dismissed):
I dnd't need to reapprove that

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
```diff
@@ -19,8 +20,11 @@
     public void canCreateUser() {
         OtpUser user = createUser(TEST_EMAIL);
         String id = user.id;
+        System.out.println("User id:" + id);
```
evansiroky: Let's always use LOG.info instead of System.out.println.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
```diff
@@ -0,0 +1,187 @@
...
+        String userId = "123456";
+        TripRequest tripRequestToDelete = createTripRequest(userId);
+        Persistence.tripRequests.removeById(tripRequestToDelete.id);
+        TripRequest tripRequest = Persistence.tripRequests.getById(tripRequestToDelete.id);
+        assertNull(tripRequest, "Deleted TripRequest should no longer exist in database (should return as null).");
+    }
+
+    @Test
+    public void canCreateTripSummaryWithError() {
+        tripSummary = createTripSummaryWithError(OTP_SERVER, OTP_SERVER_PLAN_END_POINT);
```
evansiroky: Oh yikes, sorry for not seeing this earlier, but we should be very careful about making HTTP requests during tests. Since this is merely a persistence test, we don't need to make an HTTP request. Instead, we should create a mock trip summary and save that. This has the added benefit of making the test faster and having it not fail if for some reason the HTTP request is unsuccessful.

Ideally, we'd also test the OtpDispatcher in a separate test file and use wiremock (see [this example of using wiremock](https://github.com/ibi-group/datatools-server/blob/c691cf6f22159b2cbeee45569c73b1777e1db1dc/src/test/java/com/conveyal/datatools/manager/controllers/api/UserControllerTest.java#L89) in datatools-server) to mock HTTP requests/repsonses.

configurations/test/env.yml
```diff
@@ -1,4 +1,6 @@
 AUTH0_DOMAIN: test.auth0.com
 AUTH0_API_CLIENT: test-auth0-client-id
 AUTH0_API_SECRET: test-auth0-secret
-MONGO_DB_NAME: otp_middleware_test
\ No newline at end of file
+MONGO_DB_NAME: otp_middleware_test
+OTP_SERVER: https://fdot-otp-server.ibi-transit.com
```
evansiroky: We should try to avoid storing actual live endpoints in test config. Let's keep this `http://www.example.com`.

evansiroky (changes requested):
See comments about test cases.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -44,53 +44,55 @@ public static String planning(Request request, spark.Response response) {
...
+                    request.queryParams("toPlace"), request.queryString());
+                // only save trip summary if the trip request was saved
+                if (saveTripRequest(tripRequest)) {
+                    TripPlan tripPlan = otpDispatcherResponse.response.plan;
+                    TripSummary tripSummary;
+                    if (tripPlan != null) {
+                        tripSummary = new TripSummary(tripPlan, otpDispatcherResponse.response.error, tripRequest.id);
+                    } else {
+                        tripSummary = new TripSummary(otpDispatcherResponse.response.error, tripRequest.id);
+                    }
```
landonreed: I realize that I probably could have added this to #24 (sorry about that), but I think I might prefer if we just handle the tripPlan null check in the TripSummary constructor. That way we can get rid of the overloaded TripSummary constructor (the one with two args).

landonreed (dismissed):
This is looking great. Just one more comment about the TripSummary constructor and addressing @evansiroky's testing concerns and I think it should be ready.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceTest.java
br648: Removed System.out.println

configurations/test/env.yml
br648: Parameters no longer required. Removed.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
br648: @evansiroky - The call to the OTP plan endpoint was just to obtain a plan response. I have now created static files containing these instead. This has rendered OtpDispatcherHelper.java redundant so have removed it.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
br648: Updated.

evansiroky (approved):

binh-dam-ibigroup (approved):

landonreed (approved):

PR #20, opened 2020-05-27
Check token expiration changes

landonreed:
### Checklist

### Description

@binh-dam-ibigroup, I've made some changes to your PR, let me know what you think.
- Change name of test class to match what we're testing (Auth0Connection is more so about handling the user Authorization token stuff, not the Management API stuff).
- Wrapped the TokenHolder in a TokenCache class and isolated the expiration checking there.
- provided public getter/setter for the test class to use.


src/main/java/org/opentripplanner/middleware/auth/TokenCache.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public static final long EXPIRATION_BUFFER_MILLIS = 60 * 1000;
+
+    public TokenCache(TokenHolder tokenHolder) {
+        this.tokenHolder = tokenHolder;
+        // Compute expiration time from current time + token duration in seconds.
+        expirationDate = new Date(System.currentTimeMillis() + tokenHolder.getExpiresIn() * 1000);
+    }
+
+    public boolean isTokenExpired() {
```
binh-dam-ibigroup: `isStale` captures the presence of a buffer better than `isExpired` imo.

binh-dam-ibigroup (approved):

binh-dam-ibigroup (changes requested):
Changing to `Request changes`. There is a side effect on the build that is causing a failure (probably a call to a class that relies on configuration)

src/test/java/org/opentripplanner/middleware/auth/Auth0UsersTest.java
```diff
@@ -0,0 +1,38 @@
...
+        Auth0Users.setCachedToken(new TokenCache(tokenHolder));
+    }
+
+    /**
+     * Checks that token is valid soon after creation and seconds until expiration does diminish.
+     */
+    @Test
+    public void isTokenStale() {
+        // Verify that token is not expired soon after it set.
+        assertFalse(Auth0Users.getCachedToken().isTokenExpired());
```
binh-dam-ibigroup: The call to `Auth0Users` triggers the init on all the static vars in that class, that's why I had these checks orignally in another class.

landonreed: Gotcha, yea, the pattern we're following here is to extend the OtpMiddlewareTest class, which will call its own setup method. Any new test classes should do the same.

binh-dam-ibigroup (approved):

PR #21, opened 2020-05-28
Monitored trip

br648:
### Checklist

### Description

Responsible for creating/updating/deleting monitored trips. Create, update and delete pre hooks validate a user prior to carrying out related action. Create has an additional check on the number of monitored trips previously created. This is set to 5 and will prevent new monitored trips being created if this is to be exceeded.


src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -80,6 +80,11 @@ public static void checkUserIsAdmin(Request req, Response res) {
         }
     }
 
+    /** Check if the incoming user is an admin user */
+    public static boolean isUserAdmin(Auth0UserProfile user) {
```
binh-dam-ibigroup, line 94: Make method `checkUserIsAdmin` call `isUserAdmin`.

landonreed: Actually, I wouldn't worry about making this change. I've added @br648's `isUserAdmin` method to #14 and am using the method in `checkUserIsAdmin` in that downstream PR. So I'd rather if we just leave that unchanged to avoid potential (even if unlikely) conflicts.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,79 @@
+package org.opentripplanner.middleware.models;
+
+import org.opentripplanner.middleware.otp.core.api.model.Itinerary;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A monitored trip represents
```
binh-dam-ibigroup: Finish this comment...

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class MonitoredTrip extends Model {
+
+    /**
+     * User Id. Id of user monitoring trip.
```
binh-dam-ibigroup: Say rather, "Mongo Id of user monitoring trip"

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,79 @@
...
+     * representation hasn't been used because it is irrelevant to calling systems.
+     *
+     * Expecting each effected day no more than once with the following representation:
+     *
+     * "monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"
+     */
+    public Set<String> days;
+
+    /**
+     * Specify if the monitored trip should be checked on a federal holiday.
```
binh-dam-ibigroup: US federal holiday

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Query params. Query parameters influencing trip.
+     */
+    public String queryParams;
+
+    /**
+     * The trips itinerary
+     */
+    public List<Itinerary> itinerary;
```
binh-dam-ibigroup: What was the decision on this @landonreed? Did we decide to focus on one itinerary or multiple?

binh-dam-ibigroup (changes requested):
Please merge with the latest changes from branch `user-sub-types` or `basic-user-prefs`. There is some code introduced in this PR that is no longer needed when merging with those other branches.
There are some other minor JS doc comment changes.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -148,10 +154,23 @@ protected void buildEndPoint(ApiEndpoint baseEndPoint) {
...
+    // FIXME Maybe better if the user check (and filtering) was done in a pre hook?
+    // FIXME Will require further granularity for admin
     private List<T> getMany(Request req, Response res) {
-        return persistence.getAll();
+
+        Auth0UserProfile requestingUser = getUserFromRequest(req);
+        if (isUserAdmin(requestingUser)) {
+            return persistence.getAll();
+        } else if (persistence.clazz == OtpUser.class) { 
+            Bson filter = Filters.and(eq("_id", requestingUser.otpUser.id));
```
binh-dam-ibigroup: Suggestion: You don't need to use `Filters.and` if there is just one condition. See https://github.com/ibi-group/otp-middleware/blob/4b0cb1ed711fb7ec834563340d2eafb61abba4d0/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java#L44

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -148,10 +154,23 @@ protected void buildEndPoint(ApiEndpoint baseEndPoint) {
...
+            return persistence.getFiltered(filter);
+        } else {
+            // FIXME assumes all non admin user collections will have a user id field
+            Bson filter = Filters.and(eq("userId", requestingUser.otpUser.id));
```
binh-dam-ibigroup: Same as above.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.opentripplanner.middleware.auth.Auth0Connection.isAuthorized;
+import static org.opentripplanner.middleware.auth.Auth0Connection.isValidUser;
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+/**
+ * Implementation of the {@link ApiController} abstract class for managing monitored trips. This controller connects
+ * with Auth0 services using the hooks provided by {@link ApiController}.
+ */
+public class MonitorTripController extends ApiController<MonitoredTrip> {
+    private static final long MAXIMUM_ALLOWED_MONITORED_TRIPS = 5;
```
binh-dam-ibigroup: I don't mind having this as a constant for now. I defer to the other reviewers.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -46,20 +46,15 @@
...
-        Auth0UserProfile requestingUser = Auth0Connection.getUserFromRequest(request);
-        if (requestingUser == null) {
-            logMessageAndHalt(request, HttpStatus.FORBIDDEN_403, "Authorization required.");
-        }
-
         final String userId = HttpUtils.getParamFromRequest(request, USER_ID_PARAM_NAME, false);
+
         OtpUser user = Persistence.otpUsers.getById(userId);
         if (user == null) {
             logMessageAndHalt(request, HttpStatus.FORBIDDEN_403, "Unknown user.");
```
binh-dam-ibigroup: Should this be 404?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Set;
+
+/**
+ * A monitored trip represents a trip a user would like to receive notification on if affected by a delay and/or
+ * route change.
+ */
+public class MonitoredTrip extends Model {
+
+    /**
+     * User Id. Mongo Id of user monitoring trip.
```
binh-dam-ibigroup: Suggestion: Remove first sentence in comment.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
binh-dam-ibigroup: I think it is ok for now because from the UI, we query the itinerary without real-time updates, and `queryParams` contains the index of the active itinerary.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -77,4 +80,63 @@ public static void deleteTripRequests(List<TripRequest> tripRequests) {
         }
     }
 
+    public static MonitoredTrip createMonitoredTrip(String userId) {
+        Set<String> days = new HashSet<>();
+        days.add("Monday");
```
binh-dam-ibigroup: Should the days lowercase or uppercase? See https://github.com/ibi-group/otp-middleware/blob/e85febf23a8accad01a28f49d26bcb63cc8053ad/src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java#L42

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: I think it should be 403, however, after reviewing I have removed this check as it replicates the check carried out by isValidUser.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
br648: I've changed to lowercase. I've also added a comment to the parameter in MonitoredTrip regarding case! Shame enums can't be used.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
binh-dam-ibigroup: ^ ^ I'll add this as a change to #17.

binh-dam-ibigroup (changes requested):
Minor refactors, then it's good to go for me.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Updated accordingly

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: Updated.

binh-dam-ibigroup (approved):
Nice job @br648 !

br648:
@binh-dam-ibigroup I've had to do further auth changes as the restrictions were too great. Admins didn't have the required permissions.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -155,21 +155,32 @@ public static boolean authDisabled() {
...
-        if (userId == null || profile == null || !profile.otpUser.id.equalsIgnoreCase(userId)) {
+        Auth0UserProfile profile = isValidUser(request);
+
+        // let admin do anything
+        if (profile.adminUser != null) {
+            return;
+        }
+
+        if (userId == null ||
+            (profile.otpUser != null && !profile.otpUser.id.equalsIgnoreCase(userId)) ||
```
binh-dam-ibigroup: Any reason for ignoring case on user Mongo ID? (Comment code accordingly)

binh-dam-ibigroup (changes requested):
Please resolve conflicts and justify ignoring case when checking user Mongo ID.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: I have reduced to just "equals".

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -140,18 +145,36 @@ public void bind(final SparkSwagger restApi) {
...
-        return persistence.getAll();
+
+        Auth0UserProfile requestingUser = getUserFromRequest(req);
+        if (isUserAdmin(requestingUser)) {
+            return persistence.getAll();
+        } else if (persistence.clazz == OtpUser.class) { 
+            Bson filter = Filters.eq("_id", requestingUser.otpUser.id);
+            return persistence.getFiltered(filter);
+        } else {
+            // FIXME assumes all non admin user collections will have a user id field
```
evansiroky: This seems like a reasonable design choice to me so I'd be fine with removing the FIXME.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
evansiroky: Let's make this be obtained from the application config.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    boolean preDeleteHook(MonitoredTrip monitoredTrip, Request req) {
+        // Authorization checks are done prior to this hook
+        return true;
+    }
+
+    /**
+     * Confirm that the maximum number of saved monitored trips has not been reached
+     */
+    private void reachedMaximum(String userId, Request request) {
```
evansiroky: Can we rename this to `verifyBelowMaxNumTrips`?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: This should be a single itinerary that is stored. When we check for differences we want to be checking against a single itinerary.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public List<Itinerary> itinerary;
+
+    //TODO, agree on and implement these parameters
+    /**
+    notificationThresholds
+    notifyRouteChange (true/false) - Instead of notificationThresholds
+    arrivalDelayMinutesThreshold (int minutes) - Instead of notificationThresholds
+    departureDelayMinutesThreshold (int minutes) - Instead of notificationThresholds
+    notifyDelayToTrip (true/false)
```
evansiroky, line 99: For now, I think it's best to keep these to a minimum and specific fields similar to these recommended as class members.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -16,15 +15,22 @@
...
-     * TODO: Replace with use of objectmapper?
      */
     public static String toJson(Object object) {
-        return gson.toJson(object);
+        try {
+            return mapper.writeValueAsString(object);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
+        }
+        return "{}";
```
evansiroky: If an error happens, shouldn't we return an error in JSON?

evansiroky (changes requested):
Just a few small comments. I'll approve once we figure out how to store notification trigger settings.

evansiroky:
Adding the blocked tag as this PR depends on #17 and #15.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: @landonreed @binh-dam-ibigroup can we decide what is needed so I can update please?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
br648: Updated, but not sure on syntax. Will need reviewing.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -16,15 +15,26 @@
 public class JsonUtils {
```
binh-dam-ibigroup, line 15: `HttpResponse` import is unused.

configurations/default/env.yml.tmp
```diff
@@ -10,4 +10,5 @@ MONGO_DB_NAME: otp_middleware
 #MONGO_HOST: cluster1.mongodb.net
 #MONGO_PROTOCOL: mongodb+srv
 OTP_SERVER: http://otp-server.example.com
-OTP_SERVER_PLAN_END_POINT: /otp/routers/default/plan
\ No newline at end of file
+OTP_SERVER_PLAN_END_POINT: /otp/routers/default/plan
+MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
```
binh-dam-ibigroup, line 15: Is this parameter required? In `Main.java`, there is something that assigns a default value, however that is not taking effect when `MAXIMUM_PERMITTED_MONITORED_TRIPS` is not set in the configuration.

src/main/java/org/opentripplanner/middleware/spark/Main.java
```diff
@@ -189,4 +187,22 @@ public static String getConfigPropertyAsText(String name, String defaultValue) {
...
+     * @return a config value (nested fields defined by dot notation "data.use_s3_storage") as an int or the default
+     * value if the config value is not defined (null) or cannot be converted to an int.
+     */
+    public static int getConfigPropertyAsInt(String name, int defaultValue) {
+
+        int value = defaultValue;
+
+        try {
+            JsonNode node = getConfigProperty(name);
+            value = Integer.parseInt(node.asText());
```
binh-dam-ibigroup, line 210: If `MAXIMUM_PERMITTED_MONITORED_TRIPS` is not defined in the configuration, I think this should simply use the default value assigned above.

binh-dam-ibigroup: If `MAXIMUM_PERMITTED_MONITORED_TRIPS` is not defined in the configuration, a `NullPointerException` is triggered that should be caught.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -25,12 +25,16 @@
...
+        String message;
         try {
             return mapper.writeValueAsString(object);
         } catch (JsonProcessingException e) {
             e.printStackTrace();
+            message = e.getMessage();
         }
-        return "{}";
+
+        return String.format("{Error: Unable to serialize object: %s. %s.", object.toString(), message);
```
evansiroky: This is invalid JSON. Ideally, we should just make a call to `JsonUtils#formatJSON`. See https://github.com/ibi-group/otp-middleware/blob/a323b679905673d781a19358acf679baad4ac3dd/src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java#L98-L100

evansiroky (changes requested):
See comment about JSON.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
binh-dam-ibigroup: We have not decided yet. I think you can leave as is because there is the TODO reminder for that.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,111 @@
+package org.opentripplanner.middleware.models;
+
+import org.opentripplanner.middleware.auth.Auth0UserProfile;
+import org.opentripplanner.middleware.auth.Permission;
+import org.opentripplanner.middleware.otp.response.Itinerary;
+
+import java.util.List;
```
binh-dam-ibigroup: Remove unused import.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.opentripplanner.middleware.auth.Auth0Connection.isAuthorized;
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsInt;
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+/**
+ * Implementation of the {@link ApiController} abstract class for managing monitored trips. This controller connects
+ * with Auth0 services using the hooks provided by {@link ApiController}.
+ */
+public class MonitorTripController extends ApiController<MonitoredTrip> {
```
binh-dam-ibigroup: Rename to `MonitoredTrip`

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final int MAXIMUM_PERMITTED_MONITORED_TRIPS
+        = getConfigPropertyAsInt("MAXIMUM_PERMITTED_MONITORED_TRIPS", 5);
+
+    public MonitorTripController(String apiPrefix) {
+        super(apiPrefix, Persistence.monitoredTrip, "secure/monitortrip");
```
binh-dam-ibigroup: Rename to `secure/monitoredtrip`

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -135,4 +135,7 @@ private static Date getDate(Request request, String paramName, String paramValue
                 .atZone(ZoneId.systemDefault())
                 .toInstant());
     }
+
```
binh-dam-ibigroup: Remove these spaces?

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
```diff
@@ -0,0 +1,60 @@
...
+import org.bson.conversions.Bson;
+import org.eclipse.jetty.http.HttpStatus;
+import org.opentripplanner.middleware.models.MonitoredTrip;
+import org.opentripplanner.middleware.persistence.Persistence;
+import spark.Request;
+
+import static com.mongodb.client.model.Filters.eq;
+import static org.opentripplanner.middleware.auth.Auth0Connection.isAuthorized;
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsInt;
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
```
binh-dam-ibigroup: This import appears unused.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -1,6 +1,9 @@
 package org.opentripplanner.middleware.otp.response;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
```
binh-dam-ibigroup: Unused import.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -1,6 +1,9 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.ArrayList;
```
binh-dam-ibigroup: Unused import.

binh-dam-ibigroup (changes requested):
Please advise on whether the trip limit config parameter should be required or optional. Rename a controller, and remove some unused imports.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
br648: Updated.

configurations/default/env.yml.tmp
br648: @binh-dam-ibigroup yes this parameter is required. I have updated the method to use the default value if the parameter is missing from the configuration.

src/main/java/org/opentripplanner/middleware/spark/Main.java
br648: Updated. Good spot.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
br648: Removed.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
br648: Removed.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
br648: Removed.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitorTripController.java
br648: Updated.

br648:
> 
> 
> Please advise on whether the trip limit config parameter should be required or optional. Rename a controller, and remove some unused imports.

@binh-dam-ibigroup MAXIMUM_PERMITTED_MONITORED_TRIPS is required. I have updated the code to take this into account. I've removed all unused imports from the project.

binh-dam-ibigroup (dismissed):
Good work @br648 !

evansiroky (dismissed):

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -140,18 +145,35 @@ public void bind(final SparkSwagger restApi) {
...
+        Auth0UserProfile requestingUser = getUserFromRequest(req);
+        if (isUserAdmin(requestingUser)) {
+            return persistence.getAll();
+        } else if (persistence.clazz == OtpUser.class) { 
+            Bson filter = Filters.eq("_id", requestingUser.otpUser.id);
+            return persistence.getFiltered(filter);
+        } else {
+            Bson filter = Filters.eq("userId", requestingUser.otpUser.id);
+            return persistence.getFiltered(filter);
+        }
```
landonreed, line 176: I would like to see some comments accompanying this block of conditionals describing the intent. It may be obvious to all of us at this moment, but future readers may not fully understand the context for the logic we've embedded here. E.g., "If user is an admin, the context is presumed to be the admin dashboard, so we deliver all entities (e.g., monitored trips) for management or review" etc...

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -140,18 +145,35 @@ public void bind(final SparkSwagger restApi) {
...
 
     /**
      * HTTP endpoint to get one entity specified by ID.
      */
     private T getOne(Request req, Response res) {
+        Auth0UserProfile requestingUser = Auth0Connection.getUserFromRequest(req);
         String id = getIdFromRequest(req);
-        return getObjectForId(req, id);
+        T object = getObjectForId(req, id);
+        if (!object.canBeManagedBy(requestingUser)) {
```
landonreed, line 198: I think the name `canBeManagedBy` implies that the requesting user can update or delete the entity. I'm not sure if it applies to viewing the entity. Do we need a separate method for `getOne`?

src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
```diff
@@ -74,7 +75,8 @@ public static void initialize () {
         apiUsers = new TypedPersistence(mongoDatabase, ApiUser.class);
         tripRequests = new TypedPersistence(mongoDatabase, TripRequest.class);
         tripSummaries = new TypedPersistence(mongoDatabase, TripSummary.class);
-        // TODO Add other models...
+        monitoredTrip = new TypedPersistence(mongoDatabase, MonitoredTrip.class);
```
landonreed: Please pluralize to match other collections.

src/main/java/org/opentripplanner/middleware/spark/Main.java
```diff
@@ -7,11 +7,8 @@
 import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.BasicOtpDispatcher;
 import org.opentripplanner.middleware.auth.Auth0Connection;
-import org.opentripplanner.middleware.controllers.api.AdminUserController;
-import org.opentripplanner.middleware.controllers.api.ApiUserController;
-import org.opentripplanner.middleware.controllers.api.TripHistoryController;
+import org.opentripplanner.middleware.controllers.api.*;
```
landonreed: I think there are a few places where wildcard imports have been substituted for single imports. Could you change this back? This setting needs to be modified in your IDE settings (I think you can change the threshold for switching to wildcard to 9999 or something like that).

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -83,6 +86,63 @@ public static void deleteTripRequests(List<TripRequest> tripRequests) {
...
+        days.add("friday");
+
+        MonitoredTrip monitoredTrip = new MonitoredTrip();
+        monitoredTrip.userId = userId;
+        monitoredTrip.tripName = "Commute to work";
+        monitoredTrip.tripTime = "07:30";
+        monitoredTrip.leadTimeInMinutes = 30;
+        monitoredTrip.days = days;
+        monitoredTrip.excludeFederalHolidays = true;
+        monitoredTrip.queryParams = "userId=b46266f7-a461-421b-8e92-01d99b945ab0&fromPlace=28.54894%2C%20-81.38971%3A%3A28.548944048426772%2C-81.38970606029034&toPlace=28.53989%2C%20-81.37728%3A%3A28.539893820446867%2C-81.37727737426759&date=2020-05-05&time=12%3A04&arriveBy=false&mode=WALK%2CBUS%2CRAIL&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&companies=";
```
landonreed: userId should no longer be in query params since we switched to using the Auth token, right?

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    MonitoredTrip preUpdateHook(MonitoredTrip monitoredTrip, MonitoredTrip preExisting, Request req) {
+        isAuthorized(monitoredTrip.userId, req);
+        return monitoredTrip;
+    }
+
+    @Override
+    boolean preDeleteHook(MonitoredTrip monitoredTrip, Request req) {
+        // Authorization checks are done prior to this hook
```
landonreed, line 43: It seems a little off that we would handle authorization differently for delete vs. update/create. Also, I noticed a potential bug in the isAuthorized check. I think it's possible that Auth0UserProfile could have both an apiUser and an otpUser. The way the check works now, I think it will throw a halt if (e.g.) the otpUser.id matches the monitored trip userId, but the apiUser.id does not.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,110 @@
...
+     * Enums have not been used because they are not compatible with MongoDB at this point in time. Java's number
+     * representation hasn't been used because it is irrelevant to calling systems.
+     *
+     * Expecting each effected day no more than once with the following representation:
+     *
+     * "monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"
+     *
+     * All use of this parameter will require equalsIgnoreCase() to avoid missing days.
+     */
+    public Set<String> days;
```
landonreed: I think it might be more appropriate to have boolean fields for each day. @evansiroky?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Confirm that the user updating a monitored trip has the required permissions
+     */
+    @Override
+    public boolean canBeManagedBy(Auth0UserProfile user) {
+        // If the user is attempting to update someone else's monitored trip, they must be admin.
+        boolean isManagingSelf = false;
+
+        if (user.otpUser != null) {
+            isManagingSelf = userId.equalsIgnoreCase(user.otpUser.id);
```
landonreed: Not sure we should be using ignore case.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -0,0 +1,110 @@
...
+    public MonitoredTrip() {
+    }
+
+    /**
+     * Confirm that the user updating a monitored trip has the required permissions
+     */
+    @Override
+    public boolean canBeManagedBy(Auth0UserProfile user) {
+        // If the user is attempting to update someone else's monitored trip, they must be admin.
+        boolean isManagingSelf = false;
```
landonreed: Variable name feels awkward. What about: `belongsToUser`?

landonreed (changes requested):
A few changes for you, but otherwise looking good!

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: updated

src/main/java/org/opentripplanner/middleware/persistence/Persistence.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: @landonreed I think you are right. I'm not sure how a separate method that would tie-in with the controller 'get one entity' endpoint. Perhaps the way forward is along the same lines as getMany?

src/main/java/org/opentripplanner/middleware/spark/Main.java
br648: Updated IDE and removed wildcard imports

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: I've updated to boolean fields for each day.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
br648: We would need to decide if the delete auth is to match the update/create auth... as all controllers will need to be updated to account for this.

I have separated the otpUser and apiUser userid checks.

br648:
@landonreed A couple of points which I need feedback on before completing these review.

binh-dam-ibigroup (dismissed):
LGTM

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
landonreed: Yea, I think you're right. If the object can be managed by the user, the user is obviously authorized. Else, the requesting userId must match the userId attached to the requested object.

br648: @landonreed - Thanks. Will update...

br648: @landonreed - After testing, it is the responsibility of the overriding class to define access to it's object. I have updated the method comment to clarify this. Therefore, I don't think the method should be changed.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -162,26 +161,38 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
...
         if (isUserAdmin(requestingUser)) {
+            // If the user is admin, the context is presumed to be the admin dashboard, so we deliver all entities for
+            // management or review without restriction.
             return persistence.getAll();
-        } else if (persistence.clazz == OtpUser.class) { 
+        } else if (persistence.clazz == OtpUser.class) {
+            // If the required entity is of type 'OtpUser' the assumption is that a call is being made via the
+            // OtpUserController. Therefore, the request should be limited to return just the entity matching the
+            // requesting user.
             Bson filter = Filters.eq("_id", requestingUser.otpUser.id);
```
landonreed: The filter should be on the field `userId`, right? Perhaps this should be factored into a method (it's used below and there may be other occurrences as well.

landonreed (dismissed):
Changes look good! I think it just needs one more tweak before merging (see [comment](https://github.com/ibi-group/otp-middleware/pull/21/files/4d29ba4735acc1847c790df5e145e4c8917fe4eb..76baee3085dda2f1d971897a18353c25bf8e068e#r450409161))

binh-dam-ibigroup (dismissed):

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: @landonreed  - "_id" is correct. The filter works on the columns used in Mongo. For an OTP user, the userId is held under the _id column. I have added a method to return the objects based on the these filter requirements.

br648:
@binh-dam-ibigroup My latest commit dismissed your previous review and I think approval?

binh-dam-ibigroup:
> @binh-dam-ibigroup My latest commit dismissed your previous review and I think approval?

@br648 Still stands after your last commit.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
landonreed: Right, ok. I think my confusion was around the fact that we're requesting "many" but filtering on the "_id" field, which by design should only return one result.

landonreed (dismissed):

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -4,27 +4,37 @@
...
+
 import static spark.Spark.halt;
 
 public class JsonUtils {
 
     private static final Logger LOG = LoggerFactory.getLogger(JsonUtils.class);
-    private static Gson gson = new Gson();
+
+    // Gson is not suitable for parsing OTP itineraries (errors on parsing Dates).
+    // private static Gson gson = new Gson();
```
evansiroky: Let's remove this comment then.

src/test/java/org/opentripplanner/middleware/persistence/MonitoredTripPersistenceTest.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.opentripplanner.middleware.persistence.PersistenceUtil.createMonitoredTrip;
+
+/**
+ * Tests to verify that monitored trip persistence in MongoDB collections are functioning properly. A
+ * number of {@link TypedPersistence} methods are tested here, but the HTTP endpoints defined in
+ * {@link org.opentripplanner.middleware.controllers.api.ApiController} are not themselves tested here.
+ */
+public class MonitoredTripPersistenceTest  extends OtpMiddlewareTest {
```
evansiroky: There's an extra space here before `extends`.

evansiroky (dismissed):
Good to go after addressing my 2 comments.

evansiroky (dismissed):
I just fixed a merge conflict and pushed it, so I'm reapproving given my original request to address my 2 comments.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/persistence/MonitoredTripPersistenceTest.java
br648: Updated.

PR #22, opened 2020-05-28
Get user by auth0 token

binh-dam-ibigroup:
### Checklist

### Description

The PR adds API route `/api/secure/user/fromtoken` to obtain users by Auth0 token, and removes route `/api/secure/{entity}/{find}/{value}` to restrict what can be queried.


src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -67,4 +95,44 @@ boolean preDeleteHook(User user, Request req) {
...
+
+        if (result.user == null) {
+            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, result.message,null);
+        }
+        return result.user;
+    }
+
+    /**
+     * @param profile The {@link Auth0UserProfile} from which to extract the User.
+     * @return An object containing the {@link User} entity from a {@link Auth0UserProfile}, or null and and an error message if that fails.
```
br648: Doulble 'and'

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -63,10 +62,21 @@ public ApiController (String apiPrefix, TypedPersistence<T> persistence, String
...
+            .withDescription("Interface for querying and managing '" + classToLowercase + "' entities."),
+            (q, a) -> LOG.info("Received request for '{}' Rest API", classToLowercase)
+        );
+        makeEndPoint(apiEndPoint);
+    }
 
-        restApi.endpoint(endpointPath(ROOT_ROUTE)
-            .withDescription("Interface for querying and managing '" + classToLowercase + "' entities."), (q, a) -> LOG.info("Received request for '{}' Rest API", classToLowercase))
+    /**
+     * Adds a basic implementation of the HTTP CRUD methods that can be overriden by child classes.
```
br648: Typo 'overriden' should be 'overridden'

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -67,4 +95,44 @@ boolean preDeleteHook(User user, Request req) {
...
+        public User user;
+        public String message;
+    }
+
+    /**
+     * HTTP endpoint to get the {@link User} entity from a {@link Auth0UserProfile}.
+     * (Reminder: for endpoints under 'secure', we add a {@link Auth0UserProfile} to request attributes.)
+     */
+    private User getUserFromRequest(Request req, Response res) {
+        Auth0UserProfile profile = req.attribute("user");
```
br648: I'm wondering if all actions related to user auth should be outside of the controllers as similar checks are likely to be done by other controllers. For instance, under the monitored-trip branch I have created: 

https://github.com/ibi-group/otp-middleware/blob/monitored-trip/src/main/java/org/opentripplanner/middleware/auth/Auth0Utils.java

br648 (commented):
I'm assuming once this is approved a similar update will be required to the monitor trip controller?

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
landonreed: So that everyone is aware, [`Auth0Connection#checkUser`](https://github.com/ibi-group/otp-middleware/blob/94dcefa65446530ba5854afb442aac7d416c5312/src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java#L38) checks that the user is authorized. This check is run on every request method [behind a `api/secure/*` or `api/admin/*` route](https://github.com/ibi-group/otp-middleware/blob/fced81463189960895df9cfa4eee26eef9637151/src/main/java/org/opentripplanner/middleware/spark/Main.java#L79-L86).

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -67,4 +95,44 @@ boolean preDeleteHook(User user, Request req) {
...
+    UserFromProfileResult getUserFromProfile(Auth0UserProfile profile) {
+        UserFromProfileResult result = new UserFromProfileResult();
+
+        if (profile != null) {
+            String auth0UserId = profile.user_id;
+            result.user = persistence.getOneFiltered(eq("auth0UserId", auth0UserId));
+            if (result.user == null) result.message = String.format(NO_USER_WITH_AUTH0_ID_MESSAGE, auth0UserId);
+        } else {
+            result.message = NO_AUTH0_PROFILE_MESSAGE;
+        }
```
landonreed: This else condition should never happen on an endpoint in this controller (see https://github.com/ibi-group/otp-middleware/pull/22/files#r432599004).

src/test/java/org/opentripplanner/middleware/controllers/api/UserControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+        UserController controller = new UserController("/api/");
+        UserController.UserFromProfileResult result = controller.getUserFromProfile(profile);
+
+        assertNotNull(result.user);
+        assertNull(result.message);
+        assertEquals(profile.user_id, result.user.auth0UserId);
+
+        // tidy up
+        Persistence.users.removeById(dbUser.id);
+    }
```
landonreed: These tests feel a little too contrived. If we're making controller tests, I would rather just send http requests to the API to verify the JSON response.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -63,10 +62,21 @@ public ApiController (String apiPrefix, TypedPersistence<T> persistence, String
      */
     @Override
     public void bind(final SparkSwagger restApi) {
-        LOG.info("Registering routes and enabling docs for {}", ROOT_ROUTE);
+        ApiEndpoint apiEndPoint = restApi.endpoint(endpointPath(ROOT_ROUTE)
+            .withDescription("Interface for querying and managing '" + classToLowercase + "' entities."),
+            (q, a) -> LOG.info("Received request for '{}' Rest API", classToLowercase)
+        );
+        makeEndPoint(apiEndPoint);
```
landonreed: Needs some comments.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
landonreed: This javadoc feels a little sparse. How about something like:

```
This method implements a set of basic HTTP Spark methods (e.g., getOne, getMany, delete) for CRUD operations. It can be overridden by child classes to supplement additional methods through the ApiEndpoint (after supplemental methods are added, be sure to call the super method to add the basic methods).
```

And now that I'm taking a closer look: should `apiEndpoint` just be a field on this class? And I don't think the `makeEndpoint` method needs to return the ApiEndpoint, right?

landonreed (changes requested):
Looks great, I do have a few suggestions though:
- changes for `makeEndpoint` signature/javadoc
- could we make the controller tests handle actual http requests to the user controller api endpoint?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Good catch, it is no longer necessary.
I had the return because I initially thought doing a chain add, e.g.
```java
return super.makeEndPoint()
.get(path(...))
.options(path(...))
```
but we have to call `super` after adding the custom routes. I'm updating the comments in this method accordingly.

binh-dam-ibigroup: It don't think `apiEndpoint` can become a field because the parent, `baseEndPoint`, is not available until spark-java calls the `bind` method.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
binh-dam-ibigroup: Hmm good point. We should consider making that profile a field?

src/test/java/org/opentripplanner/middleware/controllers/api/UserControllerTest.java
binh-dam-ibigroup: This unit test is an additional confidence check before going all in at the HTTP request level.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +81,40 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
+
+    /**
+     * @param profile The {@link Auth0UserProfile} from which to extract the User. Assumed not null.
+     * @return An object containing the {@link OtpUser} entity from a {@link Auth0UserProfile}, or null and an error message if that fails.
+     */
+    UserFromProfileResult getUserFromProfile(Auth0UserProfile profile) {
+        UserFromProfileResult result = new UserFromProfileResult();
+
+        String auth0UserId = profile.user_id;
+        result.user = persistence.getOneFiltered(eq("auth0UserId", auth0UserId));
```
br648: Auth0Connection#checkUser checks that the user is authorized prior to this endpoint being reached. Therefore the profile parameter already contains the OtpUser (or not) so this line is not required. The OtpUser can be obtained from profile.otpUser.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +81,40 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
+        if (result.user == null) result.message = String.format(NO_USER_WITH_AUTH0_ID_MESSAGE, auth0UserId);
```
br648: Brackets.

br648 (changes requested):
http://localhost:4567/api/secure/user/ returns a 404 "No API route configured for this path". My understanding is that this should be implemented as part of the ApiController.buildEndPoint call. It is listed under the Swagger UI endpoint (http://localhost:4567/) so I'm not sure why it fails.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
binh-dam-ibigroup: Thanks for pointing this one out, @br648, that simplifies *a lot*.

binh-dam-ibigroup:
> http://localhost:4567/api/secure/user/ returns a 404 "No API route configured for this path". My understanding is that this should be implemented as part of the ApiController.buildEndPoint call. It is listed under the Swagger UI endpoint (http://localhost:4567/) so I'm not sure why it fails.

@br648, it is either `http://localhost:4567/api/secure/user` or `http://localhost:4567/api/secure/user/{mongo_id}`.

binh-dam-ibigroup:
> * could we make the controller tests handle actual http requests to the user controller api endpoint?
@landonreed after simplification, that test file has been removed for the scope of the changes of this PR.

br648:
> 
> 
> > http://localhost:4567/api/secure/user/ returns a 404 "No API route configured for this path". My understanding is that this should be implemented as part of the ApiController.buildEndPoint call. It is listed under the Swagger UI endpoint (http://localhost:4567/) so I'm not sure why it fails.
> 
> @br648, it is either `http://localhost:4567/api/secure/user` or `http://localhost:4567/api/secure/user/{mongo_id}`.

Sorry, my mistake.

`http://localhost:4567/api/secure/user` does return all users, but this will be addressed in PR https://github.com/ibi-group/otp-middleware/pull/21

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +80,26 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
+        public OtpUser user;
+        public String message;
+    }
+
+    /**
+     * HTTP endpoint to get the {@link OtpUser} entity from a {@link Auth0UserProfile}.
+     * (Reminder: for endpoints under 'secure', we add a {@link Auth0UserProfile} to request attributes.)
+     */
+    private OtpUser getUserFromRequest(Request req, Response res) {
+        Auth0UserProfile profile = req.attribute("user");
```
br648: This should be replaced with Auth0Connection.getUserFromRequest(Request req);

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +80,26 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
         return true;
     }
+
+    /**
+     * Holds result and message from getUserFromProfile.
+     */
+    static class UserFromProfileResult {
+        public OtpUser user;
+        public String message;
+    }
```
br648: Can be deleted.

br648 (changes requested):
It is still possible to retrieve a user that doesn't match the authorized user:

http://localhost:4567/api/secure/user/841c1d8e-9ae1-4fef-9156-27cf77113a73-a-diff-user

I've addressed this issue for 'getMany' in https://github.com/ibi-group/otp-middleware/pull/21 I'm now wondering if I should make a similar update to 'getOne' and 'deleteOne'? @landonreed @binh-dam-ibigroup thoughts?

binh-dam-ibigroup:
> It is still possible to retrieve a user that doesn't match the authorized user:
> 
> http://localhost:4567/api/secure/user/841c1d8e-9ae1-4fef-9156-27cf77113a73-a-diff-user
> 
> I've addressed this issue for 'getMany' in #21 I'm now wondering if I should make a similar update to 'getOne' and 'deleteOne'? @landonreed @binh-dam-ibigroup thoughts?

@br648 How about the `canBeManagedBy` method in class `AbstractUser`? https://github.com/ibi-group/otp-middleware/blob/4b0cb1ed711fb7ec834563340d2eafb61abba4d0/src/main/java/org/opentripplanner/middleware/models/AbstractUser.java#L39

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
binh-dam-ibigroup: Good suggestion.

binh-dam-ibigroup:
@br648 `deleteOne` already has the check for whether a user can delete another user, so you could add the same check to `getOne`: https://github.com/ibi-group/otp-middleware/blob/4b0cb1ed711fb7ec834563340d2eafb61abba4d0/src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java#L152

br648:
> 
> 
> > It is still possible to retrieve a user that doesn't match the authorized user:
> > http://localhost:4567/api/secure/user/841c1d8e-9ae1-4fef-9156-27cf77113a73-a-diff-user
> > I've addressed this issue for 'getMany' in #21 I'm now wondering if I should make a similar update to 'getOne' and 'deleteOne'? @landonreed @binh-dam-ibigroup thoughts?
> 
> @br648 How about the `canBeManagedBy` method in class `AbstractUser`?
> 
> https://github.com/ibi-group/otp-middleware/blob/4b0cb1ed711fb7ec834563340d2eafb61abba4d0/src/main/java/org/opentripplanner/middleware/models/AbstractUser.java#L39

@binh-dam-ibigroup yes I think that will work.

binh-dam-ibigroup:
Latest changes pushed, @landonreed, @br648, @evansiroky.

br648:
> 
> 
> @br648 `deleteOne` already has the check for whether a user can delete another user, so you could add the same check to `getOne`:
> 
> https://github.com/ibi-group/otp-middleware/blob/4b0cb1ed711fb7ec834563340d2eafb61abba4d0/src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java#L152

@binh-dam-ibigroup thanks, yes will do that.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -60,22 +59,44 @@ public ApiController (String apiPrefix, TypedPersistence<T> persistence, String
     }
 
     /**
-     * This method is called by {@link SparkSwagger} to register endpoints and generate the docs.
+     * This method is called on each object deriving from ApiController by {@link SparkSwagger}
+     * to register endpoints and generate the docs.
+     * In this method, we add the different API paths and methods (e.g. the CRUD methods)
+     * to the restApi parameter for the applicable applicable controller.
```
evansiroky: typo: applicable applicable

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -60,22 +59,44 @@ public ApiController (String apiPrefix, TypedPersistence<T> persistence, String
...
      * @param restApi The object to which to attach the documentation.
      */
     @Override
     public void bind(final SparkSwagger restApi) {
-        LOG.info("Registering routes and enabling docs for {}", ROOT_ROUTE);
+        ApiEndpoint apiEndPoint = restApi.endpoint(endpointPath(ROOT_ROUTE)
+            .withDescription("Interface for querying and managing '" + classToLowercase + "' entities."),
```
evansiroky: Can this be formatted better so that the function arguments have the same offset from the start of the line?

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +81,18 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
+    /**
+     * HTTP endpoint to get the {@link OtpUser} entity, if it exists, from an {@link Auth0UserProfile} attribute
+     * available from a {@link Request} (this is the case for '/api/secure/' endpoints).
+     */
+    private OtpUser getUserFromRequest(Request req, Response res) {
+        Auth0UserProfile profile = Auth0Connection.getUserFromRequest(req);
+        OtpUser user = profile.otpUser;
+
+        if (user == null) {
+            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404,String.format(NO_USER_WITH_AUTH0_ID_MESSAGE, profile.user_id),null);
```
evansiroky: Please have better formatting here with spaces after commas. Also, the error string doesn't seem right because there will be an Auth0 user, just not an otpUser.

evansiroky (dismissed):
Just three little comments

landonreed (approved):
LGTM, once other comments are addressed.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
binh-dam-ibigroup: @evansiroky, A valid Auth0 user id does not guarantee existence of a corresponding `OtpUser` entity with the proper `auth0UserId` attribute, and the error case covers that. Example: First-time users who just signed up on Auth0 and have not entered their preferences.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
```diff
@@ -49,4 +81,18 @@ boolean preDeleteHook(OtpUser user, Request req) {
...
         return true;
     }
+
+    /**
+     * HTTP endpoint to get the {@link OtpUser} entity, if it exists, from an {@link Auth0UserProfile} attribute
+     * available from a {@link Request} (this is the case for '/api/secure/' endpoints).
+     */
+    private OtpUser getUserFromRequest(Request req, Response res) {
+        Auth0UserProfile profile = Auth0Connection.getUserFromRequest(req);
+        OtpUser user = profile.otpUser;
```
br648: Does this need to consider an apiUser and adminUser users?

br648 (commented):
Looks good, just the one comment. Once that has been address I can approve.

src/main/java/org/opentripplanner/middleware/controllers/api/UserController.java
binh-dam-ibigroup: `UserController` is for `OtpUser`s.

br648: Of course. Thanks for clarifying.

br648 (approved):
All good.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -16,10 +23,35 @@
...
+    static final String NO_USER_WITH_AUTH0_ID_MESSAGE = "No user with auth0UserID=%s found.";
+    private static final String TOKEN_PATH = "/fromtoken";
+
     public OtpUserController(String apiPrefix){
         super(apiPrefix, Persistence.otpUsers, "secure/user");
     }
 
+    @Override
+    protected void buildEndPoint(ApiEndpoint baseEndPoint) {
+        LOG.info("Registering user/fromtoken path.");
```
landonreed: Just noticed that this log statement doesn't make use of the `ROOT_ROUTE` + `TOKEN_PATH`.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -16,10 +23,35 @@
...
+
+        // Add the user token route BEFORE the regular CRUD methods
+        // (otherwise, /fromtoken requests would be considered
+        // by spark as 'GET user with id "fromtoken"', which we don't want).
+        ApiEndpoint modifiedEndpoint = baseEndPoint
+            // Get user from token.
+            .get(path(ROOT_ROUTE + TOKEN_PATH)
+                .withDescription("Retrieves a User entity (based on auth0UserId from request token).")
```
landonreed: This description looks to be incorrect.

binh-dam-ibigroup:
Latest changes noted by @evansiroky  and @landonreed have been incorporated. I'm proceeding to merge.

PR #24, opened 2020-06-08
OTP Request processing refactor

landonreed:
This PR makes a couple of suggestions to #17 for how to refactor the OTP request processing logic. I wasn't able to check how long the storage/auth check was taking, but I think it would be useful to take a look at that and include in #17.

### DIFF
--- a/src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
+++ b/src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
@@ -40,7 +40,7 @@ public class PersistenceUtil {
     public static TripSummary createTripSummary() {
         OtpDispatcherResponse response = getPlanFromOtp();
         Response otpResponse = response.getResponse();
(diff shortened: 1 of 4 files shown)


src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -52,7 +52,7 @@ public static TripSummary createTripSummaryWithError() {
         TripPlan tripPlan = otpResponse.getPlan();
         TripSummary tripSummary;
         if (tripPlan != null) {
-            tripSummary = new TripSummary(otpResponse.getPlan().from, otpResponse.getPlan().to, otpResponse.getError(), otpResponse.getPlan().itinerary, tripRequestId);
+            tripSummary = new TripSummary(otpResponse.getPlan(), otpResponse.getError(), tripRequestId);
```
binh-dam-ibigroup, line 55: `new TripSummary(tripPlan, ...)`

br648 (approved):
@landonreed - Thanks will merge. Regarding the storage/auth check time, is this something you would like saved? Or perhaps alerted on (via Bugsnag) if it is greater than a set threshold?

PR #25, opened 2020-06-08
fix(JsonUtils): Fix error parsing Dates in OTP Itinerary objects and …

binh-dam-ibigroup:
@br648 and @landonreed, we need to use Jackson to properly parse OTP `Itinerary` JSON objects and make the behavior consistent between `getPOJOFromRequestBody` and `getPOJOFromHttpResponse`. Feel free to merge and make additional tweaks for #21.


br648 (approved):
Looks good, will merge into #21. Thanks.

PR #38, opened 2020-07-16
feat(Reporting errors to Bugsnag): Update to report all exceptions and HTTP 500s to Bugsnag

br648:
### Checklist

### Description

The OTP middleware has been enhanced to include error reporting to Bugsnag. This covers points where exceptions are caught and HTTP status 500s are reported. No configuration changes are required, however, for this to be enabled a valid BUGSNAG_PROJECT_NOTIFIER_API_KEY must be provided. Details on how to obtain a BUGSNAG_PROJECT_NOTIFIER_API_KEY can be found in the README.

I've included unit tests, but these require a valid BUGSNAG_PROJECT_NOTIFIER_API_KEY in test\env.yml which is not permitted. Because of this, I have disabled the tests so as not to break Travis.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
+++ b/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
@@ -1,6 +1,9 @@
 package org.opentripplanner.middleware.bugsnag;
 
 import com.bugsnag.Bugsnag;
+import com.bugsnag.Report;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
 
@@ -13,26 +16,36 @@ import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
  */
 public class BugsnagReporter {
     private static Bugsnag bugsnag;
+    private static final Logger LOG = LoggerFactory.getLogger(BugsnagReporter.class);
 
     /**
-     * Initialize Bugsnag using API key when application is first loaded
+     * Initialize Bugsnag using the project notifier API key when the application is first loaded.
      */
-    private static void initializeBugsnag() {
+    public static void initializeBugsnagErrorReporting() {
         String apiKey = getConfigPropertyAsText("BUGSNAG_PROJECT_NOTIFIER_API_KEY");
         if (apiKey != null) {
             bugsnag = new Bugsnag(apiKey);
+        } else {
+            LOG.warn("Bugsnag project notifier API key not available. Bugsnag error reporting disabled.");
         }
     }
 
     /**
-     * Provide Bugsnag hook, if available, for reporting errors
+     * If Bugsnag has been configured, report error based on provided information.
      */
-    public static Bugsnag get() {
+    public static boolean reportErrorToBugsnag(Throwable throwable, String message) {
         if (bugsnag == null) {
-            initializeBugsnag();
+            LOG.warn("Bugsnag error reporting is disabled. Unable to report this message: {} ", message, throwable);
+            return false;
         }
-        return bugsnag;
+
+        if (throwable == null) {
+            LOG.warn("An exception is mandatory with Bugsnag error reporting. Unable to report this message: {} ", message);
+            return false;
+        }
+
+        Report report = bugsnag.buildReport(throwable);
+        report.setContext(message);
+        return bugsnag.notify(report);
     }
-
-
 }
--- a/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
+++ b/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
@@ -34,25 +34,33 @@ public class BugsnagReporter {
     /**
      * If Bugsnag has been configured, report error based on provided information.
      */
-    public static boolean reportErrorToBugsnag(String context, String message, Throwable throwable) {
+    public static boolean reportErrorToBugsnag(String message, Throwable throwable) {
+        return reportErrorToBugsnag(message, null, throwable);
+    }
+
+    /**
+     * If Bugsnag has been configured, report error based on provided information.
+     */
+    public static boolean reportErrorToBugsnag(String message, Object badEntity, Throwable throwable) {
+        LOG.error(message, throwable);
         if (bugsnag == null) {
             LOG.warn("Bugsnag error reporting is disabled. Unable to report to Bugsnag in this context: {}, this message: {}",
-                context,
                 message,
+                badEntity,
                 throwable);
             return false;
         }
 
         if (throwable == null) {
             LOG.warn("This error is not an exception and will not/cannot be reported to Bugsnag in this context: {}, this message: {}",
-                context,
-                message);
+                message,
+                badEntity);
             return false;
         }
 
         Report report = bugsnag.buildReport(throwable);
-        report.setContext(context);
-        report.setAppInfo("message", message);
+        report.setContext(message);
+        report.setAppInfo("entity", badEntity != null ? badEntity.toString() : "N/A");
         report.setSeverity(Severity.ERROR);
         return bugsnag.notify(report);
     }
--- a/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
+++ b/src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
@@ -44,7 +44,7 @@ public class BugsnagReporter {
     public static boolean reportErrorToBugsnag(String message, Object badEntity, Throwable throwable) {
         LOG.error(message, throwable);
         if (bugsnag == null) {
-            LOG.warn("Bugsnag error reporting is disabled. Unable to report to Bugsnag in this context: {}, this message: {}",
+            LOG.warn("Bugsnag error reporting is disabled. Unable to report to Bugsnag this message: {} for this bad entity: {}",
                 message,
                 badEntity,
                 throwable);
@@ -52,7 +52,7 @@ public class BugsnagReporter {
         }
 
         if (throwable == null) {
-            LOG.warn("This error is not an exception and will not/cannot be reported to Bugsnag in this context: {}, this message: {}",
+            LOG.warn("This error is not an exception and cannot be reported to Bugsnag. This message: {} for this bad entity: {}",
                 message,
                 badEntity);
             return false;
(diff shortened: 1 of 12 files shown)


src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -13,26 +16,36 @@
...
+            LOG.warn("Bugsnag project notifier API key not available. Bugsnag error reporting disabled.");
         }
     }
 
     /**
-     * Provide Bugsnag hook, if available, for reporting errors
+     * If Bugsnag has been configured, report error based on provided information.
      */
-    public static Bugsnag get() {
+    public static boolean reportErrorToBugsnag(Throwable throwable, String message) {
```
landonreed: I find it slightly more readable to have the string message be the first param (something about the longer string literal coming first or maybe because it mimics the Logger signature). Not a blocking change, but I think I would prefer that order.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -139,6 +140,7 @@ private static boolean saveTripRequest(TripRequest tripRequest) {
         } catch (MongoException e) {
             success = false;
             LOG.error("Unable to save trip request: " + tripRequest, e);
+            BugsnagReporter.reportErrorToBugsnag(e, "Unable to save trip request");
```
landonreed: Is there a convenient way to log the message inside the `reportErrorToBugsnag` method (ideally without losing the line number/stack trace to come back to)? Just because it's a bit redundant to have the same message repeated. Also, it seems valuable to include tripRequest details here as well (or was there a reason for not doing that?).

src/test/java/org/opentripplanner/middleware/BugsnagReportingTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test @Disabled
+    public void createErrorReportWithExceptionAndMessage() {
+        assertTrue(BugsnagReporter.reportErrorToBugsnag(new Exception("Unit test exception 1"), "Unit test 1"));
+    }
+
+    @Test @Disabled
+    public void createErrorReportWithException() {
+        assertTrue(BugsnagReporter.reportErrorToBugsnag(new Exception("Unit test exception 2"), null));
+    }
```
landonreed: Is it possible to create a test that triggers a report somewhere in the actual code (Persistence perhaps)?

landonreed (changes requested):
This looks good, but I wonder if we could add error logging within the reportError method to reduce duplicated messages? Maybe there's not really a great way to do that though...

I would also like to see a test that actually triggers a report within the code itself (rather than just the theoretical tests).

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -13,26 +16,36 @@
...
         if (bugsnag == null) {
-            initializeBugsnag();
+            LOG.warn("Bugsnag error reporting is disabled. Unable to report this message: {} ", message, throwable);
+            return false;
         }
-        return bugsnag;
-    }
 
+        if (throwable == null) {
+            LOG.warn("An exception is mandatory with Bugsnag error reporting. Unable to report this message: {} ", message);
```
binh-dam-ibigroup: What does this sentence mean?

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
binh-dam-ibigroup: +1 on the duplicate messages.

src/test/java/org/opentripplanner/middleware/BugsnagReportingTest.java
binh-dam-ibigroup: Would it make sense to create an `/makeerror` endpoint that generates a 500 error so we can catch it?

binh-dam-ibigroup (commented):
I concur with @landonreed on his comments, and that we need a way to trigger an error report.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
br648: This is misleading. It is actually the context, so Bugsnag groups errors by this value. I'm going to refactor as additional information is required.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Bugsnag cannot create a report if the exception is null, it causes a NPE. Hence logging this in case it is not provided/null.

src/test/java/org/opentripplanner/middleware/BugsnagReportingTest.java
br648: @landonreed if @binh-dam-ibigroup suggestion it ok, I will implement.

landonreed: Hmm, perhaps. I don't think we would really want this to be an open HTTP endpoint though (only accessible by Admin users). I'm kind of ambivalent here because this doesn't seem like something we really want to build into the non-test software, but it would be convenient to have a way to verify that things are configured correctly. @evansiroky?

br648: @landonreed - How about creating and documenting a test that attempts to insert a trip request into a stopped Mongo DB instance? No additional non-test code is created and the test is still under the context of a unit test.

landonreed: It will probably fail if you just try to insert a `null` TripRequest, right? That seems like an easy test to write up.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
binh-dam-ibigroup: Ok. I would change the text to something like "This error is not an exception and will not/cannot be reported to Bugsnag: {...}."

br648: Updated.

src/test/java/org/opentripplanner/middleware/BugsnagReportingTest.java
br648: @landonreed - I've made an update and created a unit test. As you will see this required additional code within the TypedPersistence -> create method and a less conventional unit test. Now that I have done this other methods in the TypedPersistence class could be updated in a similar way, but would like your view on this.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
br648: Updated to be context, message, exception. The message is saved as app info which is basically a key/value pair. It is possible to save a lot more information under app info, but to keep the Bugsnag reporting call to a minimum only a single item can be provided.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -86,8 +87,17 @@ public T create(String updateJson) {
...
+        try {
+            // TODO What happens if an object already exists with the same ID?
+            mongoCollection.insertOne(newObject);
+        } catch (Exception e) {
+            String objName = (newObject == null) ? "N/A" : newObject.getClass().getSimpleName();
+            String objContent = (newObject == null) ? "N/A" : newObject.toString();
+            String message = String.format("Unable to create object: %s with content: %s", objName, objContent);
+            LOG.error(message, e);
+            BugsnagReporter.reportErrorToBugsnag("Creating object", message, e);
+            throw e;
```
landonreed: I think this recent refactor of `reportErrorToBugsnag` perhaps should contain some of this logic. Instead of its signature being `String context, String message, Throwable e`, I think it would be more helpful to swap message for a generic `Object badEntity`. I think the easiest approach would be for me to just make a PR to this to demonstrate what I'm thinking. See #44.

landonreed:
Blocked until #44 is merged.

landonreed (approved):

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -33,19 +34,34 @@ public static void initializeBugsnagErrorReporting() {
...
+    /**
+     * If Bugsnag has been configured, report error based on provided information.
+     */
+    public static boolean reportErrorToBugsnag(String message, Object badEntity, Throwable throwable) {
+        LOG.error(message, throwable);
         if (bugsnag == null) {
-            LOG.warn("Bugsnag error reporting is disabled. Unable to report this message: {} ", message, throwable);
+            LOG.warn("Bugsnag error reporting is disabled. Unable to report to Bugsnag in this context: {}, this message: {}",
+                message,
+                badEntity,
```
binh-dam-ibigroup, line 49: Is the order for `message` and `context` swapped?

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -33,19 +34,34 @@ public static void initializeBugsnagErrorReporting() {
...
+                throwable);
             return false;
         }
 
         if (throwable == null) {
-            LOG.warn("An exception is mandatory with Bugsnag error reporting. Unable to report this message: {} ", message);
+            LOG.warn("This error is not an exception and will not/cannot be reported to Bugsnag in this context: {}, this message: {}",
+                message,
+                badEntity);
```
binh-dam-ibigroup, line 57: Is the order for `message` and `context` swapped?

binh-dam-ibigroup (commented):
Just two questions about the log messages when bugsnag is disabled.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
landonreed: @br648, when I modified the param names, I forgot to update these log messages. Should we just change the log messages to match the new param names?

br648: yep, updated.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
br648: Merged #44

binh-dam-ibigroup (approved):

PR #42, opened 2020-07-21
Get apikey fromaws

br648:
### Checklist

- [ ] Code coverage improves or is at 100% _(remove this if not merging to master)_

### Description

PR to create AWS API key and assign it to a usage plan. The API key id is then saved along with an API user. If the API user is deleted, the associated API key is deleted from AWS.

### To run

An AWS_PROFILE is required, or AWS access has been configured for your operating environment e.g. C:\Users\<username\.aws\credentials in Windows or iOS equivalent.

A DEFAULT_USAGE_PLAN_ID is required. This is needed so the newly created api key can be assigned to a usage plan. This can be obtained by logging into the AWS console and navagating to API gateway, usage plans. From here the desired usage plan can be selected and the id extracted.

### TODO
Add Bugsnag error reporting

### For consideration (see FIXME)

It appears many api key ids are expected, when only one is returned. I'm concerned that I might be creating a generic api key which allows access to all APIs... perhaps this is by design:

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-usage-plans.html

"You can use API keys together with usage plans or Lambda authorizers to control access to your APIs"

"API key values must be unique. If you try to create two API keys with different names and the same value, API Gateway considers them to be the same API key. "

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
+++ b/src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
@@ -1,0 +1,104 @@
+package org.opentripplanner.middleware.aws;
+
+import com.amazonaws.services.apigateway.AmazonApiGateway;
+import com.amazonaws.services.apigateway.AmazonApiGatewayClient;
+import com.amazonaws.services.apigateway.AmazonApiGatewayClientBuilder;
+import com.amazonaws.services.apigateway.model.CreateApiKeyRequest;
+import com.amazonaws.services.apigateway.model.CreateApiKeyResult;
+import com.amazonaws.services.apigateway.model.CreateUsagePlanKeyRequest;
+import com.amazonaws.services.apigateway.model.DeleteApiKeyRequest;
+import com.amazonaws.services.apigateway.model.GetUsagePlanRequest;
+import com.amazonaws.services.apigateway.model.GetUsagePlanResult;
+import com.amazonaws.services.apigateway.model.NotFoundException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
+
+/**
+ * Manages all interactions with AWS api gateway.
+ */
+public class AwsApiGateway {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AwsApiGateway.class);
+    private static final int SDK_REQUEST_TIMEOUT = 10 * 1000;
+    private static final String AWS_API_GATEWAY_USAGE_PLAN_ID
+        = getConfigPropertyAsText("AWS_API_GATEWAY_USAGE_PLAN_ID");
+
+    /**
+     * Create connection to AWS api gateway.
+     */
+    private static AmazonApiGateway getAmazonApiGateway() {
+        AmazonApiGatewayClientBuilder gatewayBuilder = AmazonApiGatewayClient.builder();
+        AmazonApiGateway gateway = gatewayBuilder.build();
+        return gateway;
+    }
+
+    /**
+     * Request an API key from AWS api gateway and assign it to an existing usage plan.
+     */
+    public static String getApiKeyId(String userId) {
+        long startTime = System.currentTimeMillis();
+
+        try {
+
+            AmazonApiGateway gateway = getAmazonApiGateway();
+
+            // create API key
+            CreateApiKeyRequest apiKeyRequest = new CreateApiKeyRequest();
+            apiKeyRequest.setSdkRequestTimeout(SDK_REQUEST_TIMEOUT);
+            apiKeyRequest
+                .withName(userId)
+                .withCustomerId(userId)
+                .withEnabled(true);
+            CreateApiKeyResult apiKeyResult = gateway.createApiKey(apiKeyRequest);
+
+            // get existing usage plan
+            GetUsagePlanRequest usagePlanRequest = new GetUsagePlanRequest();
+            usagePlanRequest.withUsagePlanId(AWS_API_GATEWAY_USAGE_PLAN_ID);
+            GetUsagePlanResult usagePlanResult = gateway.getUsagePlan(usagePlanRequest);
+
+            // add API key to usage plan
+            CreateUsagePlanKeyRequest usagePlanKeyRequest = new CreateUsagePlanKeyRequest();
+            usagePlanKeyRequest
+                .withUsagePlanId(usagePlanResult.getId())
+                .withKeyId(apiKeyResult.getId())
+                .withKeyType("API_KEY");
+            gateway.createUsagePlanKey(usagePlanKeyRequest);
+
+            return apiKeyResult.getId();
+        } catch (Exception e) {
+            LOG.error("Unable to get api key from AWS", e);
+            //FIXME add bugsnag
+        } finally {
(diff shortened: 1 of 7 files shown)


src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
```diff
@@ -0,0 +1,104 @@
...
+import com.amazonaws.services.apigateway.model.NotFoundException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
+
+/**
+ * Manages all interactions with AWS api gateway.
```
landonreed: The stuff in LogController should probably be refactored to use methods in here.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
```diff
@@ -0,0 +1,104 @@
...
+    private static AmazonApiGateway getAmazonApiGateway() {
+        AmazonApiGatewayClientBuilder gatewayBuilder = AmazonApiGatewayClient.builder();
+        AmazonApiGateway gateway = gatewayBuilder.build();
+        return gateway;
+    }
+
+    /**
+     * Request an API key from AWS api gateway and assign it to an existing usage plan.
+     */
+    public static String getApiKeyId(String userId) {
```
landonreed: I think this should probably be called `createApiKey` and return the entire key. We may want to do more with the key in ApiUserController after it's created.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -24,8 +29,32 @@ protected ApiUser getUserProfile(Auth0UserProfile profile) {
...
-        // TODO: Add the API keys for this user.
+        String apiKeyId = AwsApiGateway.getApiKeyId(user.id);
+        if (apiKeyId == null) {
+            logMessageAndHalt(req, HttpStatus.INTERNAL_SERVER_ERROR_500,
+                String.format("Unable to get AWS api key for user %s", user)
+                , null);
+        }
+
+        //FIXME This suggests many, where only one is returned?
+        user.apiKeyIds.add(apiKeyId);
```
landonreed: I don't think we've entirely figured out how the API keys will correspond to the various endpoints in the API, but we want to leave open the possibility that more than one API key can be given to an API user.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
```diff
@@ -0,0 +1,104 @@
...
+            apiKeyRequest.setSdkRequestTimeout(SDK_REQUEST_TIMEOUT);
+            apiKeyRequest
+                .withName(userId)
+                .withCustomerId(userId)
+                .withEnabled(true);
+            CreateApiKeyResult apiKeyResult = gateway.createApiKey(apiKeyRequest);
+
+            // get existing usage plan
+            GetUsagePlanRequest usagePlanRequest = new GetUsagePlanRequest();
+            usagePlanRequest.withUsagePlanId(AWS_API_GATEWAY_USAGE_PLAN_ID);
```
landonreed: I think perhaps this method should take a Usage Plan Id as a param for now and perhaps in the ApiUserController preCreateHook invocation, we can pass the DEFAULT_USAGE_PLAN_ID (which is derived from the config). I think we'll want to be able to manage/create keys for multiple usage plans, though how exactly that will work is not defined yet.

landonreed (changes requested):
Looks great, just a few comments.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
br648: @landonreed - Agree. I wasn't sure when this should happen so added an item to the Refactor tasks "Consolidate AWS SDK integration"

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Have removed comment.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
```diff
@@ -0,0 +1,104 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+import static org.opentripplanner.middleware.spark.Main.getConfigPropertyAsText;
+
+/**
+ * Manages all interactions with AWS api gateway.
+ */
+public class AwsApiGateway {
```
landonreed: Oh, one other thing. I think we would typically opt to place a class like this in the utils package with the class name `ApiGatewayUtils`. I think that can help with seeing the classname appear in the IDE... less chance of confusing this class with a class from the AWS library itself.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
br648: Stil LogController has been merged into dev I decided to update.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/aws/AwsApiGateway.java
br648: Updated.

landonreed (approved):
LGTM!

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -24,8 +34,32 @@ protected ApiUser getUserProfile(Auth0UserProfile profile) {
...
-        // TODO: Add the API keys for this user.
+        CreateApiKeyResult apiKey = ApiGatewayUtils.createApiKey(user.id, DEFAULT_USAGE_PLAN_ID);
+        if (apiKey == null) {
+            logMessageAndHalt(req,
+                HttpStatus.INTERNAL_SERVER_ERROR_500,
+                String.format("Unable to get AWS api key for user %s", user),
+                null);
+        }
+
+        user.apiKeyIds.add(apiKey.getId());
```
binh-dam-ibigroup, line 46: We should probably add a note that we are not storing the actual API keys.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Request an API key from AWS api gateway and assign it to an existing usage plan.
+     */
+    public static CreateApiKeyResult createApiKey(String userId, String usagePlanId) {
+        long startTime = System.currentTimeMillis();
+
+        try {
+
```
binh-dam-ibigroup: Remove extra blank line.

binh-dam-ibigroup (approved):
Looks good! Maybe add a comment on persistence that we store the key ids, not the actual keys.

PR #44, opened 2020-07-21
Refactor bugsnag reporter

landonreed:
Just a few changes for #38.

Also, @br648, now I'm wondering if we should just use this: https://docs.bugsnag.com/platforms/java/other/configuring-logback/


br648:
@landonreed I've added Bugsnag logback to the refactor tasks plan. 

For now, I'll merge for review. We can either decide to keep or bin in favour of logback.

PR #33, opened 2020-07-11
Trip monitoring MVP

evansiroky:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [ n/a ] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This is a draft PR for a minimum viable product for monitoring trips and sending out notifications.

#### Running:

Screen Shot 2020-07-11 at 12 44 50 AM

#### Configuration requirements:

- needs to have a Mongo db name set

PR #45, opened 2020-07-24
Check Monitored Trip

landonreed:
### Checklist

### Description

**NOTE:** This PR is based on #33 (and must be merged only after #33 is merged to dev). To review only the commits added by this PR:

**[:star::star:Use this commit filter:star::star:](https://github.com/ibi-group/otp-middleware/pull/45/files/53a936363d2194a3354a9519aae97085a1ecd53a..HEAD)**

Adds basic trip monitor functionality to handle checking whether a monitored trip meets any thresholds for notifying users.

Note: in order to run the disabled tests in `TripMonitorTest`, you will need to update test config with properties for the following services (in addition to changing the from email/phone to valid values):
https://github.com/ibi-group/otp-middleware/blob/e85e7be061d0fadb43e33ce1f4671951592088ee/configurations/default/env.yml.tmp#L31-L40


src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -29,6 +32,12 @@
      */
     public String tripTime;
 
+    /**
+     * Tracks information during the active monitoring of the trip (e.g., last alerts encountered, last time a check was
+     * made, etc.).
+     */
+    public JourneyState journeyState = new JourneyState();
```
evansiroky: Should the JourneyState be stored in it's own collection instead of being nested here? I'm pretty sure we don't want to expose the information to the end user or make it possible to update from a webservice.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/TripAnalyzer.java
```diff
@@ -0,0 +1,85 @@
...
+                    continue;
+                }
+
+                LOG.info("Analyzing trip {}", trip);
+
+                // place lock on trip
+                monitoredTripLocks.put(trip, true);
+
+                /////// BEGIN TRIP ANALYSIS
+                new CheckMonitoredTrip(trip).run();
```
evansiroky: Please surround with a try/catch (Exception e) to be extra sure that no error will occur here that ends up causing this job to stop executing.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * This class contains utils for sending SMS and email notifications.
+ */
+public class NotificationUtils {
+    // Find your Account Sid and Token at https://twilio.com/user/account
+    public static final String ACCOUNT_SID = getConfigPropertyAsText("TWILIO_ACCOUNT_SID");
+    public static final String AUTH_TOKEN = getConfigPropertyAsText("TWILIO_AUTH_TOKEN");
+    // From phone must be registered with Twilio account.
+    public static final String FROM_PHONE = getConfigPropertyAsText("NOTIFICATION_FROM_PHONE");
+    private static final Logger LOG = LoggerFactory.getLogger(NotificationUtils.class);
```
evansiroky: Place the `LOG` at the top before everything else.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
+
+/**
+ * This class contains utils for sending SMS and email notifications.
+ */
+public class NotificationUtils {
+    // Find your Account Sid and Token at https://twilio.com/user/account
+    public static final String ACCOUNT_SID = getConfigPropertyAsText("TWILIO_ACCOUNT_SID");
+    public static final String AUTH_TOKEN = getConfigPropertyAsText("TWILIO_AUTH_TOKEN");
```
evansiroky: Please prefix these with `TWILIO_...`

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Send a SMS message to the provided phone number
+     * @param toPhone - e.g., +15551234
+     * @param body - SMS message body
+     * @return messageId if message was sucessful (null otherwise)
+     */
+    public static String sendSMS(String toPhone, String body) {
+        if (ACCOUNT_SID == null || AUTH_TOKEN == null) {
+            LOG.warn("SMS notifications not configured correctly.");
```
evansiroky: Make this an error.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+            PhoneNumber fromPhoneNumber = new PhoneNumber(FROM_PHONE);
+            PhoneNumber toPhoneNumber = new PhoneNumber(toPhone);
+            Message message = Message.creator(
+                toPhoneNumber,
+                fromPhoneNumber,
+                body
+            ).create();
+            LOG.debug("SMS ({}) sent successfully", message.getSid());
+            return message.getSid();
+        } catch (Exception e) {
```
evansiroky, line 60: Possible to use a more specific Exception type here?

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param body - SMS message body
+     * @return messageId if message was sucessful (null otherwise)
+     */
+    public static String sendSMS(String toPhone, String body) {
+        if (ACCOUNT_SID == null || AUTH_TOKEN == null) {
+            LOG.warn("SMS notifications not configured correctly.");
+            return null;
+        }
+        try {
+            Twilio.init(ACCOUNT_SID, AUTH_TOKEN);
```
evansiroky: Seems like this could be initialized upon application load in a static block of code.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Send notification email using Sparkpost.
+     * TODO: determin if we should use sparkpost or sendgrid.
+     */
+    public static void sendEmail(String toEmail, String subject, String text, String html) {
+        if (SPARKPOST_KEY == null || FROM_EMAIL == null) {
+            LOG.warn("Notifications disabled due to invalid config. Skipping message to {} SUBJECT: {}", toEmail, subject);
+            return;
+        }
+        Client client = new Client(SPARKPOST_KEY);
```
evansiroky: Seems like this could be initialized upon application load in a static block of code.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        if (SENDGRID_API_KEY == null || FROM_EMAIL == null) {
+            LOG.warn("Notifications disabled due to invalid config. Skipping message to {} SUBJECT: {}", toEmail, subject);
+            return false;
+        }
+        Email from = new Email(FROM_EMAIL);
+        Email to = new Email(toEmail);
+        Content content = new Content("text/plain", text);
+        Mail mail = new Mail(from, subject, to, content);
+
+        SendGrid sg = new SendGrid(SENDGRID_API_KEY);
```
evansiroky, line 104: Seems like this could be initialized upon application load in a static block of code.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    /**
+     * Send notification email using Sparkpost.
+     * TODO: determin if we should use sparkpost or sendgrid.
+     */
+    public static void sendEmail(String toEmail, String subject, String text, String html) {
+        if (SPARKPOST_KEY == null || FROM_EMAIL == null) {
+            LOG.warn("Notifications disabled due to invalid config. Skipping message to {} SUBJECT: {}", toEmail, subject);
```
evansiroky: Should be an error message.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Send notification email using Sendgrid.
+     * TODO: determin if we should use sparkpost or sendgrid.
+     */
+    public static boolean sendSendGridEmail(String toEmail, String subject, String text, String html) {
+        String SENDGRID_API_KEY = getConfigPropertyAsText("SENDGRID_API_KEY");
+        if (SENDGRID_API_KEY == null || FROM_EMAIL == null) {
+            LOG.warn("Notifications disabled due to invalid config. Skipping message to {} SUBJECT: {}", toEmail, subject);
```
evansiroky: Should be an error message.

src/test/java/org/opentripplanner/middleware/trip_monitor/TripMonitorTest.java
```diff
@@ -0,0 +1,100 @@
...
+        simulatedItinerary.legs.get(1).alerts = fakeAlerts;
+        simulatedResponse.setResponse(otpResponse);
+        LOG.info("Created trip {}", monitoredTrip.id);
+        // Next, run a monitor trip check from the new monitored trip using the simulated response.
+        new CheckMonitoredTrip(monitoredTrip, simulatedResponse).run();
+        // Clear the created trip.
+        Persistence.monitoredTrips.removeById(monitoredTrip.id);
+    }
+
+    @Test @Disabled
```
evansiroky: Let's not disable these, but instead use the code used in datatools where we only run the test with an assumeTrue with an environment variable being set to something. This way we can have run configurations that can quickly enable these tests. Also, let's add a TODO to have some kind of integration test where live emails and texts are send and we verify that the were received by some third party service.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -128,5 +190,16 @@ public boolean canBeManagedBy(Auth0UserProfile user) {
         return super.canBeManagedBy(user);
     }
 
+    public Itinerary lastItinerary() {
+        if (journeyState.responses.size() > 0) {
+            return journeyState.responses.get(0).plan.itineraries.get(0);
+        }
+        return itinerary;
+    }
```
evansiroky: This seems to get the first itinerary. Please rename method and add some Javadoc.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CheckMonitoredTrip implements Runnable {
+    private static final Logger LOG = LoggerFactory.getLogger(CheckMonitoredTrip.class);
+    private final MonitoredTrip trip;
+    private OtpDispatcherResponse injectedOtpResponse;
```
evansiroky: Add some comments about this method and please rename the method to include some kind of indication that this variable is only used during testing.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -56,19 +82,14 @@ private static OtpDispatcherResponse sendOtpRequest(URI uri) {
...
-            otpDispatcherResponse = new OtpDispatcherResponse();
-            otpDispatcherResponse.responseBody = otpResponse.body();
-            otpDispatcherResponse.statusCode = otpResponse.statusCode();
-            LOG.debug("Response from OTP server: {}", otpDispatcherResponse.toString());
+            otpDispatcherResponse = new OtpDispatcherResponse(otpResponse);
         } catch (InterruptedException | IOException e) {
             LOG.error("Error requesting OTP data", e);
         }
-
         return otpDispatcherResponse;
```
evansiroky, line 93: This method can return null which can result in NPEs. I would prefer for this method to throw a non-runtime exception if the request failed so that code that uses this method can properly account for times when an OTP request fails.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
+package org.opentripplanner.middleware.trip_monitor;
```
evansiroky: This file should probably be in the jobs folder.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -61,6 +63,7 @@ public static void logMessageAndHalt(Request request, int statusCode, String mes
             return mapper.readValue(json, clazz);
         } catch (JsonProcessingException e) {
             LOG.error("Unable to get POJO from json for " + clazz.getSimpleName(), e);
+            // TODO: Bugsnag?
         }
         return null;
     }
```
evansiroky, line 69: Most of these methods in here should not return null and instead should throw the exception. This will avoid NPEs elsewhere.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -1,36 +1,75 @@
...
     public String responseBody = null;
 
     /**
      * Response. POJO version of response from an OTP server.
+     * Do not persist in case these classes change. This should always be re-instantiated from responseBody if needed.
      */
-    public Response response = null;
+    public Response getResponse() {
+        return JsonUtils.getPOJOFromJSON(responseBody, Response.class);
+    }
```
evansiroky, line 60: This method can return null.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+        OtpDispatcherResponse otpDispatcherResponse = injectedOtpResponse != null
+            ? injectedOtpResponse
+            : OtpDispatcher.sendOtpPlanRequest(trip.queryParams);
+
+        if (otpDispatcherResponse.statusCode >= 400) {
+            // TODO: report bugsnag
+            LOG.error("Could not reach OTP server. status={}", otpDispatcherResponse.statusCode);
+            return;
+        }
+        Response otpResponse = otpDispatcherResponse.getResponse();
```
evansiroky: `otpResponse` can be null.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+        // Check for notifications related to itinerary changes
+        checkTripForItineraryChanges(trip, itinerary);
+        // Add latest OTP response.
+        trip.addResponse(otpResponse);
+        Persistence.monitoredTrips.replace(trip.id, trip);
+        // Send notification to user.
+        sendNotification();
+    }
+
+    private void checkTripForItineraryChanges(MonitoredTrip trip, Itinerary itinerary) {
```
evansiroky: Can the next 4 methods be in the order that they are called in the `run` method?

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+            LOG.error("Cannot find user for id {}", trip.userId);
+            // TODO: Bugsnag / delete monitored trip?
+            return;
+        }
+        String name = trip.tripName != null ? trip.tripName : "Trip for " + otpUser.email;
+        String subject = name + " Notification";
+        StringBuilder body = new StringBuilder();
+        for (TripMonitorNotification notification : notifications) {
+            body.append(notification.body);
+        }
```
evansiroky: Ideally we'll have different text that gets sent out via sms/email, but I guess this is fine for now.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+            for (LocalizedAlert newAlert : newAlerts) {
+                boolean alertIsNew = true;
+                for (LocalizedAlert previousAlert : previousAlerts) {
+                    if (newAlert.equals(previousAlert)) {
+                        alertIsNew = false;
+                        break;
+                    }
+                }
+                if (alertIsNew) unseenAlerts.add(newAlert);
+            }
```
evansiroky: Seems like you could make a Set of LocalizedAlerts and then just call `removeAll` to have the same effect as this code.

src/main/java/org/opentripplanner/middleware/otp/response/LocalizedAlert.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Date;
+
+public class LocalizedAlert {
+    public String alertHeaderText;
+    public String alertDescriptionText;
+    public String alertUrl;
+    public Date effectiveStartDate;
+
+    public boolean equals(Object o) {
```
evansiroky, line 17: A hashCode can also be generated here to help facilitate checking these objects in a HashSet as recommended in https://github.com/ibi-group/otp-middleware/pull/45/files#r463329825.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+            if (unseenAlerts.size() > 0) {
+                // FIXME: May need to update the user about alerts that have dropped off as well.
+                enqueueNotification(TripMonitorNotification.createAlertNotification(unseenAlerts));
+            }
+        } else {
+            // If there are new alerts present in the response, create notification.
+            if (newAlerts.size() > 0) {
+                enqueueNotification(TripMonitorNotification.createAlertNotification(newAlerts));
+            }
+        }
```
evansiroky: This doesn't check for Alerts that existed in the previous itinerary that now no longer exist (aka giving an all-clear notification). I suppose that's ok for now, but maybe make a comment about it.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+
+        if (otpDispatcherResponse.statusCode >= 400) {
+            // TODO: report bugsnag
+            LOG.error("Could not reach OTP server. status={}", otpDispatcherResponse.statusCode);
+            return;
+        }
+        Response otpResponse = otpDispatcherResponse.getResponse();
+        // TODO: Should null tripRequestId be fixed?
+        // TripSummary tripSummary = new TripSummary(otpDispatcherResponse.response.plan, otpDispatcherResponse.response.error, null);
+        // TODO: Find the specific itinerary to compare against. For now, just choose the first itin.
```
evansiroky: This is a big, very important TODO. At the very least create a stubbed method to determine if the OTP response contains the exact saved itinerary that make the rest of these checks possible.

src/main/java/org/opentripplanner/middleware/utils/DateUtils.java
```diff
@@ -23,27 +23,29 @@
...
+            throw e;
+        }
+    }
 
+    public static LocalDate getDateFromString(String value, String expectedDatePattern) throws DateTimeParseException {
         // no date value to work with
-        if (paramValue == null) {
+        if (value == null) {
             return null;
         }
```
evansiroky: Don't return null here. Instead throw a new `DateTimeParseException`.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+            }
+        }
+    }
+
+    private void enqueueNotification(TripMonitorNotification tripMonitorNotification) {
+        notifications.add(tripMonitorNotification);
+    }
+
+    private boolean shouldSkipMonitoredTripCheck(MonitoredTrip trip) {
+        LocalDate now = LocalDate.now();
```
evansiroky: This LocalDate is dependent on the system timezone. To future proof this, I think the most accurate way of doing this is to do a timezone lookup of the coordinates of the target trip end (the origin if the trip is a depart at trip, or the destination if the trip is an arrive by trip).

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,176 @@
...
+            tripTime = DateUtils.getDateFromString(trip.tripTime, "HH:mm");
+        } catch (DateTimeParseException e) {
+            // TODO: Bugsnag report
+            LOG.error("Could not parse trip time", e);
+            return false;
+        }
+        // Trip is not active today.
+        return !trip.isActiveOnDate(now) &&
+            // Trip has already occurred.
+            tripTime.isAfter(now);
```
evansiroky: This will return true for the whole day. So trips that happen at like 10pm will be checked throughout the day. There should probably be some more checks in here.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
landonreed: The idea is that it gets the most recently stored itinerary. Maybe latest is a better name.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
landonreed: This seems out of scope for this PR. Will create an issue for this.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
landonreed: Again, I think this is out of scope for this MVP PR. I will add to #51.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
landonreed: There was a comment above (`FIXME: May need to update the user about alerts that have dropped off as well. `). But I have fixed this in recent changes.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
landonreed: This is an MVP, so I'm just going to add your comment in a FIXME.

src/main/java/org/opentripplanner/middleware/trip_monitor/CheckMonitoredTrip.java
landonreed: Do you mean this field (and are you referring to `injectedOtpResponse`)? Or did you mean a method somewhere else?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
landonreed: Perhaps. We definitely don't want to share this with the end user.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+        notifications.add(tripMonitorNotification);
+    }
+
+    private boolean shouldSkipMonitoredTripCheck(MonitoredTrip trip) {
+        // Default to system's zone if one can't be inferred from trip destination.
+        ZoneId zoneId = ZoneId.systemDefault();
+        // TODO: Verify that this check doesn't cost anything.
+        Optional<ZoneId> zoneIdForCoordinates = getZoneIdForCoordinates(trip.to.lat, trip.to.lon);
+        if (zoneIdForCoordinates.isEmpty()) {
+            LOG.error("Could not find timezone for monitored trip: {}", trip.id);
```
evansiroky: If the timezone can't be obtained from GPS coordinates, something really funky is going on especially since the `getZoneIdForCoordinates` should have worldwide coverage. But if it happens, I think it doesn't make sense to use the default system timezone. Instead, an error should be thrown.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+        } else {
+            zoneId = zoneIdForCoordinates.get();
+        }
+        // Get current time and trip time for comparison.
+        LocalDate now = LocalDate.now(zoneId);
+        LocalDate tripTime = trip.itinerary.startTime.toInstant()
+            .atZone(zoneId)
+            .toLocalDate();
```
evansiroky: This seems like it will constantly check the original startTime in the itinerary and not update to a new date + time each check.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+        long millisSinceLastCheck = System.currentTimeMillis() - trip.retrieveJourneyState().lastChecked;
+        long minutesSinceLastCheck = TimeUnit.MILLISECONDS.toMinutes(millisSinceLastCheck);
+        long minutesUntilTrip = Duration.between(tripTime.atStartOfDay(), now.atStartOfDay()).toMinutes();
+        // TODO: Refine these frequency intervals for monitor checks.
+        // If time until trip is greater than 60 minutes, we only need to check once every hour.
+        if (minutesUntilTrip > 60) {
+            // It's been about an hour since the last check. Do not skip.
+            if (minutesSinceLastCheck <= 60) {
+                return false;
+            }
```
evansiroky: Seems like this will result in checking the trip every time.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+            // During the last 5 minutes, check every minute (assuming the loop runs every minute).
+            if (minutesUntilTrip <= 5) {
+                return false;
+            }
+        }
+        // TODO: Check that journey state is not flagged?
+        // Default to skipping check.
+        return true;
```
evansiroky, line 235: Should this use the `MonitoredTrip.leadTimeInMinutes` field at all?

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+            return;
+        }
+        // FIXME: Need to somehow determine how to get last itinerary from responses in journey state.
+        Itinerary latestItinerary = trip.latestItinerary();
+        // Construct set from new alerts.
+        Set<LocalizedAlert> newAlerts = new HashSet<>(itinerary.getAlerts());
+        Set<LocalizedAlert> previousAlerts = new HashSet<>(latestItinerary.getAlerts());
+        // Unseen alerts consists of all new alerts that we did not previously track.
+        Set<LocalizedAlert> unseenAlerts = new HashSet<>(newAlerts);
+        unseenAlerts.addAll(previousAlerts);
```
evansiroky: I think this should be a `removeAll` operation instead of `addAll`.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/TripMonitorNotification.java
```diff
@@ -0,0 +1,38 @@
...
+import org.opentripplanner.middleware.otp.response.LocalizedAlert;
+
+import java.util.Collection;
+import java.util.List;
+
+public class TripMonitorNotification {
+    public NotificationType type;
+    public String body;
+
+    public static TripMonitorNotification createAlertNotification(Collection<LocalizedAlert> newAlerts, Collection<LocalizedAlert> resolvedAlerts) {
```
evansiroky: Long line, please spread across multiple lines.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/TripMonitorNotification.java
```diff
@@ -0,0 +1,38 @@
...
+        TripMonitorNotification notification = new TripMonitorNotification();
+        notification.type = NotificationType.ALERT_FOUND;
+        notification.body = bodyFromAlerts(newAlerts, resolvedAlerts);
+        return notification;
+    }
+
+    private static String bodyFromAlerts(Collection<LocalizedAlert> newAlerts, Collection<LocalizedAlert> resolvedAlerts) {
```
evansiroky: Long line, please spread across multiple lines.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -128,5 +199,25 @@ public boolean canBeManagedBy(Auth0UserProfile user) {
...
+
+    public void addResponse(Response response) {
+        JourneyState journeyState = retrieveJourneyState();
+        journeyState.lastChecked = System.currentTimeMillis();
+        journeyState.responses.add(0, response);
```
evansiroky, line 236: Seems like this will endlessly add OTP responses to this list. I don't see anywhere else that removes from this list.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private void enqueueNotification(TripMonitorNotification tripMonitorNotification) {
+        notifications.add(tripMonitorNotification);
+    }
+
+    private boolean shouldSkipMonitoredTripCheck(MonitoredTrip trip) {
+        // Default to system's zone if one can't be inferred from trip destination.
+        ZoneId zoneId = ZoneId.systemDefault();
+        // TODO: Verify that this check doesn't cost anything.
```
evansiroky: If we're concerned about the computation cost, this could be ran while saving a monitored trip.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+        Optional<ZoneId> zoneIdForCoordinates = getZoneIdForCoordinates(trip.to.lat, trip.to.lon);
```
evansiroky: The lat/lon should be at the trip's from location if it is a departAt type of monitored trip.

evansiroky: Oh interesting, I guess you're using the end coordinates to make sure we check up until the trip ends regardless of whether it's arrive by or depart at. I like this approach.

evansiroky: Although, now that I think about it more, maybe we want to use a combination of start and arrive by times to make sure we're checking early enough before the trip starts and all the way until the trip ends.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+            default:
+                break;
+        }
+    }
+
+    private void enqueueNotification(TripMonitorNotification tripMonitorNotification) {
+        notifications.add(tripMonitorNotification);
+    }
+
+    private boolean shouldSkipMonitoredTripCheck(MonitoredTrip trip) {
```
evansiroky: I think it'd be useful to make some unit tests for this method to make sure various use cases are handled properly. [See this stackoverflow answer to see an example of how to mock time in java](https://stackoverflow.com/a/29360514/269834). Some test cases I can think of would be:

- Returns false for weekend trip when current time is on a weekday.
- Returns false for weekday trip when current time is on a weekend.
- Returns true if trip is starting in greater than 1 hr, but the last time checked was 2 hours ago
- Returns false if trip is starting in greater than 1 hr, but the last time checked was 2 minutes ago
- Returns false if trip is starting in 45 minutes and the last time checked was 20 minutes ago
- Returns true if trip is starting in 45 minutes and the last time checked was 2 minutes ago
- Returns false if trip is starting in 45 minutes and the last time checked was 20 minutes ago
- Returns false if trip is starting in 10 minutes and the last time checked was 2 minutes ago
- Returns true if trip has ended 3 minutes ago

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+        Itinerary latestItinerary = trip.latestItinerary();
+        // Construct set from new alerts.
+        Set<LocalizedAlert> newAlerts = new HashSet<>(itinerary.getAlerts());
+        Set<LocalizedAlert> previousAlerts = new HashSet<>(latestItinerary.getAlerts());
+        // Unseen alerts consists of all new alerts that we did not previously track.
+        Set<LocalizedAlert> unseenAlerts = new HashSet<>(newAlerts);
+        unseenAlerts.addAll(previousAlerts);
+        // Resolved alerts consists of all previous alerts that no longer exist.
+        Set<LocalizedAlert> resolvedAlerts = new HashSet<>(previousAlerts);
+        resolvedAlerts.removeAll(newAlerts);
```
evansiroky: These lines of code could be extracted into a new method which could then have some unit tests added to make sure the correct set of alerts are generated.

evansiroky (changes requested):
See comments.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: throw RuntimeException

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Offset trip time with current date.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: condition should be: minutesSinceLastCheck >= 60

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Yes, will add.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
landonreed: Might be able to switch to the AuthException for now.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
```diff
@@ -0,0 +1,207 @@
...
+     * This is only used during testing to inject a mock OTP response for comparison against a monitored trip.
+     */
+    private OtpDispatcherResponse injectedOtpResponseForTesting;
+    private final List<TripMonitorNotification> notifications = new ArrayList<>();
+
+    public CheckMonitoredTrip(MonitoredTrip trip) {
+        this.trip = trip;
+    }
+
+    public CheckMonitoredTrip(MonitoredTrip trip, OtpDispatcherResponse injectedOtpResponseForTesting) {
```
landonreed, line 50: Add comment for constructor only being used during testing.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Correct.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Instead of this, we decided to snapshot the notifications field for testing.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Good suggestion

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Yea, I think we're good, will remove comment.

src/main/java/org/opentripplanner/middleware/trip_monitor/jobs/CheckMonitoredTrip.java
landonreed: Will add TODO about whether we want to continue checking over the course of the trip.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
landonreed: I'll add some code for the CheckMonitoredTrip logic to clear these out once the trip has ended.

landonreed:
@evansiroky, I'm going to go ahead and merge this into the `trip-monitor` branch and fix the conflicts so we can keep this moving.

PR #46, opened 2020-07-29
Api key management

br648:
### Checklist

- [ ] Code coverage improves or is at 100% _(remove this if not merging to master)_

### Description

PR to cover create and delete of api keys for an API user and view api usage logs. The following Swagger endpoints can be used to test the api key part of this PR (taken from http://localhost:4567/):

POST: http://localhost:4567/api/secure/application/{userId}/apikey
DELETE: http://localhost:4567/api/secure/application/{userId}/apikey/{apiKeyId}

### To Run

An AWS_PROFILE is required, or AWS access has been configured for your operating environment e.g. C:\Users<username.aws\credentials in Windows or Mac OS equivalent.

A DEFAULT_USAGE_PLAN_ID is required. This is needed so a newly created api key can be assigned to a usage plan if one isn't provided by the calling sevice. This can be obtained by logging into the AWS console and navagating to API gateway, usage plans. From here the desired usage plan can be selected and the id extracted.


src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
```diff
@@ -0,0 +1,150 @@
...
+        // delete api key from AWS
+        ApiGatewayUtils.deleteApiKeys(Collections.singletonList(apiKeyId));
+
+        return true;
+    }
+
+    /**
+     * Get API usage from AWS api gateway for a users api keys.
+     */
+    private static List<GetUsageResult> getUsageLogsForApiUser(Request req, Response res) {
```
landonreed: I was thinking that this endpoint would just go under the `LogController` and perhaps just introduce some additional logic into the existing endpoint to account for `ApiUsers`.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
```diff
@@ -0,0 +1,150 @@
...
+        = getConfigPropertyAsText("DEFAULT_USAGE_PLAN_ID");
+
+    /**
+     * Register http endpoints with {@link spark.Spark} instance at the provided API prefix.
+     */
+    public static void register(Service spark, String apiPrefix) {
+        // FIXME need to agree on endpoint (/admin or /secure or a mix) and endpoint names
+        spark.get(apiPrefix + "/admin/apikey/create", ApiKeyManagementController::createApiKeyForApiUser, JsonUtils::toJson);
+        spark.get(apiPrefix + "/admin/apikey/delete", ApiKeyManagementController::deleteApiKeyForApiUser, JsonUtils::toJson);
+        spark.get(apiPrefix + "/admin/apikey/usage", ApiKeyManagementController::getUsageLogsForApiUser, JsonUtils::toJson);
```
landonreed: @br648, I think these should ultimately go under `/secure`. Also, I think it might be more appropriate for the endpoints to be added to the ApiUserController by overriding the `buildEndpoint` method. And the approriate  http methods (post, delete) should be used for create/delete keys. Let me know if you need more info.

For ApiUsers creating keys directly, we'll probably only let them create up to some limit (e.g., 2 keys), but that might not apply for admins. Same goes for usage plan. ApiUsers can only create for the DEFAULT_USAGE_PLAN, but an admin user could pass in a query param for a different plan (which would presumably have higher request limits).

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
```diff
@@ -0,0 +1,150 @@
...
+        // FIXME assumption is that the api key will be required by the calling service
+        return apiKey;
+    }
+
+    /**
+     * Delete an api key from a given user's list of api keys (if present) and from AWS api gateway.
+     */
+    private static boolean deleteApiKeyForApiUser(Request req, Response res) {
+
+        ApiUser user = getApiUser(req);
```
landonreed: We should allow an AdminUser to delete any key and an ApiUser to delete only their own keys.

landonreed (changes requested):
Functionality is there, but I think the apikey endpoints need to be built into ApiUserController by overriding `buildEndpoint` (see comments for more info). And getusage functionality should be built into LogController.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
```diff
@@ -0,0 +1,150 @@
...
+        return null;
+    }
+
+    /**
+     * Get an Api user from Mongo DB based on the provided user id. If no user id is provided, or no matching Api user
+     * can be found, respond and halt.
+     */
+    private static ApiUser getApiUser(Request req) {
+
+        String userId = req.queryParamOrDefault("userId", null);
```
landonreed: User Id should be an optional query param. ApiUser should be derived from token if the param is not provided (assumption is that the request is coming from the ApiUser themselves in that case).

br648:
@landonreed - Updates made and tested.

Couple of things:

1) Testing, instead of spending time on this I would like your views before proceeding.
2) The 'usagePlanId' is hidden from the Swagger API, in that, there is no reference of this parameter. Is this ok given that only admin can make use of it?

br648:
One other thing I failed to mention:

The Swagger UI (http://localhost:4567/) presents the following error:

```
Errors

Resolver error at definitions.GetApiKeyResult.properties.tags.$ref
Could not resolve reference: #/definitions/Map
```
It doesn't seem to impact the endpoints as testing was fine, but it would be nice to resolve. I think this is due to serializing of the AWS GetApiKeyResult object. A simplier version containing just the required parameters may be the way forward.

binh-dam-ibigroup:
> One other thing I failed to mention:
> 
> The Swagger UI (http://localhost:4567/) presents the following error:
> 
> ```
> Errors
> 
> Resolver error at definitions.GetApiKeyResult.properties.tags.$ref
> Could not resolve reference: #/definitions/Map
> ```
> 
> It doesn't seem to impact the endpoints as testing was fine, but it would be nice to resolve. I think this is due to serializing of the AWS GetApiKeyResult object. A simplier version containing just the required parameters may be the way forward.

I have something to address that in #50.

configurations/default/env.yml.tmp
```diff
@@ -15,6 +15,7 @@ AUTH0_API_CLIENT: your-api-client-id
 AUTH0_API_SECRET: your-api-secret-id
 AWS_PROFILE: default
 DEFAULT_USAGE_PLAN_ID: usage-plan-id
+OTP_MIDDLEWARE_URI_ROOT: http://otp-middleware-server:4567
```
binh-dam-ibigroup, line 18: This only appears in a commented code block. Do we still need this?

binh-dam-ibigroup: It is only being potentially used in test, so maybe we can remove from this file and keep what you have in `test/env.yml`.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -121,6 +123,39 @@ public static User getUserByEmail(String email, boolean createIfNotExists) {
...
+    /**
+     * Method to trigger an Auth0 job to resend a verification email. Returns an Auth0 {@link Job} which can be
+     * used to monitor the progress of the job (using job ID). Typically the verification email goes out pretty quickly
+     * so there shouldn't be too much of a need to monitor the result.
+     */
+    public static Job resendVerificationEmail(String userId) {
+        try {
+            return getManagementAPI()
+                .jobs()
+                // FIXME: This may need to be the otp-admin client_id instead.
```
binh-dam-ibigroup, line 137: Maybe? An OtpUser who is self-creating an account could have a need to request another verification email.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -121,6 +123,39 @@ public static User getUserByEmail(String email, boolean createIfNotExists) {
...
+        } catch (Auth0Exception e) {
+            BugsnagReporter.reportErrorToBugsnag("Could not send verification email", e);
+            return null;
+        }
+    }
+
+    /**
+     * Get a single Auth0 user for the provided Auth0 user ID.
+     */
+    public static User getUserByAuth0Id(String auth0Id) {
```
binh-dam-ibigroup: Is this used anywhere?

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -48,8 +52,15 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
                 this::getUserFromRequest, JsonUtils::toJson
             )
 
+            .get(path(ROOT_ROUTE + VERIFICATION_EMAIL_PATH)
+                    .withDescription("Triggers a job to resend the Auth0 verification email.")
+                    .withResponseType(Job.class),
+                this::resendVerificationEmail, JsonUtils::toJson
+            )
+
             // Options response for CORS for the token path
```
binh-dam-ibigroup: Please update the comment.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -344,7 +344,7 @@ private String getIdFromRequest(Request req) {
      * Get a request parameter value.
      * This method will halt the request if paramName is not provided in the request.
      */
-    private String getRequiredParamFromRequest(Request req, String paramName) {
+    protected String getRequiredParamFromRequest(Request req, String paramName) {
```
binh-dam-ibigroup: A similar method also exists in`HttpUtils`.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,163 @@
...
+            )
+            // Delete API key
+            .delete(path(API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Deletes API key for ApiUser")
+                    .withPathParam().withName("apiKeyId").withDescription("The ID of the API key.").and()
+                    .withResponseType(persistence.clazz),
+                this::deleteApiKeyForApiUser, JsonUtils::toJson
+            )
+
+            // Options response for CORS for the token path
```
binh-dam-ibigroup: Please update the comment.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,163 @@
...
         super(apiPrefix, Persistence.apiUsers, "secure/application");
     }
 
+    @Override
+    protected void buildEndpoint(ApiEndpoint baseEndpoint) {
+        LOG.info("Registering path {}.", ROOT_ROUTE + API_KEY_PATH);
+
+        // Add the user token route BEFORE the regular CRUD methods
+        // (otherwise, /fromtoken requests would be considered
+        // by spark as 'GET user with id "fromtoken"', which we don't want).
```
binh-dam-ibigroup: Please update the comment.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,163 @@
...
+        String apiKeyId = getRequiredParamFromRequest(req, "apiKeyId");
+        if (apiKeyId == null) {
+            logMessageAndHalt(req,
+                HttpStatus.BAD_REQUEST_400,
+                "An api key id is required",
+                null);
+        }
+        if (!userHasKey(targetUser, apiKeyId)) {
+            logMessageAndHalt(req,
+                HttpStatus.BAD_REQUEST_400,
```
binh-dam-ibigroup: This should be 403 forbidden
EDIT: or perhaps 404 not found if we say the key was not found the user's keys.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -63,4 +198,42 @@ boolean preDeleteHook(ApiUser user, Request req) {
...
+            // If user Id is provided, assumption is that the requester is an AdminUser, with permission to modify an API
+            // user.
+            if (!isUserAdmin(requestingUser)) {
+                logMessageAndHalt(req, HttpStatus.FORBIDDEN_403, "Must be an admin to perform this operation.");
+            }
+            user = Persistence.apiUsers.getById(userId);
+            if (user == null) {
+                logMessageAndHalt(
+                    req,
+                    HttpStatus.BAD_REQUEST_400,
```
binh-dam-ibigroup: This should be 404 not found.

binh-dam-ibigroup (commented):
I have to review ~so~ some more. My comments so far are for adjusting comments and some HTTP return codes.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiKeyManagementController.java
br648: Updated.

configurations/default/env.yml.tmp
br648: Removed. Testing to be addressed.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
br648: Doesn't look like it, have removed.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Updated to use the method it HttpUtils.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
landonreed: I think the client_id is mainly used to tailor the contents of the verification email (i.e., it will include the application name/logo based on the provided client_id). So ultimately I think we do want to add this, but it's a bit more configuration that I wanted to defer on until we were sure.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -43,13 +47,20 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
...
-            .options(path(TOKEN_PATH), (req, res) -> "");
+            .get(path(ROOT_ROUTE + VERIFICATION_EMAIL_PATH)
+                    .withDescription("Triggers a job to resend the Auth0 verification email.")
+                    .withResponseType(Job.class),
+                this::resendVerificationEmail, JsonUtils::toJson
+            )
+
+            // Options response for CORS for the token and verification email paths
+            .options(path(TOKEN_PATH), (req, res) -> "")
+            .options(path(VERIFICATION_EMAIL_PATH), (req, res) -> "");
```
landonreed: @binh-dam-ibigroup, are these CORS methods needed anymore (now that we're using the proxy on API Gateway)?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+    }
+
+    /**
+     * Shorthand method to determine if an API user exists and has an API key.
+     */
+    private boolean userHasKey(ApiUser user, String apiKeyId) {
+        return user != null &&
+            user.apiKeys
+                .stream()
+                .filter(ApiKey -> apiKeyId.equals(ApiKey.id))
```
landonreed: I find it strange that this lambda call uses a capitalized param `ApiKey`. Isn't that typically reserved for classes (not instances)? Also, would it be simpler to user `stream().anyMatch(apiKey -> apiKeyId.equals(apiKey.id))`

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+                HttpStatus.NOT_FOUND_404,
+                String.format("User id (%s) does not have expected api key id (%s)", targetUser.id, apiKeyId),
+                null);
+        }
+
+        // Delete api key from AWS.
+        boolean success = ApiGatewayUtils.deleteApiKeys(Collections.singletonList(new ApiKey(apiKeyId)));
+        if (success) {
+            // Delete api key from user and persist
+            targetUser.apiKeys.remove(new ApiKey(apiKeyId));
```
landonreed: Maybe this a pretty standard pattern in Java, but it feels like a more reliable way to do this would be to just call `removeIf`:

```java
targetUser.apiKeys.removeIf(apiKey -> apiKeyId.equals(apiKey.id));
```

That would allow for the removal of `ApiKey#equals` and `ApiKey#hashCode`. Open to being overruled if anyone feels strongly about this.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
     }
 
+    @Override
+    protected void buildEndpoint(ApiEndpoint baseEndpoint) {
+        LOG.info("Registering path {}.", ROOT_ROUTE + API_KEY_PATH);
+
+        // Add the api key route BEFORE the regular CRUD methods
+        // TODO: Define usage plan id as an optional parameter under Spark Swagger. 'withParam' method does not register
+        //  the parameter and 'withPathParam' is not optional (even with 'withRequired(false))'. For the time being
+        //  appending '?usagePlanId=<id>' works.
```
landonreed: I think this TODO should be closer to the Create API Key method to avoid confusion.

src/main/java/org/opentripplanner/middleware/models/ApiKey.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public String id;
+
+    /**
+     * The name given to the api key.
+     */
+    public String name;
+
+    /**
+     * The api key value. Can not be transient as it is excluded from the Spark response.
```
landonreed: @br648, I think we'll end up storing the API key value at some point by accident if we don't make this transient. Maybe we should just store the value and not hide it (if we do this, we could also remove the getApiKey endpoint/method). Ultimately I'm not really sure if there's much gained by keeping it opaque/out of the db (this was just my initial instinct). Curious to hear other's thoughts.

landonreed (changes requested):
Looks good. Just a few comments regarding ApiKey#value and some syntax stuff.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -43,13 +47,20 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
...
-                .withDescription("Retrieves an " + persistence.clazz.getSimpleName() + " entity using an Auth0 access token passed in an Authorization header.")
-                .withResponseType(persistence.clazz),
+                    .withDescription("Retrieves an " + persistence.clazz.getSimpleName() + " entity using an Auth0 access token passed in an Authorization header.")
+                    .withResponseType(persistence.clazz),
                 this::getUserFromRequest, JsonUtils::toJson
             )
 
-            // Options response for CORS for the token path
-            .options(path(TOKEN_PATH), (req, res) -> "");
+            .get(path(ROOT_ROUTE + VERIFICATION_EMAIL_PATH)
```
binh-dam-ibigroup, line 57: Add comment for this GET method.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
binh-dam-ibigroup: ~Yes, I think so. With the proxy, AWS API Gateway just forwards any OPTIONS call to the middleware server, so even a generic response needs to be implemented here. Now, should `(req, res) -> ""` be refactored to a utility method?~

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -31,10 +32,11 @@
 /**
  * Generic API controller abstract class. This class provides CRUD methods using {@link spark.Spark} HTTP request
  * methods. This will identify the MongoDB collection on which to operate based on the provided {@link Model} class.
- *
+ * <p>
```
binh-dam-ibigroup: Do we need this markup?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -84,12 +87,13 @@ public void bind(final SparkSwagger restApi) {
      * (e.g., getOne, getMany, delete) for CRUD operations.
      * It can optionally be overridden by child classes to add any supplemental methods to the baseEndpoint.
      * Either before or after(*) supplemental methods are added, be sure to call the super method to add CRUD operations.
-     *
+     * <p>
```
binh-dam-ibigroup: Do we need this markup?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -106,7 +110,7 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
             .get(path(ROOT_ROUTE)
                     .withDescription("Gets a list of all '" + classToLowercase + "' entities.")
                     .withResponseAsCollection(clazz),
-                    this::getMany, JsonUtils::toJson
```
binh-dam-ibigroup, line 113: Unindent the two lines above, also?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -47,14 +49,14 @@
     private final Class<T> clazz;
 
     /**
-     * @param apiPrefix string prefix to use in determining the resource location
+     * @param apiPrefix   string prefix to use in determining the resource location
```
binh-dam-ibigroup: Do we need that extra space?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -115,7 +119,7 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
                     .withPathParam().withName("id").withDescription("The id of the entity to search.").and()
                     // .withResponses(...) // FIXME: not implemented (requires source change).
                     .withResponseType(clazz),
-                    this::getOne, JsonUtils::toJson
+                this::getOne, JsonUtils::toJson
```
binh-dam-ibigroup, line 128: Unindent the lines above, also?

binh-dam-ibigroup: Actually, is there an IntelliJ setting you are using that I might be missing?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+            .get(path(ROOT_ROUTE + API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Gets the API key value for a given api key ID")
+                    .withPathParam().withName("apiKeyId").withDescription("The api key ID of the API key.").and()
+                    .withResponseType(ApiKey.class),
+                this::getApiKey, JsonUtils::toJson
+            )
+            // Create API key
+            .post(path(API_KEY_PATH)
+                    .withDescription("Creates API key for ApiUser (with optional usagePlanId)")
+                    .withResponseType(persistence.clazz),
```
binh-dam-ibigroup, line 53: Insert this line below `withDescription` (and update the TODO comment above):
```java
.withQueryParam().withName("usagePlanId").withDescription("Optional AWS API Gateway usage plan ID.").and()
```

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+            // Delete API key
+            .delete(path(API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Deletes API key for ApiUser")
+                    .withPathParam().withName("apiKeyId").withDescription("The ID of the API key.").and()
+                    .withResponseType(persistence.clazz),
+                this::deleteApiKeyForApiUser, JsonUtils::toJson
+            )
+
+            // Options response for CORS for the api key path
+            .options(path(API_KEY_PATH), (req, res) -> "");
```
binh-dam-ibigroup: Also need a separate OPTIONS for path `API_KEY_PATH + API_KEY_ID_PARAM`.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Yes, updated.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -82,42 +88,62 @@ public static CreateApiKeyResult createApiKey(String userId, String usagePlanId)
...
+        AmazonApiGateway gateway = getAmazonApiGateway();
+        try {
+            GetApiKeyRequest getApiKeyRequest = new GetApiKeyRequest()
+                .withIncludeValue(true)
+                .withApiKey(apiKeyId);
+            GetApiKeyResult apiKey = gateway.getApiKey(getApiKeyRequest);
+            LOG.info("API key: {}", apiKey.getValue());
+            return apiKey;
+        } catch (Exception e) {
+            LOG.error("Error encountered while fetching API Key", e);
```
binh-dam-ibigroup: If you enter an unknown API key, the method returns 500 with the markup below.
```html
<html><body><h2>500 Internal Server Error</h2></body></html>
```

It should return 404 instead.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: This is way cleaner. Have updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/ApiKey.java
br648: Holding it in Mongo will save a trip to AWS. I'll wait feedback before updating this section.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
binh-dam-ibigroup: Also insert the query param for `userId`.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+            // Create API key
+            .post(path(API_KEY_PATH)
+                    .withDescription("Creates API key for ApiUser (with optional usagePlanId)")
+                    .withResponseType(persistence.clazz),
+                this::createApiKeyForApiUser, JsonUtils::toJson
+            )
+            // Delete API key
+            .delete(path(API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Deletes API key for ApiUser")
+                    .withPathParam().withName("apiKeyId").withDescription("The ID of the API key.").and()
```
binh-dam-ibigroup: Also insert the query param for userId.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -59,8 +201,46 @@ ApiUser preCreateHook(ApiUser user, Request req) {
...
+                );
+            }
+        } else {
+            // User Id is null, API user should be obtained from token.
+            user = requestingUser.apiUser;
+            if (user == null) {
+                logMessageAndHalt(
+                    req,
+                    HttpStatus.BAD_REQUEST_400,
+                    "A user id is required to create an api key",
```
binh-dam-ibigroup: Suggestion: say rather: "The userId parameter is required."

src/main/java/org/opentripplanner/middleware/controllers/api/LogController.java
```diff
@@ -32,20 +37,45 @@ public static void register(Service spark, String apiPrefix) {
      */
     private static List<GetUsageResult> getUsageLogs(Request req, Response res) {
         // keyId param is optional (if not provided, all API keys will be included in response).
```
binh-dam-ibigroup: Move this comment to under `if (apiKeys.isEmpty())`

src/main/java/org/opentripplanner/middleware/models/ApiKey.java
binh-dam-ibigroup: I defer to @evansiroky.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+                .filter(ApiKey -> apiKeyId.equals(ApiKey.id))
+                .findAny()
+                .orElse(null) != null;
+    }
+
+    /**
+     * Create a new API key and assign it to the provided usage plan. If no usage plan is provided use the default
+     * usage plan instead.
+     */
+    private ApiUser createApiKeyForApiUser(Request req, Response res) {
```
binh-dam-ibigroup, line 87: I am not really in favor of returning the entire user as a result of this operation. Could that call just return the created key?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,170 @@
...
+        // Add new API key to user and persist
+        targetUser.apiKeys.add(new ApiKey(apiKeyResult));
+        Persistence.apiUsers.replace(targetUser.id, targetUser);
+        return Persistence.apiUsers.getById(targetUser.id);
+    }
+
+    /**
+     * Delete an api key from a given user's list of api keys (if present) and from AWS api gateway.
+     */
+    private ApiUser deleteApiKeyForApiUser(Request req, Response res) {
```
binh-dam-ibigroup, line 118: Do we need to return the entire user after this operation?

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -82,42 +88,62 @@ public static CreateApiKeyResult createApiKey(String userId, String usagePlanId)
...
+     * Get API key (with key value) for the provided id.
+     */
+    public static GetApiKeyResult getApiKey(String apiKeyId) {
+        AmazonApiGateway gateway = getAmazonApiGateway();
+        try {
+            GetApiKeyRequest getApiKeyRequest = new GetApiKeyRequest()
+                .withIncludeValue(true)
+                .withApiKey(apiKeyId);
+            GetApiKeyResult apiKey = gateway.getApiKey(getApiKeyRequest);
+            LOG.info("API key: {}", apiKey.getValue());
```
binh-dam-ibigroup: Should we be logging this?

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -136,16 +162,30 @@ public static void deleteApiKeys(List<String> apiKeyIds) {
                 usageResults.add(gateway.getUsage(getUsageRequest));
             } catch (Exception e) {
                 // Catch any issues with bad request parameters (e.g., invalid API keyId or bad date format).
-                LOG.error("Unable to get usage results for key id ({}) with start date ({}) and end date ({})",
+                String message = String.format("Unable to get usage results for key id (%s) with start date (%s) and end date (%s)",
```
binh-dam-ibigroup: Suggestion: "Unable to get usage results for key id (%s) between (%s) and (%s)"

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -136,16 +162,30 @@ public static void deleteApiKeys(List<String> apiKeyIds) {
...
-                    endDate,
-                    e);
-                //FIXME add bugsnag
+                    endDate);
+                BugsnagReporter.reportErrorToBugsnag(message, e);
                 throw e;
             }
         }
-        LOG.debug("Retrieving usage logs took {} msec", System.currentTimeMillis() - startTime);
+        LOG.debug("Retrieving usage logs an api key took {} msec", System.currentTimeMillis() - startTime);
```
binh-dam-ibigroup: Incomplete phrase.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
binh-dam-ibigroup: I am leaning towards a URL scheme like `api/secure/application/:userid/apikey` would be more suitable for this POST method and `api/secure/application/:id/apikey/:apikeyid` for the DELETE method. I think we should adopt that unless there is a strong case against it.

binh-dam-ibigroup (changes requested):
I strongly think we can improve the URL scheme for POSTing and DELETEing api keys so it is consistent with the other API paths in use. 
Otherwise it is taking shape, some parameter changes and comment changes.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Never mind. IntelliJ is doing that formatting on its own with the settings I have.

binh-dam-ibigroup:
> > One other thing I failed to mention:
> > The Swagger UI (http://localhost:4567/) presents the following error:
> > ```
> > Errors
> > 
> > Resolver error at definitions.GetApiKeyResult.properties.tags.$ref
> > Could not resolve reference: #/definitions/Map
> > ```
> > 
> > 
> > It doesn't seem to impact the endpoints as testing was fine, but it would be nice to resolve. I think this is due to serializing of the AWS GetApiKeyResult object. A simplier version containing just the required parameters may be the way forward.
> 
> I have something to address that in #50.

The scope for the immediate work involving API Gateway integration has changed per team chat, so we will revisit that when we have to post-process the swagger file again.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Have removed.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Have removed.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: I'm deferring this one to @landonreed

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: I'm deferring this one to @landonreed

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
br648: Probably not, have removed.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
br648: Updated.

br648:
@binh-dam-ibigroup I have created the object `org.opentripplanner.middleware.models.ApiKey` and returning this instead. So this is no longer and issue:

        One other thing I failed to mention:
        The Swagger UI (http://localhost:4567/) presents the following error:

        Errors

        Resolver error at definitions.GetApiKeyResult.properties.tags.$ref
        Could not resolve reference: #/definitions/Map

        It doesn't seem to impact the endpoints as testing was fine, but it would be nice to resolve. I think this is due to serializing of the AWS GetApiKeyResult object. A simplier version containing just the required parameters may be the way forward.

    I have something to address that in #50.

The scope for the immediate work involving API Gateway integration has changed per team chat, so we will revisit that when we have to post-process the swagger file again.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Maybe, if I run Ctrl + Alt +l (for 'L'ima) to format the code, this is what I get. I've not made any changes from the default formatting.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/LogController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: I've inserted the query params for usagePlanId and userId.

@binh-dam-ibigroup The ApiUser would have to define their userId instead of it being obtained from their token. The userId in the URL would still need to be checked against this. I think the URL scheme approach is more suited to an admin user where the userId is required (at least for POST). I'm not against making the change, but would like agreement from @landonreed and @evansiroky before updating and testing it.

br648:
I have updated to address all but the items that require input from the wider team.

binh-dam-ibigroup (commented):
Thanks for the latest changes. Awaiting coordination on the URL scheme and return types when creating/deleting API keys, I'm going to remove my assignment for now.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,173 @@
...
+    @Override
+    protected void buildEndpoint(ApiEndpoint baseEndpoint) {
+        LOG.info("Registering path {}.", ROOT_ROUTE + API_KEY_PATH);
+
+        // Add the api key route BEFORE the regular CRUD methods
+        ApiEndpoint modifiedEndpoint = baseEndpoint
+            // Reveal the actual API key value for the key ID
+            .get(path(ROOT_ROUTE + API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Gets the API key value for a given api key ID")
+                    .withPathParam().withName("apiKeyId").withDescription("The api key ID of the API key.").and()
```
evansiroky: To better format this, can you make each `with...` on a new line?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,173 @@
...
+                    .withResponseType(ApiKey.class),
+                this::getApiKey, JsonUtils::toJson
+            )
+            // Create API key
+            .post(path(API_KEY_PATH)
+                    .withDescription("Creates API key for ApiUser (with optional AWS API Gateway usage plan ID and user ID.)")
+                    .withQueryParam().withName("usagePlanId").withDescription("Optional AWS API Gateway usage plan ID.").and()
+                    .withQueryParam().withName("userId").withDescription("Optional user ID.").and()
```
evansiroky: To better format this, can you make each `with...` on a new line?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
evansiroky: Actually nvm, I see that it is specific to this param.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
evansiroky: Actually nvm, I see that it is specific to this param.

src/main/java/org/opentripplanner/middleware/models/ApiKey.java
evansiroky: Yeah, I think it makes sense to store all this in the db and not make so many get API key requests to AWS.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -57,40 +66,54 @@ public static String httpRequest(URI uri, int connectionTimeout, REQUEST_METHOD
...
     }
+
+    /**
+     * Get a request parameter value.
+     * This method will halt the request if paramName is not provided in the request.
+     */
+    public static String getRequiredParamFromRequest(Request req, String paramName) {
+        String paramValue = req.params(paramName);
+        if (paramValue == null) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "Must provide parameter name.");
```
evansiroky: Should this include the missing paramName in the message?

evansiroky (approved):
I think there's enough other review being done here. I don't see anything very strange, so I'm giving my approval already.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
landonreed: I think this URL scheme makes sense. I'm down with this approach.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
landonreed: It's simpler in terms of API interaction to just return the API user. I think we should keep it as is.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
landonreed: Same answer as above. Let's just return the user.

binh-dam-ibigroup:
@br648 Please remember to add the swagger descriptors (e.g. `withPathParam`, `withProduces`, `withConsumes`), as in https://github.com/ibi-group/otp-middleware/blob/ee90d90cb07568119bc47bd15260b250f3dde369/src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java#L119. Thanks!

src/main/java/org/opentripplanner/middleware/models/ApiKey.java
br648: Updated to save value in DB and removed getApiKey endpoint.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
landonreed: @br648, instead of creating new params (e.g., `:userId`) that don't match what's in ApiController, I think we should be importing the param/path field from ApiController, similar to what we're doing in #33 https://github.com/ibi-group/otp-middleware/pull/33/files#diff-3772d28a6571870eef9631b391ccc391R41-R42

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
binh-dam-ibigroup: @br648 after doing a quick test on AWS, it looks like we don't need to add handlers for OPTIONS, so please remove them.
OPTIONS calls are (were already) forwarded by sparkjava to this code: https://github.com/ibi-group/otp-middleware/blob/dev/src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java#L83

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,157 @@
...
+        // Delete api key from AWS.
+        boolean success = ApiGatewayUtils.deleteApiKeys(Collections.singletonList(new ApiKey(apiKeyId)));
+        if (success) {
+            // Delete api key from user and persist
+            targetUser.apiKeys.removeIf(apiKey -> apiKeyId.equals(apiKey.id));
+            Persistence.apiUsers.replace(targetUser.id, targetUser);
+            return Persistence.apiUsers.getById(targetUser.id);
+        } else {
+            // Throw halt if API key deletion failed.
+            logMessageAndHalt(req, HttpStatus.INTERNAL_SERVER_ERROR_500, "Unknown error deleting API key.");
```
binh-dam-ibigroup, line 143: If you attempt to delete a key that doesn't exist [anymore] in AWS, right now that throws a 500 error. Should it still delete that entry from Mongo? @evansiroky what do you think?

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
br648: Removed from AbstractUserController and ApiUserController

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: @binh-dam-ibigroup I have refactored the `ApiGatewayUtils.deleteApiKeys(List<ApiKey> apiKeys)` to `ApiGatewayUtils.deleteApiKey(ApiKey apiKey)`. The older version had a return in the middle of a loop so would have only deleted one api key anyway! The newer version returns true even if the key is not found by AWS, the assumption being it has already been removed. False is still returned if any other exceptions occur.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,152 @@
...
+        // Add the api key route BEFORE the regular CRUD methods
+        ApiEndpoint modifiedEndpoint = baseEndpoint
+            // Create API key
+            .post(path(ID_PATH + API_KEY_PATH)
+                    .withDescription("Creates API key for ApiUser (with optional AWS API Gateway usage plan ID).")
+                    .withConsumes(JSON_ONLY)
+                    .withPathParam().withName(ID_PARAM).withRequired(true).withDescription("The user ID")
+                    .and()
+                    .withQueryParam().withName("usagePlanId").withRequired(false).withDescription("Optional AWS API Gateway usage plan ID.")
+                    .and()
```
binh-dam-ibigroup, line 51: Insert a `.withRequestType(<class_for_request_body>)` (see `ApiController` -> POST method).

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,152 @@
...
+
+        // Add the api key route BEFORE the regular CRUD methods
+        ApiEndpoint modifiedEndpoint = baseEndpoint
+            // Create API key
+            .post(path(ID_PATH + API_KEY_PATH)
+                    .withDescription("Creates API key for ApiUser (with optional AWS API Gateway usage plan ID).")
+                    .withConsumes(JSON_ONLY)
+                    .withPathParam().withName(ID_PARAM).withRequired(true).withDescription("The user ID")
+                    .and()
+                    .withQueryParam().withName("usagePlanId").withRequired(false).withDescription("Optional AWS API Gateway usage plan ID.")
```
binh-dam-ibigroup: Remove unnecessary `.withRequired(false)`.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,152 @@
...
+                    .and()
+                    .withProduces(JSON_ONLY)
+                    .withResponseType(persistence.clazz),
+                this::createApiKeyForApiUser, JsonUtils::toJson
+            )
+            // Delete API key
+            .delete(path(ID_PATH + API_KEY_PATH + API_KEY_ID_PARAM)
+                    .withDescription("Deletes API key for ApiUser.")
+                    .withConsumes(JSON_ONLY)
```
binh-dam-ibigroup: Don't need `.withConsumes` here.

binh-dam-ibigroup (commented):
I am still reviewing, but see my comments about method descriptors.

landonreed (approved):

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -1,28 +1,152 @@
...
     }
 
+    @Override
+    protected void buildEndpoint(ApiEndpoint baseEndpoint) {
+        LOG.info("Registering path {}.", ROOT_ROUTE + API_KEY_PATH);
+
+        // Add the api key route BEFORE the regular CRUD methods
+        ApiEndpoint modifiedEndpoint = baseEndpoint
+            // Create API key
+            .post(path(ID_PATH + API_KEY_PATH)
```
binh-dam-ibigroup, line 46: When trying this on Postman, I got `<html><body><h2>404 Not found</h2></body></html>` in the response, and
this message in the logs: `The requested route [/api/secure/application/<api-user-id>/apiKey] has not been mapped in Spark for Accept: [application/json]`

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
binh-dam-ibigroup: Resolved (spell `apikey` the correct way).

br648: Lowercase 'k' in apiKey is required.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
binh-dam-ibigroup: Sorry, the way it is implemented there is no request body, so you don't need that. Also, remove the `.withConsumes`.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -62,6 +64,8 @@ public static CreateApiKeyResult createApiKey(String userId, String usagePlanId)
             CreateApiKeyRequest apiKeyRequest = new CreateApiKeyRequest();
             apiKeyRequest.setSdkRequestTimeout(SDK_REQUEST_TIMEOUT);
             apiKeyRequest
+                //FIXME This may need to include stage key(s). Not sure what impact that places on the calling
+                // services though?
                 .withName(userId)
```
binh-dam-ibigroup, line 69: Why is the name of the create API key set to the user id?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -32,14 +33,16 @@
 /**
  * Generic API controller abstract class. This class provides CRUD methods using {@link spark.Spark} HTTP request
  * methods. This will identify the MongoDB collection on which to operate based on the provided {@link Model} class.
- *
+ * <p>
```
binh-dam-ibigroup: Markup?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: It came back from a merge.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Markup has returned (from merge)

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
br648: For reference. See list of API key in AWS: https://console.aws.amazon.com/apigateway/home?region=us-east-1#/api-keys

br648: I need to tidy this list up!

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
br648: Removed (again!)

src/main/java/org/opentripplanner/middleware/controllers/api/LogController.java
```diff
@@ -44,22 +49,22 @@ public void bind(final SparkSwagger restApi) {
         ).get(path(ROOT_ROUTE)
                 .withDescription("Gets a list of all API usage logs.")
                 .withQueryParam()
-                    .withName("keyId")
-                    .withDescription("If specified, restricts the search to the specified AWS API key ID.").and()
+                .withName("keyId")
```
binh-dam-ibigroup, line 52: I suppose IntelliJ flattened all the descriptors!

binh-dam-ibigroup (approved):
Conditional approval: There are some remaining formatting glitches, otherwise it's looking good.

PR #54, opened 2020-08-05
Rename OTP_SERVER config parameter

binh-dam-ibigroup:
### Checklist

### Description

EDITED This PR renames configuration parameter `OTP_SERVER` to `OTP_API_ROOT` for clarity. The code and documentation are updated accordingly. 

~This PR modifies the processing of OTP URLs so that the middleware does not strip the `/otp` prefix found in most deployments where the default router is at `/otp/routers/default`. (as in `http://otp-server.com/otp/routers/default`).
The current code sends routing requests to `http://otp-server.com/routers/default` which is incorrect.~


evansiroky (approved):

landonreed (commented):
@binh-dam-ibigroup, I'm not sure if this is the right approach. See changes here: https://github.com/ibi-group/otp-middleware/pull/45/files#diff-d4516ed28d393bcd0165be1cf740ba49R22-R23

We could just set the `OTP_SERVER` to contain the `/otp` path.

br648 (approved):
This is fine, but as @landonreed suggests, if this can be done via config that would be best.

binh-dam-ibigroup:
> @binh-dam-ibigroup, I'm not sure if this is the right approach. See changes here: https://github.com/ibi-group/otp-middleware/pull/45/files#diff-d4516ed28d393bcd0165be1cf740ba49R22-R23
> 
> We could just set the `OTP_SERVER` to contain the `/otp` path.

@landonreed You need to do that too. However the current `OtpRequestProcessor` code with strip `/otp` in both OTP_SERVER path and the router path, which is undesirable.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -76,6 +78,15 @@ private static String proxy(Request request, spark.Response response) {
         return otpDispatcherResponse.responseBody;
     }
 
+    /**
+     * Get request path intended for OTP API by removing the proxy endpoint (/otp).
+     */
+    static String stripFirstOtpPathPrefix(String uri) {
```
landonreed: This should be renamed `stripOtpProxyFromPath` or something similar. Also, please add example uri value.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -76,6 +78,15 @@ private static String proxy(Request request, spark.Response response) {
...
+        // Only remove the first /otp proxy endpoint.
+        // The other one, if any, is part of the OTP router path on the OTP server.
+        return uri.replaceFirst(OTP_PROXY_ENDPOINT, "");
```
landonreed: I don't think this comment needs to exist about the first vs. other `/otp` path components. Keeping a comment here makes sense but it should be reworded because we don't necessarily anticipate /otp/otp request paths.

src/test/java/org/opentripplanner/middleware/otp/OtpRequestProcessorTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class OtpRequestProcessorTest {
+
+    @Test
+    public void testStripFirstOtpPathPrefix() {
+        // Below is an example for a typical OTP instance:
+        // - The first otp is for the middleware's OTP proxy endpoint and should be removed.
+        // - The second otp is part of the OTP server default router path and should be kept.
+        // (The default router path is typically set in OTP-RR config as /otp/routers/default.)
+        final String requestUri = "/otp/otp/routers/default/park_and_ride";
```
landonreed: Please use full example uri (with sample otp middleware host name).

landonreed (changes requested):
A few comments for clarifying request uris. Also, as discussed on internal chat, please change config field OTP_SERVER to OTP_API_ROOT

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
binh-dam-ibigroup: This method is irrelevant with renaming OTP_API_ROOT.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
binh-dam-ibigroup: I have deleted this method, it is irrelevant now with the renaming of the config variable OTP_API_ROOT.

src/test/java/org/opentripplanner/middleware/otp/OtpRequestProcessorTest.java
binh-dam-ibigroup: I have deleted the corresponding method and this test.

landonreed (approved):
Approved, but please update PR description and title.

br648 (approved):
All good.

PR #58, opened 2020-08-13
Api key management unit testing

br648:
### Checklist

### Description

This PR allows developers to unit test the API key management functionality via the Swagger/Spark endpoints.  A key enhancement to authentication allows test user objects to be used instead of a temporary admin account. User objects are matched with actual database objects providing more accurate testing. 

This is achieved by placing the Auth0 user id (of the newly created object) into the Authorization header parameter for each request. If authorization is NOT disabled a bearer token is expected (normal operation), if authorization is disabled, the Auth0 user id is extracted from the Authorization header and used to load the appropriate user from the database. If the Authorization header is not defined or the user is not available in the DB, a temporary admin account is created matching the previous functionality.

The following config parameters are required for these tests to run:
 
An AWS_PROFILE is required, or AWS access has been configured for your operating environment e.g. C:\Users\<username>\\.aws\credentials in Windows or Mac OS equivalent.

DISABLE_AUTH set to true to bypass auth checks and use users defined here.
DEFAULT_USAGE_PLAN_ID set to a valid usage plan id. AWS requires this to create an api key.

### DIFF
--- a/configurations/test/env.yml
+++ b/configurations/test/env.yml
@@ -26,4 +26,3 @@ MONGO_DB_NAME: otp_middleware_test
 OTP_API_ROOT: http://otp-server.example.com/otp
 OTP_PLAN_ENDPOINT: /routers/default/plan
 MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
-DEFAULT_USAGE_PLAN_ID: ajjp1j
--- a/configurations/test/env.yml
+++ b/configurations/test/env.yml
@@ -1,6 +1,7 @@
 AUTH0_API_CLIENT: test-auth0-client-id
 AUTH0_API_SECRET: test-auth0-secret
 AUTH0_DOMAIN: test.auth0.com
+DISABLE_AUTH: true
 
 BUGSNAG_API_KEY: bugsnag-api-key
 BUGSNAG_EVENT_JOB_DELAY_IN_MINUTES: 1
--- a/configurations/test/env.yml
+++ b/configurations/test/env.yml
@@ -1,7 +1,6 @@
 AUTH0_API_CLIENT: test-auth0-client-id
 AUTH0_API_SECRET: test-auth0-secret
 AUTH0_DOMAIN: test.auth0.com
-DISABLE_AUTH: true
 
 BUGSNAG_API_KEY: bugsnag-api-key
 BUGSNAG_EVENT_JOB_DELAY_IN_MINUTES: 1
--- a/configurations/test/env.yml
+++ b/configurations/test/env.yml
@@ -26,3 +26,4 @@ MONGO_DB_NAME: otp_middleware_test
 OTP_API_ROOT: http://otp-server.example.com/otp
 OTP_PLAN_ENDPOINT: /routers/default/plan
 MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
+DEFAULT_USAGE_PLAN_ID: ajjp1j
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
@@ -23,7 +23,8 @@ public class Auth0UserProfile {
     public final String auth0UserId;
 
     /**
-     * Constructor is only used for creating a test user
+     * Constructor is only used for creating a test user. If an Auth0 user id is provided check persistence for matching
+     * user else create default user.
      */
     private Auth0UserProfile(String auth0UserId) {
         if (auth0UserId == null) {
@@ -52,12 +53,14 @@ public class Auth0UserProfile {
     }
 
     /**
-     * Utility method for creating a test user. If auth0 user Id is defined check persistence for matching user else
-     * create default user.
+     * Utility method for creating a test user. If a Auth0 user Id is defined within the Authorization header param
+     * define test user based on this.
      */
     static Auth0UserProfile createTestUser(Request req) {
         String auth0UserId = null;
         if (isAuthHeaderPresent(req)) {
+            // If the auth header has been provided get the Auth0 user id from it. This is different from normal
+            // operation as the parameter will only contain the Auth0 user id and not "Bearer token".
             auth0UserId = req.headers("Authorization");
         }
         return new Auth0UserProfile(auth0UserId);
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
@@ -7,6 +7,8 @@ import org.opentripplanner.middleware.models.AdminUser;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.models.OtpUser;
 import org.opentripplanner.middleware.persistence.Persistence;
+import org.opentripplanner.middleware.utils.HttpUtils;
+import spark.Request;
 
 import static com.mongodb.client.model.Filters.eq;
 
@@ -22,10 +24,18 @@ public class Auth0UserProfile {
 
     /** Constructor is only used for creating a test user */
     private Auth0UserProfile(String auth0UserId) {
-        this.auth0UserId = auth0UserId;
-        otpUser = new OtpUser();
-        apiUser = new ApiUser();
-        adminUser = new AdminUser();
+        if (auth0UserId == null) {
+            this.auth0UserId = "user_id:string";
+            otpUser = new OtpUser();
+            apiUser = new ApiUser();
+            adminUser = new AdminUser();
+        } else {
+            this.auth0UserId = auth0UserId;
+            Bson withAuth0UserId = eq("auth0UserId", auth0UserId);
+            otpUser = Persistence.otpUsers.getOneFiltered(withAuth0UserId);
+            adminUser = Persistence.adminUsers.getOneFiltered(withAuth0UserId);
+            apiUser = Persistence.apiUsers.getOneFiltered(withAuth0UserId);
+        }
     }
 
     /** Create a user profile from the request's JSON web token. Check persistence for stored user */
@@ -38,9 +48,11 @@ public class Auth0UserProfile {
     }
 
     /**
-     * Utility method for creating a test admin (with application-admin permissions) user.
+     * Utility method for creating a test user. If auth0 user Id is defined check persistence for matching user else
+     * create default user.
      */
-    public static Auth0UserProfile createTestAdminUser() {
-        return new Auth0UserProfile("user_id:string");
+    static Auth0UserProfile createTestUser(Request req) {
+        String auth0UserId = req.queryParamOrDefault("auth0UserId", null);
+        return new Auth0UserProfile(auth0UserId);
     }
 }
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
@@ -7,10 +7,10 @@ import org.opentripplanner.middleware.models.AdminUser;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.models.OtpUser;
 import org.opentripplanner.middleware.persistence.Persistence;
-import org.opentripplanner.middleware.utils.HttpUtils;
 import spark.Request;
 
 import static com.mongodb.client.model.Filters.eq;
+import static org.opentripplanner.middleware.auth.Auth0Connection.isAuthHeaderPresent;
 
 /**
  * User profile that is attached to an HTTP request.
@@ -22,7 +22,9 @@ public class Auth0UserProfile {
     public final AdminUser adminUser;
     public final String auth0UserId;
 
-    /** Constructor is only used for creating a test user */
+    /**
+     * Constructor is only used for creating a test user
+     */
     private Auth0UserProfile(String auth0UserId) {
         if (auth0UserId == null) {
             this.auth0UserId = "user_id:string";
@@ -38,7 +40,9 @@ public class Auth0UserProfile {
         }
     }
 
-    /** Create a user profile from the request's JSON web token. Check persistence for stored user */
+    /**
+     * Create a user profile from the request's JSON web token. Check persistence for stored user
+     */
     public Auth0UserProfile(DecodedJWT jwt) {
         this.auth0UserId = jwt.getClaim("sub").asString();
         Bson withAuth0UserId = eq("auth0UserId", auth0UserId);
@@ -52,7 +56,10 @@ public class Auth0UserProfile {
      * create default user.
      */
     static Auth0UserProfile createTestUser(Request req) {
-        String auth0UserId = req.queryParamOrDefault("auth0UserId", null);
+        String auth0UserId = null;
+        if (isAuthHeaderPresent(req)) {
+            auth0UserId = req.headers("Authorization");
+        }
         return new Auth0UserProfile(auth0UserId);
     }
 }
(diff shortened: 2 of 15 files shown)


src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
```diff
@@ -38,9 +52,14 @@ public Auth0UserProfile(DecodedJWT jwt) {
...
+     * Utility method for creating a test user. If auth0 user Id is defined check persistence for matching user else
+     * create default user.
      */
-    public static Auth0UserProfile createTestAdminUser() {
-        return new Auth0UserProfile("user_id:string");
+    static Auth0UserProfile createTestUser(Request req) {
+        String auth0UserId = null;
+        if (isAuthHeaderPresent(req)) {
+            auth0UserId = req.headers("Authorization");
+        }
```
landonreed, line 65: This might need a comment explaining what is going on. Also, an alternative approach to improve closeness of this test code with the actual run environment might be to use `Auth0Connection#getTokenFromRequest` to return the auth0UserId and construct the test user with that string. This would require the Authorization header used in tests to use the `"Authorization": "Bearer $token_or_user_id"` format as well.

EDIT: I played around with this idea for a little bit, but I don't think it's the right approach unless we do a good bit of refactoring of `getTokenFromRequest`. I think we would need to change the logMessageAndHalt invocations to just throw exceptions and handle those differently. So, I'll rescind this recommendation for now.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+        removeApiUser(apiUser.id);
+    }
+
+    /**
+     * Ensure that an {@link ApiUser} can create an API key for self.
+     */
+    @Test
+    public void canCreateApiKeyForSelf() {
+        String url = String.format("http://localhost:4567/api/secure/application/%s/apikey", apiUser.id);
+        Assertions.assertEquals(getStatusCodeFromResponse(url, HttpUtils.REQUEST_METHOD.POST, apiUser.auth0UserId), 200);
```
landonreed: I think a lot of these tests could assert more than the 200 response. Might be good to send back the full http response so that the body can be extracted/checked.

Also, the name of the method could be something like: `mockAuthenticatedRequest`. It should probably live in `TestUtils` and might even benefit from just taking a string path instead of the full URL (it could just append the path to `http://localhost:4567` in its method body).

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+
+        // refresh api keys
+        apiUser = getApiUser(apiUser.id);
+        for (ApiKey apiKey : apiUser.apiKeys) {
+            // delete api keys if present
+            String url = String.format("http://localhost:4567/api/secure/application/%s/apikey/%s",
+                apiUser.id,
+                apiKey.id);
+            getStatusCodeFromResponse(url, HttpUtils.REQUEST_METHOD.DELETE, apiUser.auth0UserId);
+        }
```
landonreed, line 66: This for loop should probably go in an `ApiUser#deleteKeys` method that also gets called in preDeleteHook. I don't think it needs to be done via http in this test.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+     * Make sure that at least one API key exists.
+     */
+    private void ensureAtLeastOneApiKeyIsAvailable() {
+        // refresh api keys
+        apiUser = getApiUser(apiUser.id);
+
+        if (apiUser.apiKeys.isEmpty()) {
+            // create key
+            String url = String.format("http://localhost:4567/api/secure/application/%s/apikey", apiUser.id);
+            getStatusCodeFromResponse(url, HttpUtils.REQUEST_METHOD.POST, adminUser.auth0UserId);
```
landonreed: Since this is outside the context of a test (and is more so just setup activity), I don't think you need to do this via http. It might be worth refactoring a bit and adding an `ApiUser#createApiKey` method.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+import org.opentripplanner.middleware.models.ApiUser;
+import org.opentripplanner.middleware.persistence.PersistenceUtil;
+import org.opentripplanner.middleware.utils.HttpUtils;
+
+import java.io.IOException;
+
+import static org.opentripplanner.middleware.persistence.PersistenceUtil.getApiUser;
+import static org.opentripplanner.middleware.persistence.PersistenceUtil.getStatusCodeFromResponse;
+import static org.opentripplanner.middleware.persistence.PersistenceUtil.removeAdminUser;
+import static org.opentripplanner.middleware.persistence.PersistenceUtil.removeApiUser;
```
landonreed: On a recent Travis run, it looks like no tests were run. https://travis-ci.org/github/ibi-group/otp-middleware/builds/717518093#L679-L700

Not sure what's happening there.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
```diff
@@ -45,6 +54,36 @@ public static ApiUser createApiUser(String email) {
...
+        user.email = email;
+        Persistence.adminUsers.create(user);
+        return user;
+    }
+
+    /**
+     * Get Api user from database
+     */
+    public static ApiUser getApiUser(String id) {
+        return Persistence.apiUsers.getById(id);
```
landonreed: I'm not sure these one-line method calls really add much at this stage. If we were doing more in the method body it might make sense, but I feel like this just obfuscates things at the moment.

landonreed (changes requested):
Looking good. For some reason all the tests are being skipped though 😱. Also, these tests should probably only be run with E2E enabled because they rely on AWS services.

Tests themselves could use a little bit of refactoring as well (e.g., delete key could check that the key ID no longer exists in user's key list and potentially with the Amazon SDK as well).

configurations/test/env.yml
```diff
@@ -1,6 +1,7 @@
 AUTH0_API_CLIENT: test-auth0-client-id
 AUTH0_API_SECRET: test-auth0-secret
 AUTH0_DOMAIN: test.auth0.com
+DISABLE_AUTH: true
```
binh-dam-ibigroup: Insert a line for the `AWS_PROFILE` parameter, or mention that in the PR description. That parameter is required to make the new tests pass.

src/test/java/org/opentripplanner/middleware/OtpMiddlewareTest.java
```diff
@@ -29,7 +32,19 @@ public static void setUp() throws RuntimeException, IOException {
 
         LOG.info("Starting server");
         OtpMiddlewareMain.inTestEnvironment = true;
-        OtpMiddlewareMain.main(new String[]{"configurations/test/env.yml"});
+        // If in the e2e environment, use the secret env.yml file to start the server.
+        // TODO: When ran on Travis CI, this file will automatically be setup.
+        String[] args = getBooleanEnvVar("RUN_E2E")
+            ? new String[] { "configurations/default/env.yml" }
+            : new String[] { "configurations/test/env.yml"};
```
binh-dam-ibigroup, line 39: The unit tests fail out the box for me. Are there other parameters needed in `test/env.yml` besides `DEFAULT_USAGE_PLAN_ID`?

binh-dam-ibigroup (commented):
I was not able to run the unit tests using the instructions from the PR description... Could you update the instructions if something is missing there?

src/main/java/org/opentripplanner/middleware/auth/Auth0UserProfile.java
br648: yeah, I looked into this and decided it would impact the existing functionality too much. Have updated comments in this class.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
evansiroky: Agree. This just checks that the OTP middleware responded with HTTP 200. This probably means everything went ok, but a better test would be to perform the HTTP request and then verify that the database has the expected records.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+
+        removeApiUser(apiUser.id);
+    }
+
+    /**
+     * Ensure that an {@link ApiUser} can create an API key for self.
+     */
+    @Test
+    public void canCreateApiKeyForSelf() {
+        String url = String.format("http://localhost:4567/api/secure/application/%s/apikey", apiUser.id);
```
evansiroky: This appears to perform a live request to the AWS API Gateway. It would be good to refactor the ApiGatewayUtils to mock the AmazonApiGateway returned in the `getAmazonApiGateway` method that can return the results we want in the test cases.

evansiroky (changes requested):
Tests are failing locally. Also, see my two comments in the `ApiKeyManagementTest.java` file.

configurations/test/env.yml
br648: @binh-dam-ibigroup , yeah thanks I missed this. I have updated the PR description.

src/test/java/org/opentripplanner/middleware/persistence/PersistenceUtil.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: @landonreed Looks like this was removed at some point from the POM:

```
            <plugin>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>2.22.2</version>
            </plugin>
```
With this back in place the Maven test option within Intellij is working again. I expect the same once I push this update.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,126 @@
...
+     */
+    @AfterAll
+    public static void tearDown() {
+        removeAdminUser(adminUser.id);
+
+        // refresh api keys
+        apiUser = getApiUser(apiUser.id);
+        for (ApiKey apiKey : apiUser.apiKeys) {
+            // delete api keys if present
+            String url = String.format("http://localhost:4567/api/secure/application/%s/apikey/%s",
```
binh-dam-ibigroup: The URL should be extracted as constant.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
binh-dam-ibigroup: Extract this into `Response createApiKey(String targetUserId, String requestingUserId)` and reuse 3 times.

binh-dam-ibigroup (changes requested):
Refactor a URL and code to create keys, plus what Evan said about checking the database record.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/OtpMiddlewareTest.java
br648: As far as I'm aware, no more than what is (now) defined in the PR description.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: I've updated to do this via ApiGatewayUtils.

src/test/java/org/opentripplanner/middleware/OtpMiddlewareTest.java
landonreed: @binh-dam-ibigroup, you're probably missing `DISABLE_AUTH=true`.

evansiroky (approved):
Good enough for now I guess. I guess it's fine for now to swallow up these tests into e2e only runs.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,167 @@
...
 /**
  * Tests for creating and deleting api keys. The following config parameters are required for these tests to run:
- *
- * DISABLE_AUTH set to true to bypass auth checks.
- * DEFAULT_USAGE_PLAN_ID set to a valid usage plan id. AWS requires this to create an api key.
- * OTP_MIDDLEWARE_URI_ROOT, e.g. http://localhost:4567.
+ *  - RUN_E2E=true the end-to-end environment variable must be set (NOTE: this is not a config value)
+ *  - An AWS_PROFILE is required, or AWS access has been configured for your operating environment e.g.
+ *    C:\Users\<username>\.aws\credentials in Windows or Mac OS equivalent.
+ *  - DISABLE_AUTH set to true to bypass auth checks and use users defined here. DEFAULT_USAGE_PLAN_ID set to a valid usage
```
binh-dam-ibigroup, line 33: ~Do `AWS_PROFILE` and `DEFAULT_USAGE_PLAN_ID` need to be added to `configurations/test/env.yml`?~
Although this concerns a test, the comment should state that because this is an E2E run, these variables need to be set in `configurations/default/env.yml`.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,167 @@
...
+     * Ensure that an {@link AdminUser} can delete an API key for an {@link ApiUser}.
+     */
+    @Test
+    public void adminCanDeleteApiKeyForApiUser() {
+        assumeTrue(getBooleanEnvVar("RUN_E2E"));
+        ensureAtLeastOneApiKeyIsAvailable();
+        // delete key
+        String keyId = apiUser.apiKeys.get(0).id;
+        HttpResponse<String> response = deleteApiKeyRequest(apiUser.id, keyId, adminUser.auth0UserId);
+        Assertions.assertEquals(response.statusCode(), 200);
```
binh-dam-ibigroup: It looks like you swapped the arguments. 200 should be the first arg.

configurations/test/env.yml
binh-dam-ibigroup: `DISABLE_AUTH=true` doesn't seem to be needed in this file (it is in the user's own `configurations/default/env.yml` for E2E tests.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
binh-dam-ibigroup: I would just write `assertEquals(...)` and import `Assertions` with an `import static org.junit.jupiter.api.Assertions.assertEquals;`

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -1,27 +1,167 @@
...
+        ApiUser userFromResponse = JsonUtils.getPOJOFromJSON(response.body(), ApiUser.class);
+        Assertions.assertTrue(userFromResponse.apiKeys.isEmpty());
```
binh-dam-ibigroup: same for `assertTrue`.

binh-dam-ibigroup (approved):
Conditional approval: swap arguments for one assertion, and clarify the comment about in which configuration to set the required E2E config parameters. (and stylistic suggestions).

landonreed (approved):

PR #56, opened 2020-08-11
Remove Spring Framework dependency

binh-dam-ibigroup:
### Checklist

- [X] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [na] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [X] All tests and CI builds passing

### Description

A hold out from the dead code removal... this PR removes dependencies to spring framework and to Gson.
The following entries are added that were previously imported by spring:
- resource encoding
- logging (slf4j and logback)

PR #59, opened 2020-08-17
Public API docs

binh-dam-ibigroup:
### Checklist

### Description (UPDATED)

This PR automates the generation and deployment of the public-facing API documentation for OTP Middleware.
A public-facing swagger.yaml document is generated during server startup. That way, documentation clients always have the most current version of the documentation.

For testing, the swagger YAML is published to https://otp-middleware-builds.s3.amazonaws.com/public-api-docs/otp-middleware-swagger.yaml. You can view an example swagger documentation at https://fdot-support.s3.amazonaws.com/index.html?url=https://otp-middleware-builds.s3.amazonaws.com/public-api-docs/otp-middleware-swagger.yaml.


pom.xml
```diff
@@ -32,6 +32,10 @@
             </resource>
         </resources>
         <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.2</version>
+            </plugin>
```
landonreed: Do we know how this got deleted previously? In any case, it's needed for tests, right? We should add a comment that describes that it is used to run tests in the code during build/package.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -38,14 +44,24 @@
...
-        spark.get(OTP_PROXY_ENDPOINT + "/*", OtpRequestProcessor::proxy);
+    @Override
+    public void bind(final SparkSwagger restApi) {
+        restApi.endpoint(
+            endpointPath(OTP_PROXY_ENDPOINT).withDescription("Proxy interface for OTP endpoints. " + OTP_DOC_MOREINFO),
+            (q, a) -> LOG.info("Received request for the OTP proxy endpoint.")
+        ).get(path("/*")
+                .withDescription("Forwards any GET request to OTP. " + OTP_DOC_MOREINFO)
+                .withProduces(List.of(APPLICATION_JSON, APPLICATION_XML)),
+            OtpRequestProcessor::proxy);
```
landonreed: I find this block's formatting very hard to read. Also, I think the `LOG.info` is basically on every controller, but is that something we really want to happen on each request?

src/main/resources/doc.yaml
```diff
@@ -12,38 +12,75 @@ info:
   license:
     name: "MIT License"
     url: "https://opensource.org/licenses/MIT"
-host: "localhost:4567"
-basePath: "/"
+host: "w7tqwg7j5a.execute-api.us-east-1.amazonaws.com"
+basePath: "/dev1"
```
landonreed: I'm not sure we should be committing these specific values in this source-controlled file.

src/main/resources/api-gateway-template.yaml
```diff
@@ -0,0 +1,59 @@
...
+        description: "Email address for contact. This must be unique in the collection."
+      auth0UserId:
+        type: "string"
+        description: "Auth0 user name."
+      isDataToolsUser:
+        type: "boolean"
+        description: "Determines whether this user has access to OTP Data Tools."
+    description: "An abstract user."
+paths:
+  /pelias/*:
```
landonreed: This should probably have a comment describing that this path must be a proxy route configured in API Gateway.

src/main/resources/doc.yaml
```diff
@@ -141,19 +255,14 @@ paths:
...
-      parameters: []
-      responses:
-        "200":
-          description: "successful operation"
   /api/secure/application/{id}:
     get:
       tags:
       - "api/secure/application"
-      description: "Returns a 'apiuser' entity with the specified id, or 404 if not\
+      description: "Returns a 'ApiUser' entity with the specified id, or 404 if not\
```
landonreed: The other endpoints (delete, update) say "Returns the" instead of "Returns a". I think we should update all of the GET endpoints to also say "Returns the".

src/main/resources/spark-swagger.conf
```diff
@@ -19,11 +19,11 @@ spark-swagger {
   # API related configs
 
   # Location where API docs will be published
-  host = "localhost:4567"
+  host = "w7tqwg7j5a.execute-api.us-east-1.amazonaws.com"
 
   # basepath for calling the API, if applicable, otherwise leave as "/"
   # as in https://example.api.com/<basepath>/user/all
-  basePath = "/"
+  basePath = "/dev1"
```
landonreed: As mentioned above, if possible I think these values should be handled in the config file, not committed to the source code.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+
+        tagsNode.removeAll();
+        tagsNode.addAll(newTagsNode);
+    }
+
+    /**
+     * Determines whether a given path is restricted (non-public).
+     */
+    private boolean isRestricted(String pathName) {
+        for (String restrictedPath: RESTRICTED_PATHS) {
```
landonreed: needs whitespace after colon.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+ * Contains test to verify that the swagger docs are generated correctly.
+ */
+public class SwaggerTest extends OtpMiddlewareTest {
+    private static final Logger LOG = LoggerFactory.getLogger(SwaggerTest.class);
+    private static final HttpClient client = HttpClient.newBuilder().build();
+    private static final Path versionControlledSwaggerFile = new File("src/main/resources/doc.yaml").toPath();
+    private static final File apiGatewayDefinitionsFile = new File("src/main/resources/api-gateway-template.yaml");
+    private static final String[] RESTRICTED_PATHS = new String[] {
+        "api/admin/", "api/secure/application", "api/secure/logs"
+    };
```
landonreed: I wonder if it would be safer to reverse this and have a set of public paths. That way, when a new private path is added to the API, it doesn't accidentally end up in the public docs.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+     */
+    @Test @Disabled
+    public void updateSwaggerDocs() throws IOException, InterruptedException {
+        String autoGeneratedSwagger = getSwaggerDocsAsString();
+        Files.writeString(versionControlledSwaggerFile, autoGeneratedSwagger);
+    }
+
+    /**
+     * Generate enhanced Swagger docs for use with AWS API Gateway. This handles things like adding auth headers and key
+     * parameters used with API Gateway as well as injecting info needed to make CORS work.
```
landonreed: This comment needs to be updated. We're no longer injecting CORS and not trying to generate a file for import into API Gateway. The name of the method should probably change to `generatePublicApiDocs` or something.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+        String geocoderTag = peliasTemplate.at("/get/tags").get(0).asText();
+
+        // Insert geocoder tag.
+        ((ArrayNode)rootNode.get("tags")).add(JsonNodeFactory.instance.objectNode()
+            .put("name", geocoderTag)
+            .put("description", "The Pelias Geocoder interface. Refer to <a href='https://github.com/pelias/documentation/#endpoint-descriptions'>Pelias Geocoder Documentation</a> for API resources supported by Pelias Geocoder.")
+        );
+
+        // Insert Pelias geocoder template in the paths.
+        ObjectNode pathsNode = (ObjectNode) rootNode.get("paths");
```
landonreed: Why is this ObjectNode cast as an ObjectNode? Perhaps this is leftover code?

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+     */
+    private void insertPeliasGeocoderEndpoint(ObjectNode rootNode, JsonNode templatePathsNode) {
+        String geocoderPath = getFieldNames(templatePathsNode, path -> path.contains("pelias")).get(0);
+        JsonNode peliasTemplate = templatePathsNode.get(geocoderPath);
+        String geocoderTag = peliasTemplate.at("/get/tags").get(0).asText();
+
+        // Insert geocoder tag.
+        ((ArrayNode)rootNode.get("tags")).add(JsonNodeFactory.instance.objectNode()
+            .put("name", geocoderTag)
+            .put("description", "The Pelias Geocoder interface. Refer to <a href='https://github.com/pelias/documentation/#endpoint-descriptions'>Pelias Geocoder Documentation</a> for API resources supported by Pelias Geocoder.")
```
landonreed: Isn't this description already in the template file? I'm a bit confused here.

landonreed (changes requested):
See comments for suggestions. Also, per discussion on company chat, we should probably be serving this file from the application.

src/main/resources/doc.yaml
evansiroky: Yeah, definitely not.

src/main/resources/spark-swagger.conf
```diff
@@ -19,11 +19,11 @@ spark-swagger {
...
   docPath = "/doc"
```
evansiroky: Does this make the `http://localhost:4567/doc.yaml` endpoint work? If so, can you add a comment about it?

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,393 @@
...
+        HttpResponse<String> swaggerResponse = getSwaggerDocs();
+        assertEquals(swaggerResponse.statusCode(), 200);
+        String autoGeneratedSwagger = swaggerResponse.body();
+        JsonNode swaggerJson = YamlUtils.yamlMapper.readTree(autoGeneratedSwagger);
+        String title = swaggerJson.get("info").get("title").asText();
+        LOG.info("Found swagger docs title: {}", title);
+        assertEquals(title, "OTP Middleware");
+        String versionControlledSwagger = Files.readString(versionControlledSwaggerFile);
+        assertEquals(autoGeneratedSwagger, versionControlledSwagger);
+    }
```
evansiroky: It seems like this is the only actual test in this file. Most of the code in this file is for actually generating the docs. It seems like putting this in a test is the wrong place for this since it is actual code that performs a task instead of code that verifies a task is done correctly. Therefore, I think every method besides this one should be put into another file in the `src/main/...` directory instead. And the way to initiate documentation generation shouldn't be via running tests, instead it would be better to have some kind of main method that could parse arguments for certain settings of building the docs.

evansiroky (changes requested):
My primary concern is stated here: https://github.com/ibi-group/otp-middleware/pull/59#discussion_r472606190

src/main/resources/spark-swagger.conf
binh-dam-ibigroup: spark-swagger puts the auto-generated, non-edited yaml file at `http://localhost:4567/doc.yaml`, and that has nothing to do with the `host` or `basePath` attribute. (`host` or `basePath` are written into the document, so that readers can try out making API calls to that API Gateway similar to how they'd do it using tools such as Postman).

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: This is the description of the `tag` element, which is what you see on the document when you collapse the header under which all endpoint methods are listed. It is separate from the description of the GET method in the template file. What I could do is to copy the description of the GET method from the template file into the tag description.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: `rootNode.get(...)` is of type `JsonNode`.

pom.xml
binh-dam-ibigroup: It may be one of the implicit dependencies of the SpringFramework that needed to be added back when we did #56. Will add comment.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
binh-dam-ibigroup: I agree we don't need that log entry every time. That argument is actually a filter where we could scan request content and stop execution early if we wanted to (for another refactor?).

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: I ended up moving the tag description to the template file.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -24,6 +25,8 @@
 
     public static final List<String> JSON_ONLY = List.of(APPLICATION_JSON);
 
+    public static final Filter NO_FILTER = (q, a) -> {};
```
landonreed: Please give this a comment and/or change the variable name to something a little more obvious.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -54,312 +36,16 @@ public void updateSwaggerDocs() throws IOException, InterruptedException {
     }
 
     /**
-     * Generate enhanced Swagger docs for use with AWS API Gateway. This handles things like adding auth headers and key
-     * parameters used with API Gateway as well as injecting info needed to make CORS work.
+     * Convenience test to generate the docs.
```
landonreed: Does this need to be a test? It doesn't seem to be testing anything...

binh-dam-ibigroup: Hehe no, I'm still working on this.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,375 @@
...
+ */
+public class PublicApiDocGenerator {
+    private static final Logger LOG = LoggerFactory.getLogger(PublicApiDocGenerator.class);
+    private static final HttpClient client = HttpClient.newBuilder().build();
+    private static final File apiTemplateDefinitionsFile = new File("src/main/resources/api-swagger-template.yaml");
+    private static final String[] PUBLIC_PATHS = new String[] {
+        "api/secure/monitoredtrip", "api/secure/triprequests", "api/secure/user", "otp", "pelias"
+    };
+    public static final String DEFINITIONS_REF = "#/definitions/";
+    public static final String $REF_FIELD = "$ref";
```
landonreed: Why does this variable start with a `$` that seems a bit heterodox.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,375 @@
...
+import java.util.function.Predicate;
+
+/**
+ * Class that generates the public-facing API documentation.
+ */
+public class PublicApiDocGenerator {
+    private static final Logger LOG = LoggerFactory.getLogger(PublicApiDocGenerator.class);
+    private static final HttpClient client = HttpClient.newBuilder().build();
+    private static final File apiTemplateDefinitionsFile = new File("src/main/resources/api-swagger-template.yaml");
+    private static final String[] PUBLIC_PATHS = new String[] {
```
landonreed, line 50: Might be good to give `PUBLIC_PATHS` a comment suggesting that any new public endpoints should be added here.

binh-dam-ibigroup:
I broke the API controller route in this PR... Adding the BLOCKED tag.

src/main/resources/spark-swagger.conf
```diff
@@ -19,12 +19,14 @@ spark-swagger {
...
+  # (will be overwritten by PublicApiDocGenerator for the public-facing swagger document).
+  host = "AWS_API_SERVER"
 
   # basepath for calling the API, if applicable, otherwise leave as "/"
-  # as in https://example.api.com/<basepath>/user/all
-  basePath = "/"
-  docPath = "/doc"
+  # as in https://example.api.com<basepath>/user/all
+  # (will be overwritten by PublicApiDocGenerator for the public-facing swagger document).
+  basePath = "/AWS_API_STAGE"
```
binh-dam-ibigroup: This change to a value other than `"/"` broke the endpoints. spark-swagger uses the initial value of this parameter to compose the final path of all endpoints.

binh-dam-ibigroup:
This PR is ready again.

.travis.yml
```diff
@@ -41,9 +41,10 @@ before_deploy:
 - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH;
   else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
 - export BRANCH_CLEAN=${BRANCH//\//_}
-# `git describe --tags HEAD` returns a string of the form v0.0.0-52-ge10d02d.
+# `git describe --tags --always HEAD` returns a string of the form v0.0.0-52-ge10d02d.
 # It assumes you have pushed a tag on a commit on github (e.g. a commit on the dev branch).
-- export JAR_VERSION=$(git describe --tags HEAD)
+# If for some reason git can't fine a tag, fallback with --always to a commit sha.
```
evansiroky: typo: `fine` should be `find`

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Run this test to update the Swagger docs file located at {@link #versionControlledSwaggerFile}. This is disabled
+     * because it serves as a utility for keeping the docs file up-to-date.
+     */
+    @Test @Disabled
+    public void updateSwaggerDocs() throws IOException, InterruptedException {
+        String autoGeneratedSwagger = getSwaggerDocsAsString();
+        Files.writeString(versionControlledSwaggerFile, autoGeneratedSwagger);
+    }
```
evansiroky: This test should be removed. It is providing a mechanism to perform a task, not to verify something. Furthermore, it would update something in version control.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+
+    /**
+     * Removes unused types (not referenced in a $ref entry anywhere) from the definitions section.
+     */
+    private void removeUnusedTypes() {
+        int unusedTypeCount;
+        // Execute until there are no unused types.
+        do {
+            // Find return type references anywhere (only store one instance of each).
+            LinkedHashSet<String> typeReferences = new LinkedHashSet<>();
```
evansiroky: I learned a new thing about what a LinkedHashSet is, but do we really need it here? Wouldn't a regular HashSet suffice?

evansiroky (changes requested):
See new comments.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: Sure... but where should that code go instead?

binh-dam-ibigroup: Or @evansiroky do you think this should be a manual process?

binh-dam-ibigroup: I think I found a workable way to do that in https://github.com/ibi-group/otp-middleware/pull/59/commits/28533eecedcccc7e6e4afae2faf33a8c5753ee77

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+ * to update the version-controlled spark-swagger output.
+ */
+public class VersionControlledSwaggerUpdater {
+    public static final String SWAGGER_CONTROLLED_FILE = "src/main/resources/spark-swagger-controlled-output.yaml";
+
+    private static final Logger LOG = LoggerFactory.getLogger(VersionControlledSwaggerUpdater.class);
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        OtpMiddlewareMain.main(new String[0]);
+        updateSwaggerDocs();
```
evansiroky: This is getting closer, but is it necessary to start OtpMiddlewareMain here? Can the docs be retrieved directly from the `PublicApiDocGenerator` class and written to a file without starting up a server?

evansiroky (changes requested):
I like where this is headed, but see https://github.com/ibi-group/otp-middleware/pull/59#discussion_r478778358.

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
binh-dam-ibigroup: The server has to be run to accomplish this task. We use the swagger generated at runtime as skeleton to create the public-facing version of it, at runtime too.

evansiroky: Very well, I'll approve then.

evansiroky (approved):

configurations/default/env.yml.tmp
```diff
@@ -10,10 +10,9 @@ BUGSNAG_ORGANIZATION: organization
 BUGSNAG_PROJECT_NOTIFIER_API_KEY: project-notifier-api-key
 #BUGSNAG_REPORTING_WINDOW_IN_DAYS: 14
 
-AUTH0_DOMAIN: your_domain.auth0.com
-AUTH0_API_CLIENT: your-api-client-id
-AUTH0_API_SECRET: your-api-secret-id
 AWS_PROFILE: default
+AWS_API_SERVER: aws-api-id.execute-api.us-east-1.amazonaws.com
+AWS_API_STAGE: stage-name
```
landonreed, line 17: I think these items should be commented/clarified as non-essential to the functioning of the application (and only used for generating swagger docs).

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -63,25 +66,37 @@ private static void initializeHttpEndpoints() throws IOException {
...
-            OtpRequestProcessor.register(spark);
         } catch (RuntimeException e) {
             LOG.error("Error initializing API controllers", e);
             System.exit(1);
         }
+
+        // Generate the public facing API docs after startup,
+        // and create an undocumented endpoint to serve the document.
+        Path publicDocPath = new PublicApiDocGenerator().generatePublicApiDocs();
+        spark.get("/publicapi", (request, response) -> {
```
landonreed: Would prefer this endpoint to be at `/public` or `/docs`. Probably just `/docs`. Or if there's a convention in the swagger world for where this file is hosted, that'd be good too.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -63,25 +66,37 @@ private static void initializeHttpEndpoints() throws IOException {
...
+        });
+
+        // Generic response for all OPTIONS requests on all endpoint paths.
         spark.options("/*",
             (request, response) -> {
                 logMessageAndHalt(request, HttpStatus.OK_200, "OK");
                 return "OK";
             });
 
+        // Security checks for admin and /secure/ endpoints.
```
landonreed, line 99: Digging the extra comments.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -46,10 +47,11 @@
     protected static final String ID_PATH = "/:" + ID_PARAM;
     protected final String ROOT_ROUTE;
     private static final String SECURE = "secure/";
+    private final String className;
     private static final Logger LOG = LoggerFactory.getLogger(ApiController.class);
-    private final String classToLowercase;
     final TypedPersistence<T> persistence;
     private final Class<T> clazz;
+    private final T[] typedArray;
```
landonreed: I don't think this needs to be a field. Couldn't you just pass `Array.newInstance(clazz, 0).getClass()` directly to `withResponseType`?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -111,17 +116,18 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
...
-                    .withDescription("Gets a list of all '" + classToLowercase + "' entities.")
+                    .withDescription("Gets a list of all '" + className + "' entities.")
                     .withProduces(JSON_ONLY)
-                    // Note: unlike what the name suggests, withResponseAsCollection does not generate an array
-                    // as the return type for this method. (It does generate the type for that class nonetheless.)
-                    .withResponseAsCollection(clazz),
+                    // Set the return type as the array of clazz objects.
+                    // Note: there exists a method withResponseAsCollection, but unlike what its name suggests,
+                    // it does exactly the same as .withResponseType and does not generate a return type array.
+                    .withResponseType(typedArray.getClass()),
```
landonreed: Please add URL to the issue you created in the spark-swagger project.

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.Path;
+
+import static org.opentripplanner.middleware.docs.PublicApiDocGenerator.getSwaggerDocsAsString;
+
+/**
+ * This class is solely intended to be executed on a development machine
+ * to update the version-controlled spark-swagger output.
+ */
+public class VersionControlledSwaggerUpdater {
+    public static final String SWAGGER_CONTROLLED_FILE = "src/main/resources/spark-swagger-controlled-output.yaml";
```
landonreed: Would prefer this variable and file name to be something like `LATEST_SWAGGER_FILE`/`src/main/resources/latest-spark-swagger-output.yaml`

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
landonreed: Since we're having a discussion about this, there should be a comment saying why middleware main is called. But, to @evansiroky's point, couldn't you just run `initializeHttpEndpoints`?

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        OtpMiddlewareMain.main(new String[0]);
+        updateSwaggerDocs();
+
+        System.exit(0);
+    }
+
+    /**
+     * Performs an update to the Swagger docs file located at {@link #SWAGGER_CONTROLLED_FILE},
```
landonreed: Instead of `Performs an update to...` just say `Updates...`.

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,45 @@
...
+        String autoGeneratedSwagger = swaggerResponse.body();
+        JsonNode swaggerJson = YamlUtils.yamlMapper.readTree(autoGeneratedSwagger);
+        String title = swaggerJson.get("info").get("title").asText();
+        LOG.info("Found swagger docs title: {}", title);
+        assertEquals(title, "OTP Middleware");
+
+        Path versionControlledSwaggerFile = new File(SWAGGER_CONTROLLED_FILE).toPath();
+        String versionControlledSwagger = Files.readString(versionControlledSwaggerFile);
+        assertEquals(autoGeneratedSwagger, versionControlledSwagger,
+            "If you modified any endpoint or API configuration, remember to run the VersionControlledSwaggerUpdater class to update the version-controlled swagger.");
```
landonreed: `... remember to run VersionControlledSwaggerUpdater#main to update...`

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Contains test to verify that the swagger docs are generated correctly.
+ */
+public class SwaggerTest extends OtpMiddlewareTest {
+    private static final Logger LOG = LoggerFactory.getLogger(SwaggerTest.class);
+
+    /**
+     * Verify that version-controlled doc.yaml file is up to date with the latest changes to swagger file auto-generated
+     * from the spark-swagger library (available at {@link PublicApiDocGenerator#getSwaggerDocs()}.
```
landonreed: `Verify that {@link SWAGGER_CONTROLLED_FILE} matches the latest changes to the API as generated by spark-swagger (available at {@link PublicApiDocGenerator#getSwaggerDocs()}. If there are changes to the API, run {@link VersionControlledSwaggerUpdater#main} to update the file.`

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.net.http.HttpResponse;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.opentripplanner.middleware.docs.PublicApiDocGenerator.getSwaggerDocs;
+import static org.opentripplanner.middleware.docs.VersionControlledSwaggerUpdater.SWAGGER_CONTROLLED_FILE;
+
+/**
+ * Contains test to verify that the swagger docs are generated correctly.
```
landonreed: addition to comment: `... and up to date.`

src/main/resources/spark-swagger.conf
```diff
@@ -19,12 +19,14 @@ spark-swagger {
...
 
   # Location where API docs will be published
+  # PublicApiDocGenerator will overwrite this for the public-facing swagger document.
   host = "localhost:4567"
 
-  # basepath for calling the API, if applicable, otherwise leave as "/"
-  # as in https://example.api.com/<basepath>/user/all
+  # basePath must be "/" so that endpoints are in the form localhost:4567/endpoints.
+  # PublicApiDocGenerator will overwrite this for the public-facing swagger document.
+  # (Enter a path other than "/" to make if you want all endpoints to be in the form localhost:4567/basepath/endpoints.)
```
landonreed: There are two comments about `PublicApiDocGenerator will overwrite this`, but I have no idea what "this" is.

src/main/resources/api-swagger-template.yaml
```diff
@@ -0,0 +1,67 @@
+securityDefinitions:
```
landonreed: File needs a comment explaining its purpose.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -22,8 +23,17 @@
...
 
+    /**
+     * A constant for a list of MIME types containing application/json only.
+     */
     public static final List<String> JSON_ONLY = List.of(APPLICATION_JSON);
 
+    /**
+     * A constant method compliant with {@link:spark.Filter} that does nothing,
+     * but that is needed as a parameter for the spark-swagger endpoint definition method.
+     */
```
landonreed, line 34: 👍

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -24,7 +31,7 @@
  * is of interest the response is intercepted and processed. In all cases, the response from OTP (content and HTTP status)
  * is passed back to the requester.
  */
-public class OtpRequestProcessor {
+public class OtpRequestProcessor implements Endpoint {
```
landonreed, line 36: Should this class move to the `controllers` package? That seems like a more appropriate home for it at this point.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+        LOG.info("Wrote public API documentation to: {}", outputPath);
+
+        return outputPath;
+    }
+
+    /**
+     * Extracts the version number from the JAR file name
+     * (modified from https://stackoverflow.com/questions/14189162/get-name-of-running-jar-or-exe#19045510).
+     */
+    private String getVersionFromJar() {
```
landonreed: This method can be static.

I'm not sure we should rely on the jar file for this. See Data Tools for a method that shows how it extracts git properties https://github.com/ibi-group/datatools-server/blob/9f74b821cf351efcdaf7c9c93a3ae8b694d3c3b1/src/main/java/com/conveyal/datatools/manager/DataManager.java#L181-L212. This can be non-blocking, but it might be worth adding a TODO about this. Also, we may want to share this info elsewhere in the app, so it might be worth moving this to ConfigUtils.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+        tagsNode = (ArrayNode) rootNode.get("tags");
+
+        // Load definitions from a template file.
+        templateJson = YamlUtils.yamlMapper.readTree(OtpMiddlewareMain.class.getClassLoader().getResourceAsStream(API_TEMPLATE_FILE));
+    }
+
+    /**
+     * Generate an enhanced public-facing Swagger document from the one auto-generated by spark-swagger.
+     * This handles things like adding security parameters, method responses, and removing restricted API resources
+     * from the public-facing document.
```
landonreed: "Public-facing" should be clarified a bit more. 

```
Generate an enhanced public-facing (i.e., the collection of endpoints available to an {@link ApiUser}, which excludes admin endpoints like viewing application errors or request logs) Swagger document....
```

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
+
+/**
+ * Class that generates the public-facing API documentation.
+ */
+public class PublicApiDocGenerator {
+    private static final Logger LOG = LoggerFactory.getLogger(PublicApiDocGenerator.class);
+    private static final HttpClient client = HttpClient.newBuilder().build();
+    private static final String API_TEMPLATE_FILE = "api-swagger-template.yaml";
```
landonreed: Let's call this `SWAGGER_SUPPLEMENTS_FILE`/`swagger-supplements.yaml`

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+
+    private static final String AWS_API_SERVER = getConfigPropertyAsText("AWS_API_SERVER");
+    private static final String AWS_API_STAGE = getConfigPropertyAsText("AWS_API_STAGE");
+
+    // Cached variables.
+    private final ObjectNode rootNode;
+    private final JsonNode templateJson;
+    private final ObjectNode pathsNode;
+    private final ObjectNode definitionsNode;
+    private final ArrayNode tagsNode;
```
landonreed: These variables don't need the `Node` suffix. That's like calling an array of strings `stringsArray` instead of `strings`. Recommend changing:
- rootNode -> swaggerRoot
- templateJson -> supplements
- pathsNode -> paths
- defintionsNode -> defitinitions
- tagsNode -> tags

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+
+            unusedTypeCount = typesToRemove.size();
+        } while(unusedTypeCount > 0);
+    }
+
+    /**
+     * Insert the definition for {@link:AbstractUser}
+     * and add fields (with description) from AbstractUser to OtpUser.
+     * (spark-swagger ignores parent classes when generating swagger.)
+     * TODO: Find a better way to include parent classes.
```
landonreed, line 366: Might be good to mention that this skips the other user child classes because they are not "public-facing".

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
```diff
@@ -0,0 +1,416 @@
...
+        String autoGeneratedSwagger = getSwaggerDocsAsString();
+
+        // The document we are working off and related nodes.
+        rootNode =  YamlUtils.yamlMapper.readTree(autoGeneratedSwagger).deepCopy();
+        definitionsNode = (ObjectNode) rootNode.get("definitions");
+        pathsNode = (ObjectNode) rootNode.get("paths");
+        tagsNode = (ArrayNode) rootNode.get("tags");
+
+        // Load definitions from a template file.
+        templateJson = YamlUtils.yamlMapper.readTree(OtpMiddlewareMain.class.getClassLoader().getResourceAsStream(API_TEMPLATE_FILE));
```
landonreed: Definitions not loaded at this point (comment needs updating). Also, I think this could be moved from a field on the class to a local variable in `generatePublicApiDocs`.

landonreed (changes requested):
This looks really good. A handful of comments, but they're mostly about comments.

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
binh-dam-ibigroup: Unfortunately, `initializeHttpEndpoints` is not enough. There are persistence classes involved which must be also initialized, so better run `OtpMiddleware.main` as in the tests.

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

configurations/default/env.yml.tmp
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/test/java/org/opentripplanner/middleware/docs/SwaggerTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/resources/spark-swagger.conf
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/resources/api-swagger-template.yaml
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/docs/VersionControlledSwaggerUpdater.java
binh-dam-ibigroup: I added a comment about this in https://github.com/ibi-group/otp-middleware/pull/59/commits/b327a799b261096bd67fcaf0d680d92b1090fc2a

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Addressed in https://github.com/ibi-group/otp-middleware/pull/59/commits/4e3253e4b58640c8a42db4a24d02ac2c3319a3a7

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Good catch. Addressed in https://github.com/ibi-group/otp-middleware/pull/59/commits/4e3253e4b58640c8a42db4a24d02ac2c3319a3a7

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Addressed  in https://github.com/ibi-group/otp-middleware/pull/59/commits/4e3253e4b58640c8a42db4a24d02ac2c3319a3a7

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Method moved and comment updated in https://github.com/ibi-group/otp-middleware/pull/59/commits/aee52b65a0058fc403f8e00a8c7699babd0398ed

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
binh-dam-ibigroup: Moved in https://github.com/ibi-group/otp-middleware/pull/59/commits/aee52b65a0058fc403f8e00a8c7699babd0398ed

src/main/java/org/opentripplanner/middleware/docs/PublicApiDocGenerator.java
binh-dam-ibigroup: Method moved and TODO added in https://github.com/ibi-group/otp-middleware/pull/59/commits/4e3253e4b58640c8a42db4a24d02ac2c3319a3a7

landonreed (approved):
Looks perfect.

PR #64, opened 2020-08-27
Cascade delete

br648:
### Checklist

### Description

Tests to simulate API user flow. The following config parameters must be set in configurations/default/env.yml for these end-to-end tests to run: 

- AUTH0_DOMAIN set to a valid Auth0 domain 
- AUTH0_API_CLIENT set to E2E Test client id (see Auth0)
- AUTH0_API_SECRET set to E2E Test client secret (see Auth0)
- DEFAULT_USAGE_PLAN_ID set to a valid usage plan id (AWS requires this to create an api key). 
- OTP_API_ROOT set to `http://localhost:8080/otp` so the mock OTP server is used 
- OTP_PLAN_ENDPOINT set to `/routers/default/plan`
- An AWS_PROFILE is required, or AWS access has been configured for your operating environment e.g. C:\Users\<username>\.aws\credentials in Windows or Mac OS equivalent.
 
 The following environment variable must be set for these tests to run: 
- `RUN_E2E: true`

Auth0 must be correctly configured as described here: https://auth0.com/docs/flows/call-your-api-using-resource-owner-password-flow.


src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -0,0 +1,169 @@
...
+public class ApiUserFlowTest {
+    private static ApiUser apiUser;
+    private static OtpUser otpUser;
+
+    /**
+     * Create an {@link ApiUser} and an {@link AdminUser} prior to unit tests
+     */
+    @BeforeAll
+    public static void setUp() throws IOException {
+        OtpMiddlewareTest.setUp();
```
evansiroky, line 62: Needs a `assumeTrue(getBooleanEnvVar("RUN_E2E"));` on the first line of this method so that this only runs during E2E testing.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+    }
+
+    /**
+     * Delete the users if they were not already deleted during the test script.
+     */
+    @AfterAll
+    public static void tearDown() {
+        apiUser = Persistence.apiUsers.getById(apiUser.id);
```
evansiroky, line 95: Needs a `assumeTrue(getBooleanEnvVar("RUN_E2E"));` on the first line of this method so that this only runs during E2E testing.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -47,7 +44,6 @@
     @BeforeAll
     public static void setUp() throws IOException {
         OtpMiddlewareTest.setUp();
```
evansiroky, line 49: Needs a `assumeTrue(getBooleanEnvVar("RUN_E2E"));` on the first line of this method so that this only runs during E2E testing.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -59,13 +55,9 @@ public static void setUp() throws IOException {
     public static void tearDown() {
         // Delete admin user.
```
evansiroky, line 63: Needs a `assumeTrue(getBooleanEnvVar("RUN_E2E"));` on the first line of this method so that this only runs during E2E testing.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -150,6 +155,34 @@ public T getById(String id) {
         return mongoCollection.find(filter).limit(maximum).into(new ArrayList<>());
     }
 
+    /**
+     * Build a filter for querying Mongo based on userId and from/to dates.
+     */
+    public static Bson buildFilter(String userId, Date fromDate, Date toDate) {
```
evansiroky: Please rename to something more descriptive like `filterByUserAndDate`.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -150,6 +155,34 @@ public T getById(String id) {
...
+        if (toDate != null) {
+            clauses.add(lte("dateCreated", toDate));
+        }
+        return Filters.and(clauses);
+    }
+
+    /**
+     * Build a filter for querying Mongo based on userId.
+     */
+    public static Bson buildFilter(String userId) {
```
evansiroky: Please rename to something more descriptive like `filterByUserId`.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
```diff
@@ -61,11 +77,18 @@ private static String proxy(Request request, spark.Response response) {
...
-        OtpDispatcherResponse otpDispatcherResponse = OtpDispatcher.sendOtpRequest(request.queryString(), otpRequestPath);
-        if (otpDispatcherResponse == null || otpDispatcherResponse.responseBody == null) {
-            logMessageAndHalt(request, HttpStatus.INTERNAL_SERVER_ERROR_500, "No response from OTP server.");
-            return null;
+
+        OtpDispatcherResponse otpDispatcherResponse;
+        // If the request is to the plan end point and E2E is enabled mock the plan response
+        if (otpRequestPath.endsWith(OTP_PLAN_ENDPOINT) && runE2E) {
+            otpDispatcherResponse = getMockPlanResponse();
+        } else {
```
evansiroky: I believe that in general it is a good practice to minimize the amount of test-specific code that is present in the regular code. It seems like the purpose of this is to return custom OTP responses and not rely on a live OTP server. A better approach might be to have the test code setup a local OTP instance at a certain port and then control what responses that server sends back.

evansiroky: Maybe do something like this: https://stackoverflow.com/a/43176400/269834

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -224,4 +236,33 @@ private static String getAuth0Url() {
             ? "http://locahost:8089"
             : "https://" + AUTH0_DOMAIN;
     }
+
+    /**
+     * Get an 0Auth token by using the Auth0 'Call Your API Using Resource Owner Password Flow' approach. Auth0 setup
```
evansiroky: Typo `0Auth` should be `Auth0`.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -224,4 +236,33 @@ private static String getAuth0Url() {
...
+     * can be reviewed here: https://auth0.com/docs/flows/call-your-api-using-resource-owner-password-flow. If the user
+     * is successfully validated by Auth0 a bearer access token is returned, which is extracted and returned to the
+     * caller. In all other cases, null is returned.
+     */
+    public static String get0AuthToken(String username, String password) {
```
evansiroky: Typo `get0AuthToken` should be `getAuth0Token`.

evansiroky (changes requested):
See comments. Also, I'm having trouble running the E2E tests locally. Let's discuss my configuration details off Github.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/otp/OtpRequestProcessor.java
br648: @evansiroky agree. I have implemented a second Spark server to mock the OTP for review.

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -34,4 +60,69 @@ public static boolean getBooleanEnvVar(String var) {
...
+     * disabled, use Auth0 user ID to authenticate else attempt to get a valid 0auth token from Auth0 and use this.
+     */
+    private static HashMap<String, String> getMockHeaders(AbstractUser requestingUser) {
+        HashMap<String, String> headers = new HashMap<>();
+        // If auth is disabled, simply place the Auth0 user ID in the authorization header, which will be extracted from
+        // the request when received.
+        if ("true".equals(OtpMiddlewareMain.getConfigPropertyAsText("DISABLE_AUTH"))) {
+            headers.put("Authorization", requestingUser.auth0UserId);
+        } else {
+            // Otherwise, get a valid oauth token for the user
```
binh-dam-ibigroup, line 92: Typo: oauth should be Auth0.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -74,8 +68,8 @@ public static void tearDown() {
     @Test
     public void canCreateApiKeyForSelf() {
         assumeTrue(getBooleanEnvVar("RUN_E2E"));
-        HttpResponse<String> response = createApiKeyRequest(apiUser.id, apiUser.auth0UserId);
-        assertEquals(200, response.statusCode());
+        HttpResponse<String> response = createApiKeyRequest(apiUser.id, apiUser);
```
binh-dam-ibigroup, line 76: I think this piece broke, it was previously containing the added API key, and with this change it is not.

binh-dam-ibigroup (changes requested):
The mock `createApiKeyRequest` call and something in `canSimulateApiUserFlow` broke with the changes introduced. (On the dev branch the tests are working fine for me.)

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -11,24 +11,30 @@
...
+import static org.opentripplanner.middleware.utils.JsonUtils.getSingleNodeValueFromJSON;
 import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
 
 /**
  * This class contains methods for querying Auth0 users using the Auth0 User Management API. Auth0 docs describing the
  * searchable fields and query syntax are here: https://auth0.com/docs/api/management/v2/user-search
  */
 public class Auth0Users {
-    private static final String AUTH0_DOMAIN = getConfigPropertyAsText("AUTH0_DOMAIN");
+    public static final String AUTH0_DOMAIN = getConfigPropertyAsText("AUTH0_DOMAIN");
```
binh-dam-ibigroup, line 38: What is the purpose of making this public?

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -216,7 +217,7 @@ private T getOne(Request req, Response res) {
     /**
      * HTTP endpoint to delete one entity specified by ID.
      */
-    private String deleteOne(Request req, Response res) {
+    private T deleteOne(Request req, Response res) {
```
binh-dam-ibigroup, line 223: With this change in return type, the endpoint documentation above probably needs to be updated.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -12,6 +15,15 @@ public OtpUserController(String apiPrefix) {
         super(apiPrefix, Persistence.otpUsers, "secure/user");
     }
 
+    @Override
+    OtpUser preCreateHook(OtpUser user, Request req) {
+        // Check API key and assign user to appropriate collection.
```
binh-dam-ibigroup: This comment may need to be updated.

src/main/java/org/opentripplanner/middleware/models/AbstractUser.java
```diff
@@ -51,4 +54,12 @@ public boolean canBeManagedBy(Auth0UserProfile user) {
         // Fallback to Model#userCanManage.
         return super.canBeManagedBy(user);
     }
+
+    @Override
+    public boolean delete() throws Auth0Exception {
+        // If the user subclass is the only user in existence for the
```
binh-dam-ibigroup: Incomplete sentence.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -32,4 +39,35 @@
...
+                LOG.error("Error deleting user's ({}) monitored trip {}", this.id, trip.id);
+                return false;
+            }
+        }
+        try {
+            super.delete();
+        } catch (Auth0Exception e) {
+            LOG.error("Could not delete Auth0 user. Aborting delete user.", e);
+            return false;
+        }
```
binh-dam-ibigroup: This same block appears in the `AbstractUser` child classes and I'm wondering if it can be moved into `AbstractUser` and return, for instance, a boolean indicating the deletion was successful.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -79,8 +89,8 @@ public static TokenCache getCachedToken() {
     }
 
     /**
-     * Gets an Auth0 API access token for authenticating requests to the Auth0 Management API. This will either create
-     * a new token using the oauth token endpoint or grab a cached token that it has already created (if it has not
+     * Gets an Auth0 API access token for authenticating requests to the Auth0 Management API. This will either create a
+     * new token using the oauth token endpoint or grab a cached token that it has already created (if it has not
```
binh-dam-ibigroup, line 97: I suggest we use Auth0 in place of oauth.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Tests to confirm that an otp user, related monitored trip and plan can be created and deleted leaving no orphaned
+     * records. This also includes Auth0 users if auth is enabled.
+     */
+    @Test
+    public void canSimulateApiUserFlow() {
+        assumeTrue(getBooleanEnvVar("RUN_E2E"));
+
+        // create otp user as api user
+        HttpResponse<String> createUserResponse = mockAuthenticatedPost("api/secure/user",
```
binh-dam-ibigroup, line 110: This is tricky. In order to complete, this test needs to obtain a valid Auth0 `accessToken`.
The fake `TokenHolder` produced by `Auth0UsersTest` does not provide one, and the call to the Auth0 management API fails as a result.

binh-dam-ibigroup (commented):
I added more comments.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -53,12 +54,17 @@ public static void checkUser(Request req) {
         try {
             DecodedJWT jwt = verifier.verify(token);
             Auth0UserProfile profile = new Auth0UserProfile(jwt);
+            if (!isValidUser(profile)) {
```
binh-dam-ibigroup, line 64: This change is breaking the OTP user signup process. Do we need this check for something in particular? After a new OTP User has signed up, an Auth0 profile is available, however no `OtpUser` (or `ApiUser` or `AdminUser`) exist for that user. (The `OtpUser` is created when the user finishes the account creation wizard.)

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -12,6 +15,15 @@ public OtpUserController(String apiPrefix) {
...
+        String apiKey = req.headers("x-api-key");
+        ApiUser apiUser = Persistence.apiUsers.getOneFiltered(Filters.eq("apiKeys.value", apiKey));
+        user.applicationId = apiUser.id;
```
binh-dam-ibigroup: This needs a null check. ApiUser will be null if this `OtpUser` is signed up from MOD UI.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -118,13 +118,6 @@ U preUpdateHook(U user, U preExistingUser, Request req) {
      */
     @Override
     boolean preDeleteHook(U user, Request req) {
-        try {
-            deleteAuth0User(user.auth0UserId);
```
binh-dam-ibigroup, line 122: Also remove unused import.

binh-dam-ibigroup (commented):
See one more comment that is breaking the existing user sign up on MOD UI.

src/test/java/org/opentripplanner/middleware/TestUtils.java
landonreed: That's not a typo, that's the type of token that's being requested.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: Agree on the null check, but we should also change MOD UI to use an API key assigned to an ApiUser. If the ApiUser is null, we should probably halt.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -134,53 +135,76 @@ public T getById(String id) {
...
         return this.mongoCollection;
     }
 
     /**
-     * Get all objects satisfying the supplied Mongo filter.
-     * This ties our persistence directly to Mongo for now but is expedient.
-     * We should really have a bit more abstraction here.
+     * Get all objects satisfying the supplied Mongo filter. This ties our persistence directly to Mongo for now but is
+     * expedient. We should really have a bit more abstraction here.
      */
```
landonreed, line 208: There are a lot of formatting changes just to comments, we should try to avoid this because it clutters up PRs.

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -1,29 +1,55 @@
...
+
+    /**
+     * Mock plan endpoint used by the mock OTP server.
+     */
+    static final String MOCK_OTP_PLAN_ENDPOINT = "otp/plan";
+
+    /**
+     * Password used to create and validate temporary Auth0 users
+     */
+    static final String TEMP_AUTH0_USER_PASSWORD = "t3mp-pa$$w0rd";
```
landonreed: I don't like the idea of storing a password in code like this, even if it's just temporary. Any reason not to use a random UUID?

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Creating a single ApiUser for MOD UI might make sense... however, even with that in place, the problem here is that for a fresh Auth0 id from a new signup from MOD UI, there will be no ApiUser for that Auth0 id.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
landonreed: Hmm, that's a case that I think we would want to handle in a special way. I extracted the `isValidUser` check from `isAuthorized` because in nearly all cases, we want to stop a request in `checkUser` if it's not being made by a valid user in our database.

src/test/java/org/opentripplanner/middleware/TestUtils.java
landonreed: Fixed in https://github.com/ibi-group/otp-middleware/pull/64/commits/6ff1fd1e10f9a19ca588b960984e4d4c68aca61d

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: Fixed in 39bfee48488d070febab2255505565f5e875de37

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
landonreed: So that it can be used elsewhere in the application (we're not re-getting it).

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
landonreed: Fixed in 942c804be5fb46febc330767bd77772414d2221a

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
landonreed: Fixed

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
landonreed: Fixed in 4ae39456a04be9cf45d0571714bcaffede23c88d

binh-dam-ibigroup:
There are new conflicts as a result of recent merging of PRs.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -53,18 +59,48 @@ public static void checkUser(Request req) {
...
+     * Check for POST requests that are creating an {@link AbstractUser} (a proxy for OTP/API users).
+     */
+    private static boolean isCreatingSelf(Request req, Auth0UserProfile profile) {
+        // First, check the request method/path.
+        String uri = req.uri();
+        String method = req.requestMethod();
+        if (method.equalsIgnoreCase("POST") && (uri.endsWith(OTP_USER_PATH) || uri.endsWith(API_USER_PATH))) {
+            // Next, get the user object from the request body, verifying that the Auth0UserId matches.
+            try {
+                AbstractUser user = getPOJOFromRequestBody(req, AbstractUser.class);
```
binh-dam-ibigroup: A piece of code might still be missing here.
For a new signup from MOD UI, an exception is thrown:
```log
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.opentripplanner.middleware.models.AbstractUser` (no Creators, like default construct, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: (String)"{"auth0UserId":"auth0|5f****************",
<req.body() content below>}"; line: 1, column: 1]
```

`req.body()` returns something such as 
```json
{
"auth0UserId":"auth0|5f**************",
"email":"email@example.com",
"hasConsentedToTerms":true,
"notificationChannel":"email",
"phoneNumber":"",
"savedLocations":[],
"storeTripHistory":false
}
```

binh-dam-ibigroup: The easiest workaround is probably to pass `ApiUser.class` or `OtpUser.class` to the get POJO method depending on the path.

README.md
```diff
@@ -79,3 +90,26 @@ From here, click on the organization name and then copy the name from the pop-up
...
+
+One special note is that the default directory must be set at the Account/tenant level. Otherwise, authorization will not
+be successful. See this StackOverflow answer for more info: https://stackoverflow.com/a/43835563/915811
+
+The special E2E client settings should be defined in `env.yml`:
+
+| Parameter | Default | Description |
+| --- | --- | --- |
+| AUTH0_CLIENT_ID | N/A | Special E2E application client ID. |
+| AUTH0_CLIENT_SECRET | N/A | Special E2E application client secret. |
```
binh-dam-ibigroup, line 113: These new variables should appear in `env.yml.tmp`.

binh-dam-ibigroup (changes requested):
User signup is still broken, for a different reason this time. Merge conflicts need to be resolved.

README.md
landonreed: I'm not sure that's the case if they're only intended for e2e.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
landonreed: OK, fixed and tested. https://github.com/ibi-group/otp-middleware/pull/64/commits/869713da7fde4b06d71eb64ac001530ae7f4cc73

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -217,8 +219,8 @@ private T getOne(Request req, Response res) {
     /**
      * HTTP endpoint to delete one entity specified by ID.
      */
-    private String deleteOne(Request req, Response res) {
-        long startTime = DateTimeUtils.currentTimeMillis();
+    private T deleteOne(Request req, Response res) {
+        long startTime = System.currentTimeMillis();
```
evansiroky: Please use DateTimeUtils for getting the current time in case this is tested with mocked time.

landonreed: Oops, sorry. Missed that in the merge.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -22,32 +22,33 @@
...
+    // TODO: It might be useful to allow this to run without DISABLE_AUTH set to true (in an end-to-end environment
+    //  using real tokens from Auth0.
+    private static boolean testsShouldRun = getBooleanEnvVar("RUN_E2E") && authDisabled();
 
     /**
      * Create an {@link ApiUser} and an {@link AdminUser} prior to unit tests
      */
     @BeforeAll
     public static void setUp() throws IOException {
+        assumeTrue(testsShouldRun);
```
evansiroky: This is never returning true because the config has yet to be loaded, causing the authDisabled call to always return false.

evansiroky: Wait nvm. This wasn't working because I was running the tests in just this class isolated from the rest of the test suite. Regardless, a good general principle of testing (I know we don't adhere to this with the datatools e2e test) is that any test should be able to run independent of others and it might be a good idea to refactor things a bit so it works here.

evansiroky (approved):
See comment here: https://github.com/ibi-group/otp-middleware/pull/64#discussion_r482418729. I guess it's good to go after that.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
landonreed: Good catch. Fixed in 69c562fa9223d09d1fde73eac6f92de28d58d3d4

src/main/java/org/opentripplanner/middleware/models/AbstractUser.java
```diff
@@ -51,4 +57,17 @@ public boolean canBeManagedBy(Auth0UserProfile user) {
         // Fallback to Model#userCanManage.
         return super.canBeManagedBy(user);
     }
+
+    @Override
+    public boolean delete() {
+        // FIXME: Check that the Auth0 user ID being deleted does not exist as a different child class?
```
binh-dam-ibigroup, line 63: Yep. Admin users will have to be careful when deleting ApiUsers or OtpUsers.

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -47,6 +48,12 @@ public static boolean getBooleanEnvVar(String var) {
         return variable != null && variable.equals("true");
     }
 
+    /**
+     * Helper method to determine if end to end is enabled and auth is disabled. (Used for checking if tests should run.)
+     */
+    public static boolean isEndToEndAndAuthIsDisabled() {
+        return getBooleanEnvVar("RUN_E2E") && authDisabled();
```
binh-dam-ibigroup, line 55: Philosophical question: do we still need to have the DISABLE_AUTH config parameter in that case?

binh-dam-ibigroup (approved):
Great job @br648  and @landonreed

src/test/java/org/opentripplanner/middleware/TestUtils.java
landonreed: Totally agree with what you're getting at. I think I commented this somewhere else. I'll create an issue: https://github.com/ibi-group/otp-middleware/issues/68

PR #65, opened 2020-08-27
Create key w better name; restrict key deletion

landonreed:
### Checklist

### Description

This PR adds a couple of more changes to API Key management:
- restricts key deletion to admin users -- the thinking here is that we may not want people to constantly churn through API keys. That could cause us/admins to lose track of API usage as they're deleted.
- creates keys with more descriptive names (this way it's easier to track back the key in the AWS console)

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
@@ -8,6 +8,7 @@ import org.opentripplanner.middleware.models.ApiKey;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.persistence.Persistence;
 import org.opentripplanner.middleware.utils.ApiGatewayUtils;
+import org.opentripplanner.middleware.utils.CreateApiKeyException;
 import org.opentripplanner.middleware.utils.HttpUtils;
 import org.opentripplanner.middleware.utils.JsonUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


evansiroky (approved):

br648 (approved):
LGTM.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -54,30 +56,36 @@ private static AmazonApiGateway getAmazonApiGateway() {
...
+    public static ApiKey createApiKey(ApiUser user, String usagePlanId) {
+        if (user == null || user.id == null || usagePlanId == null) {
             LOG.error("All required input parameters must be provided.");
             return null;
         }
         long startTime = System.currentTimeMillis();
         try {
             AmazonApiGateway gateway = getAmazonApiGateway();
-            // create API key
+            // Before creating key, verify usage plan exists.
```
binh-dam-ibigroup: Perhaps add a comment explaining what happens if the usage plan does not exist. (does that trigger an exception?)

binh-dam-ibigroup (approved):
One additional comment clarification could be added when checking the existence of the usage plan ID, otherwise this PR is good to go.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
landonreed: Yes, exactly. I added some changes to clarify the behavior (there's a new exception class as well).

landonreed:
Made some minor tweaks to how exceptions get generated, so I thought I should get some quick feedback before merging.

evansiroky (approved):

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
```diff
@@ -99,25 +100,30 @@ private ApiUser createApiKeyForApiUser(Request req, Response res) {
...
         return Persistence.apiUsers.getById(targetUser.id);
     }
 
     /**
      * Delete an api key from a given user's list of api keys (if present) and from AWS api gateway.
      */
     private ApiUser deleteApiKeyForApiUser(Request req, Response res) {
+        Auth0UserProfile requestingUser = Auth0Connection.getUserFromRequest(req);
+        // Do not permit key deletion unless user is an admin.
+        if (!isUserAdmin(requestingUser)) {
```
binh-dam-ibigroup, line 124: As a result of this new restriction, test `ApiKeyManagementTest#canDeleteApiKeyForSelf()` should be removed.

binh-dam-ibigroup (changes requested):
Sorry, I need to change approval status. One E2E test may be obsolete, is failing as noted, and may need to be removed.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiUserController.java
landonreed: Good catch. How about we change it to cannotDeleteApiKeyForSelf?

binh-dam-ibigroup: That should do too, and we'll have a test to police the new restriction.

binh-dam-ibigroup (approved):
Looks good!

PR #69, opened 2020-09-03
Run e2e

landonreed:
### Checklist

### Description

Adds E2E to Travis. It's a similar approach to how data tools works, but it's a bit less intensive as far as re-constructing the yml file (which data tools does, but this just circumvents getting config variables from that file in favor of using `System.getEnv` if running on Travis CI).


src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -26,13 +26,31 @@
...
+    }
+
     /**
      * Load config files from either program arguments or (if no args specified) from
      * default configuration file locations. Config fields are retrieved with getConfigProperty.
      */
     public static void loadConfig(String[] args) throws IOException {
         FileInputStream envConfigStream;
-
+        if (isRunningCi()) {
```
evansiroky: All these checks could use a comment in the method they are contained in to note the use of environment variables when running on Travis.

evansiroky (approved):
Just a comment about adding more comments.

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -40,19 +41,11 @@
...
-    }
-
     /**
      * Helper method to determine if end to end is enabled and auth is disabled. (Used for checking if tests should run.)
      */
-    public static boolean isEndToEndAndAuthIsDisabled() {
-        return getBooleanEnvVar("RUN_E2E") && authDisabled();
+    public static boolean isEndToEnd() {
+        return getBooleanEnvVar("RUN_E2E");
     }
```
evansiroky: This should probably be a static variable since this'll read from the environment variables over and over.

src/main/java/org/opentripplanner/middleware/models/AbstractUser.java
```diff
@@ -60,6 +61,10 @@ public boolean canBeManagedBy(Auth0UserProfile user) {
 
     @Override
     public boolean delete() {
+        if (Auth0Connection.isAuthDisabled()) {
+            LOG.warn("Skipping delete user call (Auth is disabled).");
```
binh-dam-ibigroup: Change the text to "Skipping call to delete Auth0 user (Auth is disabled)"

I would tweak this check as follows because the if condition only applies to deleting the Auth0 user:
```java
if (Auth0Connection.isAuthDisabled()) {
  ...
} else {
  try {
    deleteAuth0User(...)
  } catch (...) {
    ...
  }
}
```

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -26,13 +26,30 @@
...
+        String variable = System.getenv(var);
+        return variable != null && variable.equals("true");
+    }
+
+    /**
+     * Check if running in Travis CI. A list of default environment variables from Travis is here:
+     * https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
+     */
+    public static boolean isRunningCi() {
+        return getBooleanEnvVar("TRAVIS") && getBooleanEnvVar("CONTINUOUS_INTEGRATION");
```
binh-dam-ibigroup: The result of this should probably be cached in a static variable too.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -53,14 +54,23 @@
     private static ApiUser apiUser;
     private static OtpUser otpUser;
 
+    /**
+     * Whether tests for this class should run. End to End must be enabled and Auth must NOT be disabled. This should be
+     * evaluated after the middleware application starts up (to ensure default disableAuth value has been applied from
+     * config).
+     */
+    private static boolean testsShouldRun() {
+        return isEndToEnd() && !isAuthDisabled();
```
binh-dam-ibigroup: The result of this should probably be cached too.

src/test/java/org/opentripplanner/middleware/OtpMiddlewareTest.java
```diff
@@ -34,9 +35,14 @@ public static void setUp() throws RuntimeException, IOException, InterruptedExce
         OtpMiddlewareMain.inTestEnvironment = true;
         // If in the e2e environment, use the secret env.yml file to start the server.
         // TODO: When ran on Travis CI, this file will automatically be setup.
-        String[] args = getBooleanEnvVar("RUN_E2E")
-            ? new String[]{"configurations/default/env.yml"}
-            : new String[]{"configurations/test/env.yml"};
+        String[] args;
+        if (getBooleanEnvVar("RUN_E2E")) {
```
binh-dam-ibigroup: Use function `isEndToEnd()`

binh-dam-ibigroup (changes requested):
Requesting a couple of minor tweaks, including caching the result of some function calls, and this PR should be good to go.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
landonreed: I'm not really sure why we're concerned with caching these results? So that we save a microsecond on evaluating the expression?

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
landonreed: I really don't think this one is necessary.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -76,7 +76,7 @@ public static void tearDown() {
      */
     @Test
     public void canCreateApiKeyForSelf() {
-        assumeTrue(getBooleanEnvVar("RUN_E2E"));
+        assumeTrue(isEndToEnd);
```
binh-dam-ibigroup: Do we still need this assertions? It is already checked in the @BeforeAll code.

binh-dam-ibigroup (approved):

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
landonreed: good catch.

PR #72, opened 2020-09-10
Add pagination to bugsnag and generic api controller

landonreed:
### Checklist

### Description

Adds pagination to bugsnag (our dev instance currently returns thousands of records) and the generic API controller. otp-admin-ui PR is https://github.com/ibi-group/otp-admin-ui/pull/23 (still need changes in otp-react-redux)


src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -119,7 +120,7 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
                     // Note: there exists a method withResponseAsCollection, but unlike what its name suggests,
                     // it does exactly the same as .withResponseType and does not generate a return type array.
                     // See issue https://github.com/manusant/spark-swagger/issues/12.
-                    .withResponseType(Array.newInstance(clazz, 0).getClass()),
+                    .withResponseType(ResponseList.class),
```
binh-dam-ibigroup, line 133: Need to add the spark-swagger code to document the new query parameters `limit` and `page`.

src/main/java/org/opentripplanner/middleware/controllers/api/BugsnagController.java
```diff
@@ -56,17 +59,22 @@ public void bind(final SparkSwagger restApi) {
     /**
      * Get all Bugsnag events from Mongo and replace the project id with the project name and return
      */
-    private static List<EventSummary> getEventSummary(Request request, Response response) {
+    private static ResponseList<EventSummary> getEventSummary(Request req, Response res) {
         List<EventSummary> eventSummaries = new ArrayList<>();
         List<BugsnagEvent> events = bugsnagEvents.getAll();
+        int limit = Integer.parseInt(req.queryParamOrDefault("limit", "10"));
```
binh-dam-ibigroup: Add the spark-swagger code to document the new query parameters `limit` and `page`. (This is an admin endpoint so this is for folks who develop against this endpoint).

src/main/java/org/opentripplanner/middleware/controllers/response/ResponseList.java
```diff
@@ -0,0 +1,29 @@
...
+     * Constructor for generating a paginated response list.
+     * @param data - list of entities to limit
+     * @param page - from index to abbreviate data
+     * @param limit - number of entities to include in list
+     */
+    public ResponseList(List<T> data, int page, int limit){
+        this.page = page;
+        this.limit = limit;
+        this.total = data.size();
+        this.data = data.subList(page * limit, Math.min(this.total, (page + 1) * limit));
```
binh-dam-ibigroup: It looks like each "page" request involves re-querying everything from Mongo ior Busgsnag, just for us to extract the desired subset extract. Could we just tell Mongo or Bugsnag to return the desired subset directly? That would be way more efficient, and we would just forward the results.

binh-dam-ibigroup (changes requested):
I will need to test some more. For now, the obvious changes are documenting the request parameters, and there is probably a way to request a subset of the data directly from Mongo and/or Bugsnag.

br648 (approved):
Looks good so approving.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -112,7 +113,7 @@ private static void initializeHttpEndpoints() throws IOException, InterruptedExc
          * _encoded_ URL e.g. http://localhost:3000/#/register which allows for greater flexibility.
          */
         spark.get("/register", (request, response) -> {
-            String route = HttpUtils.getRequiredQueryParamFromRequest(request, "route", false);
+            String route = getQueryParamFromRequest(request, "route", false);
```
evansiroky: I think using static imports like this might actually be a bad practice. See [this perspective here](https://stackoverflow.com/a/421127/269834).

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -103,12 +104,35 @@ public static String httpRequest(URI uri, int connectionTimeout, REQUEST_METHOD
         return null;
     }
 
+    public static int getQueryParamFromRequest(Request req, String name, boolean optional, int min, int defaultValue, int max) {
```
evansiroky: Needs JavaDoc

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -103,12 +104,35 @@ public static String httpRequest(URI uri, int connectionTimeout, REQUEST_METHOD
...
+        int value = defaultValue;
+        String requestValue = null;
+        try {
+            // Attempt to get value from query param.
+            requestValue = HttpUtils.getQueryParamFromRequest(req, name, optional);
+            if (requestValue != null) {
+                value = Integer.parseInt(requestValue);
+                // If requested value is out of range, pin to min/max.
+                if (value < min) value = min;
+                else if (max != -1 && value > max) value = max;
```
evansiroky: To specify no limit on the max value, something like `Integer.MAX_VALUE` or `Double.POSITIVE_INFINITY` should be used. Since -1 is a real number, this design could introduce a bug if -1 is actually an acceptable maximum value.

src/main/java/org/opentripplanner/middleware/controllers/response/ResponseList.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param filter - filter (query/sort) to apply to find operation (null is OK)
+     * @param offset - number of results by which to offset
+     * @param limit - number of results by which the response should be limited
+     */
+    public ResponseList(MongoCollection<T> collection, Bson filter, int offset, int limit){
+        FindIterable<T> iterable = filter != null ? collection.find(filter) : collection.find();
+        this.data = iterable.skip(offset).limit(limit).into(new ArrayList<>());
+        this.offset = offset;
+        this.limit = limit;
+        this.total = collection.countDocuments();
```
evansiroky: Wouldn't this count all documents and not just those that the query would filter to? Seems like the `total` should be the total amount of records found in the query.

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -67,12 +67,24 @@ paths:
...
+        description: "If specified, the number of records to skip/offset."
+        required: false
+        type: "string"
+        default: "0"
       responses:
         "200":
           description: "successful operation"
           schema:
-            $ref: "#/definitions/AdminUser[]"
+            $ref: "#/definitions/ResponseList"
```
evansiroky, line 87: I suppose this is a little better than the AdminUser[] type since the definition for that was just an empty object. It might be out of scope for this PR, but it would be nice to have the swagger docs include refs to the proper object in an array.

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -684,8 +737,25 @@ definitions:
...
       id:
         type: "string"
-  AdminUser[]:
+  ResponseList:
     type: "object"
+    properties:
+      data:
+        type: "array"
+        items:
+          type: "string"
```
evansiroky, line 760: This is going to be incorrect since at the very least the items will be an object. It might be necessary to have separate ResponseList types that include the correct reference to the subclass that they return.

evansiroky (changes requested):
See comments

src/main/resources/latest-spark-swagger-output.yaml
landonreed: @binh-dam-ibigroup, do you have any thoughts on how to ammend this swagger def?

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
landonreed: What is your recommendation? That we qualify the method call with the util class? I'm fine with that, but we should define an approach for our projects generally on this. Perhaps we should create an issue and handle that in a separate project-wide PR?

src/main/resources/latest-spark-swagger-output.yaml
binh-dam-ibigroup: The most thorough way to handle this would be to change the document generator class to create a `ResponseList_Subtype` object definition for each of the `ResponseList` responses we are publishing. We can also try creating a separate descendant of `ResponseBody` per @evansiroky's suggestion above and see how spark-swagger handles that.

binh-dam-ibigroup: To follow up on my comment above, you could modify class `PublicApiDocsGenerator` to do the following:
- Scan all `ResponseList` return entries in the autogenerated swagger.
- For each instance, find the type of the elements in the list from one of the parent elements.
- Generate a custom return type definition that correspond to `ResponseList<type>`
- Replace `ResponseList` with the custom type generated above.

src/main/java/org/opentripplanner/middleware/controllers/api/BugsnagController.java
```diff
@@ -54,19 +60,24 @@ public void bind(final SparkSwagger restApi) {
     }
 
     /**
-     * Get all Bugsnag events from Mongo and replace the project id with the project name and return
+     * Get all Bugsnag events from Mongo and replace the project id with the project name and return.
```
binh-dam-ibigroup: Comment is outdated. Are we still fetching all events? The description for the GET method should be updated too.

src/main/java/org/opentripplanner/middleware/controllers/api/ApiController.java
```diff
@@ -112,19 +124,19 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
 
         baseEndpoint
             // Get multiple entities.
-            .get(path(ROOT_ROUTE)
+            .get(
+                path(ROOT_ROUTE)
                     .withDescription("Gets a list of all '" + className + "' entities.")
```
binh-dam-ibigroup: This description should mention pagination.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -60,72 +61,56 @@ public void bind(final SparkSwagger restApi) {
                     .withName("userId")
                     .withRequired(true)
                     .withDescription("The OTP user for which to retrieve trip requests.").and()
-                .withQueryParam()
-                    .withName(LIMIT_PARAM_NAME)
-                    .withDefaultValue(String.valueOf(DEFAULT_LIMIT))
-                    .withDescription("If specified, the maximum number of trip requests to return, starting from the most recent.").and()
-                .withQueryParam()
-                    .withName(FROM_DATE_PARAM_NAME)
+               .withQueryParam(LIMIT)
```
binh-dam-ibigroup, line 64: The description of the GET method should be updated 
(remove 'all' and mention pagination).

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -270,7 +271,7 @@ public void clearRealtimeInfo() {
     /**
      * Get monitored trips for the specified {@link OtpUser} user Id.
      */
-    public static List<MonitoredTrip> tripsForUser(String userId) {
```
binh-dam-ibigroup, line 273: `List` is an unused import.

src/main/java/org/opentripplanner/middleware/persistence/TypedPersistence.java
```diff
@@ -6,10 +6,12 @@
 import com.mongodb.client.model.Filters;
 import com.mongodb.client.model.FindOneAndUpdateOptions;
 import com.mongodb.client.model.ReturnDocument;
+import com.mongodb.client.model.Sorts;
```
binh-dam-ibigroup: This import shows as unused.

src/main/resources/latest-spark-swagger-output.yaml
binh-dam-ibigroup: @evansiroky Note that `AdminUser` is a subclass of `AbstractUser`, but spark-swagger ignores inherited fields at the time this was implemented. We insert inherited fields into the `AdminUser`swagger type using the `PublicApiDocGenerator` class.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
```diff
@@ -60,72 +61,56 @@ public void bind(final SparkSwagger restApi) {
                     .withName("userId")
                     .withRequired(true)
                     .withDescription("The OTP user for which to retrieve trip requests.").and()
-                .withQueryParam()
-                    .withName(LIMIT_PARAM_NAME)
-                    .withDefaultValue(String.valueOf(DEFAULT_LIMIT))
-                    .withDescription("If specified, the maximum number of trip requests to return, starting from the most recent.").and()
```
binh-dam-ibigroup, line 66: Is sorting from most recents still happening?

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -154,7 +157,10 @@ public void canSimulateApiUserFlow() throws IOException {
             HttpUtils.REQUEST_METHOD.GET
         );
         assertEquals(HttpStatus.OK_200, tripRequestResponse.statusCode());
-        List<TripRequest> tripRequests = JsonUtils.getPOJOFromJSONAsList(tripRequestResponse.body(), TripRequest.class);
```
binh-dam-ibigroup, line 160: `List` is an unused import.

binh-dam-ibigroup (changes requested):
Some light documentation updates, then it should be good to go.

src/main/java/org/opentripplanner/middleware/controllers/api/TripHistoryController.java
landonreed: Yep, updated endpoint description.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
evansiroky: Let's discuss with the team together sometime and then if we decide we want to do this, then we can handle all the changes in another PR.

evansiroky (changes requested):
- Build is failing
- PR has merge conflicts
- Figuring out documentation of ResponseList typing seems to be unresolved

binh-dam-ibigroup:
Does a dependency need to be updated for the build to pass?

src/main/resources/latest-spark-swagger-output.yaml
landonreed: Further discussion should be had in https://github.com/ibi-group/otp-middleware/issues/85

Issue #70, opened 2020-09-03
Trip Monitoring TODO list


landonreed:
Pending back end items for trip monitoring (note: UI work on hold until progress is made here):
- [x] save monitored trip flow (validate that itinerary exists before persisting monitored trip) - @landonreed => @binh-dam-ibigroup => #77 
- [x] prevent multiple, consecutive "itinerary not found" notifications from being sent to user - @binh-dam-ibigroup => See comments below.
- [x] (issue with decoding the OTP response, too many files open error) - @binh-dam-ibigroup => #74 
- [ ] http endpoint for monitoring trip (in addition to current sms/email) (consider navigation when designing) = ?
- [x] the check for if a trip is possible/exists during monitoring is very primitive right now... definitely needs some work (determine best approach possibly checking GTFS for trips, stops) - ?? => #357, #364
- [x] Pause trip monitoring if some threshold of "trip not found" is met, set status enum (TRIP_NO_LONGER_POSSIBLE, etc.) on monitored trip (that we can show in the UI) - ?? => #326 
- [x] change in monitored trip monitoring window settings - @evansiroky => #75
- [x] Recalculate for current day (plan trip at the appropriate time) - @evansiroky => #75

- ~The phone number validation flow looks a bit awkward as it requires extra /api/secure/user PUT operations in order to get the phone number validated. See https://github.com/ibi-group/otp-middleware/pull/33/files#r483187826. (current phone validation PR https://github.com/opentripplanner/otp-react-redux/pull/224 implements a workaround.)~ => Moved to #79.

binh-dam-ibigroup:
I found the root cause for most of of the "Itinerary not found" notifications I was receiving. Basically, the stop id changed between the time I set the notification and when the checks occurred, causing a mismatch in the itinerary comparisons. The stop id changed because of a GTFS change that modified an agency id (agency ids are used in OTP's global stop ids). Recreating a monitored trip seems to have resolved the issue.

binh-dam-ibigroup:
Another cause for an "Itinerary not found" notification is that the UI is saving real-time itineraries (which is incorrect) and trying to compare such itineraries with the updated itineraries from OTP that also include real-time updates. Comparing itineraries with real-time updates is unlikely to succeed IMO.

PR #75, opened 2020-09-16
Trip monitor timing refactor

evansiroky:
### Checklist

### Description

This PR refactors the timing aspects of the trip monitor. The previous implementation was a little naive and had trouble calculating the proper time to check since it didn't use information from previous checks. This PR changes that and now will store the next upcoming itinerary for a trip and will use that check as a baseline for determining if a trip check should be skipped and how long to wait before the lead time. This results in making requests on the proper date and tracking a trip all the way until the end even if it goes longer than scheduled.

So, this PR addresses both the `change in monitored trip monitoring window settings` and `Recalculate for current day (plan trip at the appropriate time)` items in #70.

Here is a summary of changes:

- Clarifies the purpose of a few variables by renaming them
  - `YYYY_MM_DD` -> `DEFAULT_DATE_FORMAT_PATTERN`
  - `org.opentripplanner.middleware.otp.response.Response` -> `org.opentripplanner.middleware.otp.response.OtpResponse`
- Using mocked date/time for items that create Date objects
- JourneyState
  - now saves just the matchingItinerary instead of an otpResponse
  - also saves the target date for easier OTP querying
  - renames some variables to have `...millis` to make it clear that the values are milliseconds
- MonitoredTrip
  - Constructor automatically sets the tripTime from the query params
  - Adds an `isInActive` method for checking if the trip is never active
  - Adds method `getTimezoneForTargetLocation` and associated helper methods to get the appropriate timezone to calculate a target time with based off of whether the trip is departing at or arriving by
  - Adds convenience methods `getHour` and `getMinute`.
- Itinerary/Leg
  - replaces equality/hashcode check of `startTime` and `endTime` with `duration` instead. This is probably a temporary fix and more nuance will be needed in the future especially when dealing with realtime itineraries
- CheckMonitoredTrip
  - Prefixes all log outputs in the class with the trip ID. (this required a change to the logback.xml file)
  - Refactors a lot of methods to not be static and instead be instance methods to avoid repetitive arguments/variables
  - Refactored timing checks to rely heavily on refactored implementation of `shouldSkipMonitoredTripCheck`. This method now calculates itineraries if none have been calculated before in order to have a baseline to use when determining if a formal check of the trip is warranted that might result in notifications being sent out.
  - Introduces new method `calculateNextItinerary` which advances the trip to the appropriate date and sends off a request to OTP to obtain a new baseline itinerary to compare with in future checks of the trip.
- TestUtils
  - Adds to the mock OTP server by allowing a series of mock responses to be added for individual test cases
- CheckMonitoredTripTest
  - significantly modifies the testSkipMonitoredTripCheck ParameterizedTest to check a variety of test cases that test various aspects of the shouldSkipMonitoredTripCheck. These tests sometimes will result in requests to the mock OTP server and as such mock responses are created. These tests hit a lot of the logic around generating OTP queries and therefore, the draft test of `canGenerateTripPlanQueryParams` was removed.


br648:
@evansiroky the unit test "canMonitorTrip()" fails with the following exception:

java.io.FileNotFoundException: src\test\resources\org\opentripplanner\middleware\planResponse.json (The system cannot find the file specified)

This file is located in src\test\resources\org\opentripplanner\middleware\persistence.

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
```diff
@@ -28,56 +28,52 @@ public JourneyState(MonitoredTrip monitoredTrip) {
         this.userId = monitoredTrip.userId;
     }
 
-    /**
-     * The {@link MonitoredTrip} id that this journey state is tracking.
-     */
-    public String monitoredTripId;
+    public int lastArrivalDelay;
```
br648, line 34: Javadoc for all new variables.

br648 (approved):
On the face of it this looks really good. I don't feel I have enough experience to comment on the finer workings so will leave that to others. Perhaps a walk through would be useful, at least for me?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -163,6 +180,15 @@ public MonitoredTrip updateWeekdays(boolean value) {
         return this;
     }
 
+    /**
+     * Returns true if the trip is not active overall or if all days of the week are set to false
+     */
+    public boolean isInActive() {
```
binh-dam-ibigroup: Typo, should be `isInactive`.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +302,68 @@ public boolean delete() {
         // TODO: Add journey state deletion.
         return Persistence.monitoredTrips.removeById(this.id);
     }
+
+    public List<NameValuePair> getParsedParams() throws URISyntaxException {
```
binh-dam-ibigroup: I think this should return a `Map`, and it would simplify method `isArriveBy` below and the class constructor.

binh-dam-ibigroup: Should the be a `@BsonIgnore` annotation?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -44,121 +48,124 @@
...
-     * Find an itinerary from the OTP response that matches the monitored trip's stored itinerary.
+     * Find and set the matching itinerary from the OTP response that matches the monitored trip's stored itinerary if a
+     * match exists.
+     *
+     * FIXME: the itinerary might actually still be possible, but for some reason the OTP plan didn't find the same
+     *          match. Some additional checks should be performed to make sure the itinerary really isn't possible by
+     *          verifying that the same transit schedule/routes exist and that the street network is the same
      */
-    private static int findMatchingItineraryIndex(MonitoredTrip trip, Response otpResponse) {
+    private boolean findMatchingItinerary() {
```
binh-dam-ibigroup: I would rename to `updateMatchingItinerary` (this imperative method changes class fields, and one would expect a `find` method to return the desired object instead of a boolean).

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -44,121 +48,124 @@
...
     public void run() {
-        // Check if the trip check should be skipped (based on time, day of week, etc.)
-        checkSkipped = shouldSkipMonitoredTripCheck(trip);
-        if (checkSkipped) {
-            LOG.debug("Skipping check for trip: {}", trip.id);
-            return;
+        // Add a prefix of the current trip ID for logging purposes to every log message generated from within an
+        // instance of this class. This assumes that the logback.xml file is properly configured to print this variable.
+        // See http://logback.qos.ch/manual/mdc.html for more info.
+        MDC.put("prefix", String.format("[Trip ID: %s]", trip.id));
```
binh-dam-ibigroup, line 90: This is clever!

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,201 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
-        long minutesUntilTrip = Duration.between(now, tripTime).toMinutes();
-        LOG.info("Trip {} starts in {} minutes", trip.id, minutesUntilTrip);
-        // TODO: Refine these frequency intervals for monitor checks.
+        LOG.info("{} minutes since last checking trip", minutesSinceLastCheck);
+        long minutesUntilTrip = (tripStartInstant.getEpochSecond() - now.toEpochSecond()) / 60;
+        LOG.info("Trip starts in {} minutes", minutesUntilTrip);
+        // skip check if the time until the next trip starts is longer than the requested lead time
+        if (minutesUntilTrip > trip.leadTimeInMinutes) {
+            LOG.info(
+                "Next trip start time is greater than lead time of {} minute(s). Skipping trip.",
```
binh-dam-ibigroup: Tweak log text... maybe "The time until this trip begins again is more than the {}-minute lead time."?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,201 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
-     *
-     * TODO: finish implementation
      */
-    public static String generateTripPlanQueryParams(MonitoredTrip trip) throws URISyntaxException {
+    public String generateTripPlanQueryParams() throws URISyntaxException {
         // parse query params
         List<NameValuePair> params = URLEncodedUtils.parse(
             new URI(String.format("http://example.com/%s", trip.queryParams)),
             UTF_8
         );
```
binh-dam-ibigroup: Use method `MonitoredTrip#getParsedParams`.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,201 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
 
-        // begin building a new list and along the way and figure out if this trip is a depart at or arrive by trip
+        // building a new list by copying all values, except for the date which is set to the target date
         List<NameValuePair> newParams = new ArrayList<>();
```
binh-dam-ibigroup, line 470: Echoing my comment for `MonitoredTrip#getParsedParams`, a map might be useful here too.

binh-dam-ibigroup (commented):
I will need to play some more with it, but see my comments so far.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -132,10 +140,19 @@
...
+    public MonitoredTrip(OtpDispatcherResponse otpDispatcherResponse) throws URISyntaxException {
         queryParams = otpDispatcherResponse.requestUri.getQuery();
         TripPlan plan = otpDispatcherResponse.getResponse().plan;
         itinerary = plan.itineraries.get(0);
+
+        // extract trip time from parsed params
+        List<NameValuePair> parsedParams = getParsedParams();
+        for (NameValuePair parsedParam : parsedParams) {
+            if (parsedParam.getName().equals("time")) {
+                tripTime = parsedParam.getValue();
```
binh-dam-ibigroup: At least of these things should also take place with this change:
- Update all `MonitoredTrip` records in Mongo with a value for `tripTime` (and on UI side, make sure that field gets populated too),
- If reading a `MonitoredTrip` record from Mongo with no `tripTime` value set, initialize `tripTime` on first use of `getHour()` or `getMinute()`, which would be more fool-proof.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
binh-dam-ibigroup: I have been trying to see if the stuff plays well with the UI. This is what I understand is happening:
- ~`@BsonIgnore` seems to prevent reflection on the field/method on serialization, but~ reflection on the underlying types is still attempted on deserialization.
- ~`@JsonIgnore` seems to prevent reflection both on serialization and on deserialization. This is what we should use.~

binh-dam-ibigroup: Never mind that didn't seem to work. So this method should probably be renamed, e.g. `parseParams()`

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +302,68 @@ public boolean delete() {
...
+        return false;
+    }
+
+    /**
+     * Gets the timezone of the target location. When planning a trip, we either want to depart at a certain time or 
+     * arrive by a certain time. When departing, we use the local time present at the origin and when arriving by we 
+     * use the local time at the destination. Therefore, this method will return the timezone at the destination if this
+     * trip is an arriveBy trip, or the timezone at the origin if the trip is a depart at trip.
+     */
+    @BsonIgnore
```
binh-dam-ibigroup: Same here. `@BsonIgnore` and `JsonIgnore` are... ignored by the Mongo POJO codec on serialization. The easiest way is to rename the method (`findTimezoneForTargetLocation`?). If a `get` prefix is still needed, we can use the `transient` keyword ([we've faced a similar issue here](https://github.com/ibi-group/otp-middleware/commit/a5baa3c85c6741aac58473e2110ab206221aa993))

binh-dam-ibigroup (changes requested):
- Some `MonitoredTrip` methods need to be modified to play well with serialization.
- `MonitoredTrip#tripTime` may need to be initialized on first use, or the databases need to be updated, let me know what approach you'd prefer.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I disagree because this variable is sent to the `URLEncodedUtils.format` which needs some kind of Iterable.

evansiroky:
Should be ready for review again.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
binh-dam-ibigroup: This is fine for this PR. I ended up needing to do the map in #77 and using it a dozen times.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -132,10 +144,17 @@
...
-    public MonitoredTrip(OtpDispatcherResponse otpDispatcherResponse) {
+    public MonitoredTrip(OtpDispatcherResponse otpDispatcherResponse) throws URISyntaxException {
         queryParams = otpDispatcherResponse.requestUri.getQuery();
         TripPlan plan = otpDispatcherResponse.getResponse().plan;
         itinerary = plan.itineraries.get(0);
+
+        // extract trip time from parsed params
+        Map<String, String> params = parseQueryParams();
+        tripTime = params.get("time");
+        if (tripTime == null) {
```
binh-dam-ibigroup: To follow up on https://github.com/ibi-group/otp-middleware/pull/75#discussion_r489636219,
the UI does not currently initialize `tripTime`, so `tripTime` also needs to be initialized when polling `tripTimeHour` or `tripTimeMinute`. (Those methods are called by `CheckMonitoredTrip#shouldSkipMonitoredTripCheck`).

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +304,61 @@ public boolean delete() {
...
+        } 
+        return fromZoneId.get();
+    }
+
+    /**
+     * Returns the target hour of the day that the trip is either departing at or arriving by
+     */
+    @BsonIgnore
+    public int tripTimeHour() {
+        return Integer.valueOf(tripTime.split(":")[0]);
```
binh-dam-ibigroup, line 336: There should still be a null check (at least) here and in `tripTimeMinutes`.

binh-dam-ibigroup (changes requested):
Sorry, there is one gap regarding the initialization of `tripTime` I would like to get resolved.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Ah, ok I see what's happening. I added some logic to the MonitoredTripController's preCreateHook and preUpdateHook to enforce the existence of the trip time before it gets saved to the database.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: The tripTime should be guaranteed to exist due to https://github.com/ibi-group/otp-middleware/pull/75#discussion_r493967601.

binh-dam-ibigroup (approved):
I think this is ready to go now. Thanks for the changes!

evansiroky:
Assigning to myself to fix merge conflicts

src/main/java/org/opentripplanner/middleware/controllers/api/LogController.java
```diff
@@ -24,7 +24,7 @@
 import static com.beerboy.ss.descriptor.EndpointDescriptor.endpointPath;
 import static com.beerboy.ss.descriptor.MethodDescriptor.path;
 import static org.opentripplanner.middleware.auth.Auth0Connection.isUserAdmin;
-import static org.opentripplanner.middleware.utils.DateTimeUtils.YYYY_MM_DD;
+import static org.opentripplanner.middleware.utils.DateTimeUtils.DEFAULT_DATE_FORMAT_PATTERN;
```
landonreed, line 27: In my view, this variable just became harder to read with this name change.

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
```diff
@@ -29,55 +29,63 @@ public JourneyState(MonitoredTrip monitoredTrip) {
...
 
-    public Set<TripMonitorNotification> lastNotifications = new HashSet<>();
-
-    public long lastNotificationTime;
+    /**
+     * The current or upcoming matching itinerary from plan requests made over the course of monitoring a trip.
+     */
+    public Itinerary matchingItinerary;
 
-    public Set<LocalizedAlert> lastSeenAlerts = new HashSet<>();
```
landonreed, line 61: Why were lastSeenAlerts removed?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -132,19 +144,29 @@
...
         TripPlan plan = otpDispatcherResponse.getResponse().plan;
         itinerary = plan.itineraries.get(0);
-        initializeFromItinerary();
+
+        // extract trip time from parsed params and itinerary
+        initializeFromItineraryAndQueryParams();
     }
 
-    public void initializeFromItinerary() {
+    public void initializeFromItineraryAndQueryParams() throws IllegalArgumentException, URISyntaxException {
```
landonreed, line 157: Needs Javadoc.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +307,61 @@ public boolean delete() {
...
+        return URLEncodedUtils.parse(
+            new URI(String.format("http://example.com/plan?%s", queryParams)),
+            UTF_8
+        ).stream().collect(Collectors.toMap(NameValuePair::getName, NameValuePair::getValue));
+    }
+
+    public boolean isArriveBy() throws URISyntaxException {
+        // if arriveBy is not included in query params, OTP will default to false, so initialize to false
+        return parseQueryParams().getOrDefault("arriveBy", "false").equals("true");
+    }
```
landonreed, line 330: I find the code in parseQueryParams and isArriveBy to be a little hard to read and overly clever. I would prefer the use of one or two variables to improve legibility. Also, these could use some basic javadoc.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +307,61 @@ public boolean delete() {
...
+
+    /**
+     * Gets the timezone of the target location. When planning a trip, we either want to depart at a certain time or 
+     * arrive by a certain time. When departing, we use the local time present at the origin and when arriving by we 
+     * use the local time at the destination. Therefore, this method will return the timezone at the destination if this
+     * trip is an arriveBy trip, or the timezone at the origin if the trip is a depart at trip.
```
landonreed: As long as this is how OpenTripPlanner handles this, I'm OK with this approach. Can you confirm with a code ref? (Might be good to include in this javadoc as well.) Method body could use some inline comments as well.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +307,61 @@ public boolean delete() {
...
+            );
+            throw new RuntimeException(message);
+        } 
+        return fromZoneId.get();
+    }
+
+    /**
+     * Returns the target hour of the day that the trip is either departing at or arriving by
+     */
+    @BsonIgnore
```
landonreed: Is BsonIgnore needed?

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +307,61 @@ public boolean delete() {
...
+    public int tripTimeHour() {
+        return Integer.valueOf(tripTime.split(":")[0]);
+    }
+
+    /**
+     * Returns the target minute of the hour that the trip is either departing at or arriving by
+     */
+    @BsonIgnore
```
landonreed: Is BsonIgnore needed?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+     * Determine whether to skip checking the monitored trip at this instant. The decision on whether to skip the check
+     * takes into account the current time, the lead time prior to the itinerary start and the last time that the trip
+     * was checked. Skipping the check should only occur if the previous trip has ended and the next trip meets the
+     * following criteria for skipping a check:
+     *
+     * - the current time is before when the lead time before the next itinerary starts
+     * - the current time is after the lead time before the next itinerary starts, but is over an hour until the
+     *     itinerary start time and the trip has already been checked within the last 60 minutes
+     * - the current time is after the lead time before the next itinerary starts and between 60-15 minutes prior to the
+     *     itinerary start time, but a check has occurred within the last 15 minutes
```
landonreed, line 276: This feels a bit wordy. The language could be simplified to improve how it reads (without losing clarity). Also, I don't think this accounts for the every minute check within 30 minutes that has been added.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+        // check if the resulting itinerary has a matching itinerary
+        if (!updateMatchingItinerary()) {
+            // TODO: figure out how to notify the user that the itinerary is not possible anymore
+            // for now, just say the next itinerary couldn't be found
+            return false;
+        }
+        return true;
+    }
+
+    private boolean calculateOtpResponse() {
```
landonreed, line 440: Javadoc needed. Also, I feel like the heavy reliance on boolean return values and setting class variables inside of methods kind of ties us into a design pattern that is more difficult to refactor down the road. Part of the reason that I initially made some of these methods static was so that they could be moved into util files or elsewhere at some point. For example, this method doesn't really do anything except send on otp plan request with the query params, but its name "calculateOtpResponse" and lack of parameters suggest that it might be more complicated than it really is.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
     /**
-     * Generate the appropriate OTP query params for the trip for the current check. This currently involves replacing
+     * Calculate the next possible time that the itinerary is possible, advancing the targetZonedDateTime and setting
+     * the targetZonedDateTime, targetDate, otpResponse and matchingItinerary. Returns false if an error occurs or a
+     * matching itinerary isn't found on the next possible date the trip should be monitored on.
+     */
+    private boolean calculateNextItinerary() {
+        // if the trip is not active on the current zoned date time, advance until a day is found when the trip is
+        // active. It is guaranteed that the trip will be active on a certain date because a call to
+        // {@link MonitoredTrip#isInactive} should already have been made in the shouldSkipMonitoredTripCheck method.
```
landonreed: This guarantee is valid only if this method is only called in the correct location (otherwise there is no guarantee and this could result in an endless loop). Instead of making comments like this, I think a better approach is just to make a call to isInactive in this method body just to make sure.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+        while (!trip.isActiveOnDate(targetZonedDateTime)) {
+            targetZonedDateTime = targetZonedDateTime.plusDays(1);
+        }
+        targetDate = targetZonedDateTime.format(DATE_FORMATTER);
```
landonreed, line 415: Would it be more appropriate to just make targetDate a `LocalDate` type instead of a `String`? It seems like there are a lot of places where we're converting to String, but it might be simpler to read/fewer lines to just handle conversion to String in `generateTripPlanQueryParams` (which I think is the only spot it's needed as a string).

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+            if (journeyState.targetDate != null) {
+                LocalDate lastDate = DateTimeUtils.getDateFromString(
+                    journeyState.targetDate,
+                    DEFAULT_DATE_FORMAT_PATTERN
+                );
+                if (
+                    lastDate.getYear() == targetZonedDateTime.getYear() &&
+                        lastDate.getMonthValue() == targetZonedDateTime.getMonthValue() &&
+                        lastDate.getDayOfMonth() == targetZonedDateTime.getDayOfMonth()
+                ) {
```
landonreed, line 321: Is this check the same as `lastDate.getYear() == targetZonedDateTime.getYear() && lastDate.getDayOfYear() == targetZonedDateTime.getDayOfYear()`? Not that much easier to read I suppose. It might be good to just have a DateTimeUtil to compare the date for a LocalDate and ZonedDateTime.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+            // calculate target time for the next trip plan request
+            // find the next possible day the trip is active by initializing the the appropriate target time. Start by
+            // checking today's date at the earliest in case the user has paused trip monitoring for a while
+            targetZonedDateTime = DateTimeUtils.nowAsZonedDateTime(targetZoneId)
+                .withHour(trip.tripTimeHour())
+                .withMinute(trip.tripTimeMinute())
+                .withSecond(0);
+
+            // if a previous journeyState target date exists, check if the previous target date was today's date. If so,
+            // immediately advance to the next day
```
landonreed: This comment kind of just describes what's happening without any context. A better one might be: "If journeyState indicates that today has already been checked, advance targetDate to tomorrow to find matching itinerary for the next series of checks.".

Also, what does it mean if targetDate is null?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+            // calculate the next possible itinerary. If the calculation failed, skip this trip check.
+            if (!calculateNextItinerary()) return true;
+
+            // check if the matching itinerary has already ended
+            if (matchingItinerary.endTime.before(DateTimeUtils.nowAsDate())) {
+                // itinerary is done today, advance to the next day and recheck
+                targetZonedDateTime = targetZonedDateTime.plusDays(1);
+                // calculate the next possible itinerary. If the calculation failed, skip this trip check.
+                if (!calculateNextItinerary()) return true;
+            }
```
landonreed, line 337: I'm trying to understand the net effect of these multiple calls to `calculateNextItinerary`. Let's say it's Thursday and the trip has ended. On line 319, we advance the target date to Friday and call `calculateNextItinerary`. This is where it starts to break down for me... having both of these methods modify `targetZonedDateTime` makes it really hard to follow the logic. We need some kind of real-world example included in the comments. This feels a bit too spaghetti-like for my tastes and seems to be introducing a lot of embedded complexity.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
-            if (minutesUntilTrip <= trip.leadTimeInMinutes) {
+            // If the trip starts within 30 minutes, check the trip every minute (assuming the loop runs every minute).
+            if (minutesUntilTrip <= 30) {
                 // TODO: Change log level.
-                LOG.info(
-                    "Trip {} happening within user's lead time ({} minutes). Checking every minute.",
-                    trip.id,
-                    trip.leadTimeInMinutes
-                );
+                LOG.info("Trip happening within 30 minutes. Checking every minute.");
```
landonreed: This should use a variable that is also used in the conditional check (so that they don't become out of sync).

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -257,133 +263,199 @@ private void enqueueNotification(TripMonitorNotification ...tripMonitorNotificat
...
+            );
+            return true;
+        }
         // If time until trip is greater than 60 minutes, we only need to check once every hour.
         if (minutesUntilTrip > 60) {
             // It's been about an hour since the last check. Do not skip.
             if (minutesSinceLastCheck >= 60) {
                 // TODO: Change log level.
-                LOG.info("Trip {} not checked in at least an hour. Checking.", trip.id);
+                LOG.info("Trip not checked in at least an hour. Checking.");
```
landonreed: This log statement should use a variable that is also used in the conditional.

src/main/resources/logback.xml
```diff
@@ -2,7 +2,7 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
-            <pattern>%highlight(%-5level) %d{yyyy-MM-dd'T'HH:mm:ss.SSS} %yellow([%thread]) %blue(%logger{36}\(%file:%line\)) %msg%n</pattern>
+            <pattern>%highlight(%-5level) %d{yyyy-MM-dd'T'HH:mm:ss.SSS} %yellow([%thread]) %blue(%logger{36}\(%file:%line\)) %yellow(%X{prefix}) %msg%n</pattern>
```
landonreed, line 6: I think we should add a comment about the prefix/MDC thing that notes its usage in CheckMonitoredTrip.java

src/main/java/org/opentripplanner/middleware/otp/response/OtpResponse.java
```diff
@@ -11,7 +11,7 @@
  * Pare down version of class original produced for OpenTripPlanner.
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class Response {
+public class OtpResponse {
```
landonreed, line 14: I'm pretty sure @br648 named this based on the version of the file from OpenTripPlanner. So, the javadoc should at least have a qualified class name and/or URL to the file in the OTP codebase if we're changing its name.

landonreed (changes requested):
This is looking good, but I do have some concerns about how some of the CheckMonitoredTrips methods have been structured and how they update class variables (could use some touch ups to comments as well).

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
evansiroky: This variable wasn't used anymore in the currently implemented logic for checking alerts, so I removed it. See https://github.com/ibi-group/otp-middleware/blob/c9499ccea48990678ed14e52b0761cff80aba104/src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java#L166-L185

src/main/java/org/opentripplanner/middleware/controllers/api/LogController.java
evansiroky: I disagree and believe it has resulted in a more explicit meaning with intended usage conveyed through its name.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Le sigh, OTP does not handle it this way... See https://github.com/ibi-group/OpenTripPlanner/blob/8ec56c10702413f19bcea32e40d1ede0a18ee3f0/src/main/java/org/opentripplanner/routing/graph/Graph.java#L955-L984. The Itinerary responses don't return a timezone identifier, so I think we're going to have to set one in the config.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: The `targetDate` is also persisted via the JourneyState. I'm not sure I'm seeing what you're seeing as this appears to be the only place that there is a conversion from a ZonedDateTime to a string in this file.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I suppose that saves one line of code. This is the only place this kind of comparison occurs, so I don't think it's worth moving into a util function yet. Lmk how strongly you feel about this.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I guess I see your point when getting the date from a string. Another thing I'm concerned about is how a LocalDate might get serialized and deserialized in a database. It's sort of less worry to control all of that interaction by keeping it in a string.

src/main/java/org/opentripplanner/middleware/otp/response/OtpResponse.java
evansiroky: Added in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/resources/logback.xml
evansiroky: Added in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Added in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Added in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: That Thursday example assumes that the `targetDate` variable is not null in the JourneyState which would not be the case the very first time that the CheckMonitoredTrip runs for a certain MonitoredTrip. That's why the check is needed to ensure that after calculating the first target date and matching itinerary that the target date is advanced to the next date that the trip is possible. I added some comments about this in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Revised comments in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Added double-checking code in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I feel like this method is useful to have here since it relies on a number of instance variables. I added JavaDoc in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I updated this a little bit in 6fd00a976c063c5b46e7254a995fbbb628b90445, but prefer to be hyper-specific when talking about timing matters.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Removed in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Removed in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: I also agree that the parseQueryParams is clever, but don't think it's too clever. I think breaking it up wouldn't really add much value. I went ahead and added JavaDoc though in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Added in 6fd00a976c063c5b46e7254a995fbbb628b90445

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
evansiroky: Refactored in 02379c18665691badae829898367975031e3ce37

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
landonreed: Fair enough. Thanks for considering.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
landonreed: Feeling pretty lukewarm about this one.

landonreed (approved):
Thanks for the changes/consideration.

PR #76, opened 2020-09-16
Dynamic redirect

br648:
### Checklist

### Description (Copied from the issue, thanks Landon!)

During sign up for new OTP users, Auth0 provides a way to redirect users to the OTP UI, but because the site is built as a single page app (SPA) directing to the correct path doesn't quite work properly. Fortunately, this is well-documented and Auth0 has a fix that includes adding a server-side endpoint to handle the redirect: https://auth0.com/docs/auth0-email-services/customize-email-templates#dynamic-redirect-to-urls. 

Landon also mentions this:

We should add an endpoint for this to improve the sign up. It should probably account for signups for both OTP users (otp-react-redux) and API users (otp-admin-ui).

I'm not sure if the end point should exist under the API and OTP controllers since coming from Auth0 no creds would be provided. For now I've created an unrestricted end point in OtpMiddlewareMain.java. I can ammend if this is not correct.

No additional configuration parameters are currently required for this update.

I should add that I have not tested this from a UI point of view.


evansiroky (approved):

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -129,11 +129,16 @@ public static int getConfigPropertyAsInt(String name, int defaultValue) {
      * (modified from https://stackoverflow.com/questions/14189162/get-name-of-running-jar-or-exe#19045510).
      * TODO: Extract git properties from JAR, see
      * https://github.com/ibi-group/datatools-server/blob/9f74b821cf351efcdaf7c9c93a3ae8b694d3c3b1/src/main/java/com/conveyal/datatools/manager/DataManager.java#L181-L212.
+     *
+     * "/" is used instead of File.separator, see
+     * https://stackoverflow.com/questions/24749007/how-to-use-file-separator-for-a-jar-file-resource/24749976#24749976
```
binh-dam-ibigroup, line 134: Thanks for catching that.

binh-dam-ibigroup (approved):

PR #77, opened 2020-09-17
Check trip existence

binh-dam-ibigroup:
### Checklist

### Description

This PR causes the `MonitoredTripController` to verify that a same-day, non-realtime itinerary exists for all monitored days for a new or existing `MonitoredTrip`. The verification is done using a 7-day window that starts from the date of the saved OTP query. Same-day means from 3am local-time to 3am the following day for the OTP request date. If there is no same-day itinerary, a 400 error is returned. Upon check, `MonitoredTripController` replaces the itinerary provided in the request body with a non-realtime equivalent.

A new `ItineraryUtils` and `ItineraryExistenceChecker` classes (and accompanying tests) break down the core logic for the itinerary checks.

### Things to finalize:

- name for the `itinerarycheck` endpoint,
- whether the `itinerarycheck` endpoint should return all days at once or should be broken out to check individual days (I have not noticed lags in the combined response, so it has not been a primary concern for now).


src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -279,5 +283,23 @@ public boolean delete() {
...
+                "Could not find coordinate's (lat=%.6f, lon=%.6f) timezone for monitored trip %s",
+                from.lat,
+                from.lon,
+                id
+            );
+            throw new RuntimeException(message);
+        } else {
+            return fromZoneId.get();
+        }
+    }
```
evansiroky: Please replace this with the `getTimezoneForTargetLocation` method that I made in #75 (see [here](https://github.com/ibi-group/otp-middleware/pull/75/files#diff-a3ee245c312c932d083bd39feecf69f3R331)) so that we get the timezone from the proper location depending on whether the trip is arrive by or depart at.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -56,4 +107,30 @@ private void verifyBelowMaxNumTrips(String userId, Request request) {
...
+        ItineraryExistenceChecker itineraryChecker = new ItineraryExistenceChecker(OtpDispatcher::sendOtpPlanRequest);
+        try {
+            ItineraryExistenceChecker.Result checkResult = itineraryChecker.checkAll(ItineraryUtils.getItineraryExistenceQueries(trip));
+            if (!checkResult.allItinerariesExist) {
+                logMessageAndHalt(
+                    request,
+                    HttpStatus.BAD_REQUEST_400,
+                    "An itinerary does not exist for some of the monitored days for the requested trip."
+                );
+            }
```
evansiroky: In order for the UI to be able to show which days of the week the trip is possible, shouldn't this return some kind of map of the days of the week that the trip is possible?

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,160 @@
...
+    public void testItineraryDepartsSameDay() {
+        // All times EDT (GMT-04:00)
+        testItineraryDepartsSameDay(QUERY_TIME, true,
+            1597302000000L, // August 13, 2020 3:00:00 AM
+            1597377599000L, // August 13, 2020 11:59:59 PM
+            1597388399000L // August 14, 2020 02:59:59 AM, considered to be Aug 13.
+        );
+        testItineraryDepartsSameDay("1:23", true,
+            1597215600000L, // August 12, 2020 3:00:00 AM
+            1597291199000L, // August 12, 2020 11:59:59 PM
```
evansiroky: I'd actually expect both of these to return false. The logic is that the request date is still August 13th, so one would not expect an itinerary with a start time on August 12th to show up at all. Despite August 12th being the same service day at the time, it is not the same request date.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,185 @@
...
+        if (itinerary != null) {
+            trip.itinerary = itinerary;
+            trip.initializeFromItinerary();
+        }
+    }
+
+    /**
+     * @return true if the itinerary's startTime is one the same day as the day of the specified date and time.
+     */
+    public static boolean itineraryDepartsSameDay(Itinerary itinerary, String date, String time, ZoneId zoneId) {
```
evansiroky: All of this method is oriented around an assumption that the request date and time is relative to a "depart at" kind of query. There needs to be logic related to an "arrive by" query too. What about a hypothetical long trip that was arriving by 3:01am that had to depart at 11:59pm the previous day?

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
evansiroky: Actually, the second one might make sense for a trip that was "arriving by" 1:23am.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,160 @@
...
+                    expected ? "should" : "should not",
+                    QUERY_DATE,
+                    time
+                )
+            );
+        }
+    }
+
+    @Test
+    public void testItineraryDepartsSameDay() {
```
evansiroky: I like these tests, but think we need some more for trips that "arrive by".

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,185 @@
...
+        // Insert '?' so others can parse the resulting query string.
+        return "?" + toQueryString(params);
+    }
+
+    /**
+     * Replaces the trip itinerary field value with one from the verified itineraries queried from OTP.
+     * The ui_activeItinerary query parameter is used to determine which itinerary to use,
+     * TODO/FIXME: need a trip resemblance check to supplement the ui_activeItinerary param used in this function.
+     */
+    public static void updateTripWithVerifiedItinerary(MonitoredTrip trip, List<Itinerary> verifiedItineraries) throws URISyntaxException {
```
evansiroky: This method can result in a completely different itinerary being saved than what the user selected. We should discuss things a little more so that we're on the same page of what to do when a user wants to save an itinerary that is only made possible due to unique realtime data. For example, say a bus A is running early and is able to get to a bus stop just before Bus B leaves because Bus B is late. That result won't ever show up in these additional checks here. We probably need some kind of way to determine if an itinerary is typically possible at all or was made possible by the current day's realtime conditions.

evansiroky (changes requested):
See comments. I think we need to have a work session around this topic and others.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: This method is called on PUT or POST. But the next step is indeed to create an endpoint for the UI to report the checks... Should that be in this PR or a separate one?

binh-dam-ibigroup: I guess I'll go ahead and build the endpoint for the UI too.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -27,11 +35,54 @@ public MonitoredTripController(String apiPrefix) {
     MonitoredTrip preCreateHook(MonitoredTrip monitoredTrip, Request req) {
         verifyBelowMaxNumTrips(monitoredTrip.userId, req);
         monitoredTrip.initializeFromItinerary();
+        ItineraryExistenceChecker.Result checkResult = checkItineraryExistence(monitoredTrip, req);
```
br648: Is it worth creating E2E tests for this, similar to ApiUserFlowTest?

br648 (commented):
Looking good so far. I understand that additional work is still to be done under this PR, so will review again once that is complete.

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
binh-dam-ibigroup: Fixed in https://github.com/ibi-group/otp-middleware/pull/77/commits/a2e1ec3403a6fa2f002dcfd1c59f89594a6ca3a8

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: I added the arrive-by tests in https://github.com/ibi-group/otp-middleware/pull/77/commits/0aba7da8c43c1e2b26da517ee70d6f821b9a933b

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: Cast your votes for adding e2e tests.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
binh-dam-ibigroup: I added the arrive-by tests in https://github.com/ibi-group/otp-middleware/pull/77/commits/0aba7da8c43c1e2b26da517ee70d6f821b9a933b.

binh-dam-ibigroup:
I'm going ahead and submit this for review. Note that this PR is now dependent on merging #75 as it has some utility code used in this PR. To make the review easier, I'm setting the base branch to that of #75. I'm also adding the BLOCKED label pending merging of #75.

src/main/java/org/opentripplanner/middleware/controllers/api/ItineraryCheckController.java
```diff
@@ -0,0 +1,121 @@
...
+            // Convert the dates in the result to weekdays,
+            // and fill the same-day itineraries in each day, if any.
+            // Note: At this time, the endpoint checks all days of the week at once before returning a response.
+            for (Map.Entry<String, OtpResponse> r : checkResult.labeledResponses.entrySet()) {
+                String dateString = r.getKey();
+                LocalDate date = DateTimeUtils.getDateFromString(dateString, DateTimeUtils.DEFAULT_DATE_FORMAT_PATTERN);
+                if (r.getValue().plan != null) {
+                    // Only keep same-day itineraries.
+                    List<Itinerary> sameDayItineraries = ItineraryUtils.getSameDayItineraries(r.getValue().plan.itineraries, trip, dateString);
+                    if (sameDayItineraries.size() != 0) {
```
br648: Could replace with !sameDayItineraries.isEmpty()

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -36,6 +44,12 @@ MonitoredTrip preCreateHook(MonitoredTrip monitoredTrip, Request req) {
                 e
             );
         }
+        ItineraryExistenceChecker.Result checkResult = checkItineraryExistence(monitoredTrip, req);
+
+        // Replace the provided trip's itinerary with a verified, non-real-time version of it.
+        if (checkResult != null) {
+            updateTripWithVerifiedItinerary(monitoredTrip, req, checkResult.labeledResponses);
+        }
```
br648: According to Intellij, the preCreateHook and preUpdateHook contain duplicate code. I think this should be refactored into a single method?

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -90,4 +99,34 @@ public OtpDispatcherResponse clone() {
 //        clonedObject.responseBody = this.responseBody;
 //        return clonedObject;
     }
+
```
br648: Not strictly part of this PR, but what is happening with the clone method above this? I think the comments should at least be removed.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,190 @@
...
+     * Checks that the specified itinerary is on the same day as the specified date/time.
+     * @param itinerary the itinerary to check.
+     * @param date the request date to check.
+     * @param time the request time to check.
+     * @param checkArrival true to check the itinerary endtime, false to check the startTime.
+     * @return true if the itinerary's startTime or endTime is one the same day as the day of the specified date and time.
+     */
+    public static boolean isSameDay(Itinerary itinerary, String date, String time, ZoneId zoneId, boolean checkArrival) {
+        // TODO: Make SERVICEDAY_START_HOUR an optional config parameter.
+        final int SERVICEDAY_START_HOUR = 3;
```
br648: I think this should be SERVICE_DAY_START_HOUR

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
```diff
@@ -0,0 +1,79 @@
...
+public class ItineraryExistenceCheckerTest {
+    /**
+     * Map query strings to mock OTP responses.
+     * We use {@link HashMap#get} function to mock {@link OtpDispatcher#sendOtpPlanRequest}
+     * (both have the same signature).
+     */
+    private static HashMap<String, OtpDispatcherResponse> queryToResponse;
+
+    /** Mock (significantly abbreviated) OTP response when an itinerary is not found. */
+    private final static String MOCK_RESPONSE_WITH_ERROR = "{\"requestParameters\":{},\"error\":{\"id\":404,\"msg\":\"No trip found...\",\"message\":\"PATH_NOT_FOUND\",\"noPath\":true},\"debugOutput\":{}}";
```
br648: Perhaps change to JSON ObjectNodes?

br648 (approved):
This looks good to me. Some minor items, but approving.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,190 @@
...
+     * Creates a map of new query strings based on the one provided,
+     * with the date changed to the desired one.
+     * @param params a map of the base OTP query parameters.
+     * @param dates a list of the desired dates in YYYY-MM-DD format.
+     * @return a map of query strings with, and indexed by the specified dates.
+     */
+    public static Map<String, String> getQueriesFromDates(Map<String, String> params, List<String> dates) {
+        // Create a copy of the original params in which we change the date.
+        Map<String, String> paramsCopy = new HashMap<>(params);
+        Map<String, String> result = new HashMap<>();
```
evansiroky: Just my personal preference, `result` could be named `queryParamsByDate`.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
evansiroky: Or import a file from the resources folder.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @BeforeAll
+    public static void setUp() throws IOException {
+        // Mock OTP responses for when itineraries exist for a query.
+        String mockPlanResponse = FileUtils.getFileContents(
+            TEST_RESOURCE_PATH + "persistence/planResponse.json"
+        );
+        queryToResponse = new HashMap<>();
+        URI uri = URI.create("http://www,example.com");
```
evansiroky: typo of a comma in the URL, but I guess it doesn't affect anything else.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
binh-dam-ibigroup: Good idea! I updated the class to use the error response file that is already in the resources folder. See https://github.com/ibi-group/otp-middleware/pull/77/commits/d331144c173deeb3e2e5cc8542f174d42312ff80.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
binh-dam-ibigroup: I don't know... Those changes are from commit 38dafa4b71983ae5f2b8f47bc6e7568007ba5fb1 by @landonreed.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: Good catch! Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/d331144c173deeb3e2e5cc8542f174d42312ff80 along with your other suggestions.

src/main/java/org/opentripplanner/middleware/controllers/api/ItineraryCheckController.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    private static ItineraryExistenceResult checkItinerary(Request request, Response response) {
+        ItineraryExistenceResult result = new ItineraryExistenceResult();
+
+        try {
+            MonitoredTrip trip = getPOJOFromRequestBody(request, MonitoredTrip.class);
+            trip.initializeFromItineraryAndQueryParams();
+
+            ItineraryExistenceChecker itineraryChecker = new ItineraryExistenceChecker(OtpDispatcher::sendOtpPlanRequest);
+            ItineraryExistenceChecker.Result checkResult = itineraryChecker.checkAll(ItineraryUtils.getItineraryExistenceQueries(trip, true), trip.isArriveBy());
```
evansiroky: Line is too long, please break up.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -90,4 +99,30 @@ public OtpDispatcherResponse clone() {
...
+        HashMap<String, String> reqParams = response.requestParameters;
+        if (plan != null && plan.itineraries != null && reqParams != null) {
+            String requestDate = reqParams.get(DATE_PARAM);
+            String requestTime = reqParams.get(TIME_PARAM);
+            if (requestDate != null && requestTime != null) {
+                // Get the zone id for this plan. Don't search for itineraries in this plan if the zone cannot be determined.
+                Optional<ZoneId> fromZoneId = getZoneIdForCoordinates(plan.from.lat, plan.from.lon);
+                if (fromZoneId.isPresent()) {
+                    for (Itinerary itinerary : plan.itineraries) {
+                        if (ItineraryUtils.isSameDay(itinerary, requestDate, requestTime, fromZoneId.get(), checkArrival)) {
```
evansiroky: Line is too long, please break up.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param time the request time to check.
+     * @param checkArrival true to check the itinerary endtime, false to check the startTime.
+     * @return true if the itinerary's startTime or endTime is one the same day as the day of the specified date and time.
+     */
+    public static boolean isSameDay(Itinerary itinerary, String date, String time, ZoneId zoneId, boolean checkArrival) {
+        // TODO: Make SERVICE_DAY_START_HOUR an optional config parameter.
+        final int SERVICE_DAY_START_HOUR = 3;
+        DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT_PATTERN);
+
+        ZonedDateTime startDate = ZonedDateTime.ofInstant(itinerary.getStartOrEndTime(checkArrival).toInstant(), zoneId);
```
evansiroky: The `Itinerary#getStartOrEndTime` is only ever called here. I'd recommend just extracting the logic and inline it here.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Checks that the specified itinerary is on the same day as the specified date/time.
+     * @param itinerary the itinerary to check.
+     * @param date the request date to check.
+     * @param time the request time to check.
+     * @param checkArrival true to check the itinerary endtime, false to check the startTime.
+     * @return true if the itinerary's startTime or endTime is one the same day as the day of the specified date and time.
+     */
+    public static boolean isSameDay(Itinerary itinerary, String date, String time, ZoneId zoneId, boolean checkArrival) {
```
evansiroky: The `checkArrival` variable is used here, but in the `Itinerary#getStartOrEndTime` it's then called `isArrival` and in `MonitoredTrip` there is a method called `isArriveBy`. In OTP there is simply an `arriveBy` request parameter. Let's be consistent and try to use `arriveBy` everywhere.

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
```diff
@@ -0,0 +1,66 @@
...
+        this.otpDispatcherFunction = otpDispatcherFunction;
+    }
+
+    /**
+     * Checks that, for each query provided, an itinerary exists.
+     * @param labeledQueries a map containing the queries to check, each query having a key or
+     *                       label that will be used in Result.labeledResponses for easy identification.
+     * @return An object with a map of results and summary of itinerary existence.
+     */
+    public Result checkAll(Map<String, String> labeledQueries, boolean checkArrival) {
```
evansiroky: I'd recommend renaming `chechArrival` to `tripIsArriveBy`.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -98,6 +97,18 @@ public void clearAlerts() {
         }
     }
 
+    public Date getStartOrEndTime(boolean isArrival) {
+        return isArrival ? endTime : startTime;
+    }
+
+    public void setStartOrEndTime(Date startOrEnd, boolean isArrival) {
```
evansiroky: This is only ever used during testing. Please move this logic to a test class.

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
```diff
@@ -0,0 +1,66 @@
...
+         * and values as OTP responses for the corresponding queries.
+         */
+        public final Map<String, OtpResponse> labeledResponses;
+
+        private Result(boolean itinerariesExist, Map<String, OtpResponse> labeledResponses) {
+            this.allItinerariesExist = itinerariesExist;
+            this.labeledResponses = labeledResponses;
+        }
+    }
+}
```
evansiroky: This file is small enough that it'd probably be fine to merge it into the `ItineraryUtils` class.

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
```diff
@@ -0,0 +1,66 @@
...
+    private final Function<String, OtpDispatcherResponse> otpDispatcherFunction;
+
+    /**
+     * @param otpDispatcherFunction a function that takes String as input and returns {@link OtpDispatcherResponse},
+     *                              example: {@link OtpDispatcher#sendOtpPlanRequest}.
+     */
+    public ItineraryExistenceChecker(Function<String, OtpDispatcherResponse> otpDispatcherFunction) {
+        if (otpDispatcherFunction == null) throw new NullPointerException();
+        this.otpDispatcherFunction = otpDispatcherFunction;
+    }
```
evansiroky: I'm not a huge fan of this constructor because it appears the sole purpose of it is to inject testing code. I think it'd be better to have the tests always use the `OtpDispatcher` and then properly mock some requests/responses using the `TestUtils.setupOtpMocks` method.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
```diff
@@ -0,0 +1,80 @@
...
+        queryToResponse.put("exist2", new OtpDispatcherResponse(mockPlanResponse, uri));
+        queryToResponse.put("exist3", new OtpDispatcherResponse(mockPlanResponse, uri));
+
+        // Query for which an itinerary is not found.
+        queryToResponse.put("not found", new OtpDispatcherResponse(mockPlanErrorResponse, uri));
+    }
+
+    @Test
+    public void testAllTripsExist() {
+        ItineraryExistenceChecker tripChecker = new ItineraryExistenceChecker(queryToResponse::get);
```
evansiroky: I can see how it does simplify testing the class, but it still seems a bit problematic. Is there a reason why this is preferred over using the `TestUtils.setupOtpMocks` method?

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,252 @@
...
+    public static final long _2020_08_12__03_00_00_EDT = 1597215600000L; // Aug 12, 2020 3:00:00 AM
+    public static final long _2020_08_12__23_59_59_EDT = 1597291199000L; // Aug 12, 2020 11:59:59 PM
+    public static final long _2020_08_13__02_59_59_EDT = 1597301999000L; // Aug 13, 2020 2:59:59 AM, considered to be Aug 12.
+    public static final long _2020_08_13__03_00_00_EDT = 1597302000000L; // Aug 13, 2020 3:00:00 AM
+    public static final long _2020_08_13__23_59_59_EDT = 1597377599000L; // Aug 13, 2020 11:59:59 PM
+    public static final long _2020_08_14__02_59_59_EDT = 1597388399000L; // Aug 14, 2020 2:59:59 AM, considered to be Aug 13.
+    public static final long _2020_08_14__03_00_00_EDT = 1597388400000L; // Aug 14, 2020 3:00:00 AM
+
+    @Test
+    public void testGetQueriesFromDates() throws URISyntaxException {
```
evansiroky: Please add JavaDoc to each test case.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,252 @@
...
+        );
+        OtpDispatcherResponse otpDispatcherResponse = new OtpDispatcherResponse(mockResponse, URI.create("http://www.example.com"));
+        List<Itinerary> itineraries = otpDispatcherResponse.getResponse().plan.itineraries;
+
+        ItineraryUtils.updateTripWithVerifiedItinerary(trip, itineraries);
+
+        Assertions.assertEquals(itineraries.get(1), trip.itinerary);
+    }
+
+    private void testIsSameDay(String time, boolean isArrival, boolean expected, Long... startTimes) throws URISyntaxException {
```
evansiroky: This method along with all the methods that call it could be combined into a single ParameterizedTest. The test result looks a lot cleaner and if one of the test cases fails, the others will still be ran.

evansiroky (changes requested):
See various comments. 

Also, https://github.com/ibi-group/otp-middleware/pull/77#discussion_r491251522 is still something that has been unresolved. I am increasingly thinking that making OTP plan requests is the wrong way to go about checking for whether the itinerary is valid in the future. The OTP plan endpoint is not designed to have any comparison or checking abilities upon an itinerary that has been returned. Instead, it is portal to performing a shortest-path optimization given input parameters. 

I think a better solution would involve building in a new endpoint/set of calculations into OTP itself which would be able to calculate the itinerary validity. In doing so, OTP would have much greater detail since it would have all of the transit/road network information readily available. 

I'd hold off on further changes to this PR until we get on the same page with how to check for itinerary validity.

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
binh-dam-ibigroup: Good suggestion.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
binh-dam-ibigroup: I was trying to avoid setting up a mock server altogether, but sounds like you feel strongly about doing that instead.

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
binh-dam-ibigroup: Merged in https://github.com/ibi-group/otp-middleware/pull/77/commits/91812109313baeba91b81c2fea66fb1fa954cca9.

src/test/java/org/opentripplanner/middleware/utils/ItineraryExistenceCheckerTest.java
binh-dam-ibigroup: Using mockOtpServer as of https://github.com/ibi-group/otp-middleware/pull/77/commits/81d7fa65c0868e2811e6aa4889919b9e419ecb71

src/main/java/org/opentripplanner/middleware/utils/ItineraryExistenceChecker.java
binh-dam-ibigroup: Using mpckOtpServer as of https://github.com/ibi-group/otp-middleware/pull/77/commits/81d7fa65c0868e2811e6aa4889919b9e419ecb71

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/097c7f2bae3c9e29bcd6894de370cedefdaa3e12.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/097c7f2bae3c9e29bcd6894de370cedefdaa3e12.

br648 (approved):
Still looks good to me.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -90,4 +99,32 @@ public OtpDispatcherResponse clone() {
...
+    public Itinerary findItineraryDepartingSameDay(boolean tripIsArriveBy) {
+        OtpResponse response = this.getResponse();
+        TripPlan plan = response.plan;
+        HashMap<String, String> reqParams = response.requestParameters;
+        if (plan != null && plan.itineraries != null && reqParams != null) {
+            String requestDate = reqParams.get(DATE_PARAM);
+            String requestTime = reqParams.get(TIME_PARAM);
+            if (requestDate != null && requestTime != null) {
+                // Get the zone id for this plan. Don't search for itineraries in this plan if the zone cannot be determined.
+                Optional<ZoneId> fromZoneId = getZoneIdForCoordinates(plan.from.lat, plan.from.lon);
```
evansiroky: I removed this ZoneId lookup in #75 as it turns out that this is not how OTP deals with timezones. Instead you'll want to use `DateTimeUtils. getOtpZoneId`.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Obtains the monitored dates for the given trip, for which we should check that itineraries exist.
+     * The dates include each day to be monitored in a 7-day window starting from the trip's query start date.
+     * @param trip The trip from which to extract the monitored dates to check.
+     * @param checkAllDays true if all days of the week are included regardless of the trip's monitored days.
+     * @return A list of date strings in YYYY-MM-DD format corresponding to each day of the week to monitor.
+     */
+    public static List<String> getDatesToCheckItineraryExistence(MonitoredTrip trip, boolean checkAllDays) throws URISyntaxException {
+        List<String> result = new ArrayList<>();
+        ZoneId zoneId = trip.timezoneForTargetLocation();
```
evansiroky: I removed this ZoneId lookup in #75 as it turns out that this is not how OTP deals with timezones. Instead you'll want to use `DateTimeUtils. getOtpZoneId`.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+     * Obtains the monitored dates for the given trip, for which we should check that itineraries exist.
+     * The dates include each day to be monitored in a 7-day window starting from the trip's query start date.
+     * @param trip The trip from which to extract the monitored dates to check.
+     * @param checkAllDays true if all days of the week are included regardless of the trip's monitored days.
+     * @return A list of date strings in YYYY-MM-DD format corresponding to each day of the week to monitor.
+     */
+    public static List<String> getDatesToCheckItineraryExistence(MonitoredTrip trip, boolean checkAllDays) throws URISyntaxException {
```
evansiroky: Line is too long, please break up.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,235 @@
...
+        List<String> result = new ArrayList<>();
+        ZoneId zoneId = trip.timezoneForTargetLocation();
+        Map<String, String> params = trip.parseQueryParams();
+
+        // Start from the query date, if available.
+        // If there is no query date, start from today.
+        String queryDateString = params.getOrDefault(DATE_PARAM, DateTimeUtils.getStringFromDate(DateTimeUtils.nowAsLocalDate(), DEFAULT_DATE_FORMAT_PATTERN));
```
evansiroky: Line is too long, please break up.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,235 @@
...
+        LocalDate queryDate = DateTimeUtils.getDateFromString(queryDateString, DEFAULT_DATE_FORMAT_PATTERN);
+
+        // Get the trip time.
+        ZonedDateTime queryZonedDateTime = ZonedDateTime.of(queryDate, LocalTime.of(trip.tripTimeHour(), trip.tripTimeMinute()), zoneId);
```
evansiroky: Line is too long, please break up.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+
+        Assertions.assertEquals(itineraries.get(1), trip.itinerary);
+    }
+
+    @ParameterizedTest
+    @MethodSource("createSameDayTestCases")
+    void testIsSameDay(SameDayTestCase testCase) throws URISyntaxException {
+        MonitoredTrip trip = makeTestTrip(testCase.isArrival);
+        // The time zone for trip (and testCase.tripTime) is US Eastern per trip location.
+        ZoneId zoneId = trip.timezoneForTargetLocation();
```
evansiroky: I removed this ZoneId lookup in #75 as it turns out that this is not how OTP deals with timezones. Instead you'll want to use `DateTimeUtils. getOtpZoneId`.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    private static List<SameDayTestCase> createSameDayTestCases() {
+        return List.of(
+            // Same-day departures
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__03_00_00_EDT, false, true),
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__23_59_59_EDT, false, true),
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__02_59_59_EDT, false, true),
+            new SameDayTestCase("1:23", _2020_08_12__03_00_00_EDT, false, true),
+            new SameDayTestCase("1:23", _2020_08_12__23_59_59_EDT, false, true),
```
evansiroky: Since OTP shouldn't ever return a "depart at" itinerary that begins before the date + time request parameters, I don't believe such a case would occur. But if it did, I guess this is what I'd expect for being the same day though, so I guess it's fine to keep these test cases.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            new SameDayTestCase("1:23", _2020_08_13__02_59_59_EDT, false, true),
+
+            // Not same-day departures
+            new SameDayTestCase(QUERY_TIME, _2020_08_12__23_59_59_EDT, false, false),
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__02_59_59_EDT, false, false),
```
evansiroky: Since OTP shouldn't ever return a "depart at" itinerary that begins before the date + time request parameters, I don't believe such a case would occur. But if it did, I guess this is what I'd expect for being the same day though, so I guess it's fine to keep these test cases.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__03_00_00_EDT, false, false),
+            new SameDayTestCase("1:23", _2020_08_13__03_00_00_EDT, false, false),
```
evansiroky: This one threw me off a bit, but I now see that the line is drawn right at 3am, so this makes sense.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            new SameDayTestCase("1:23", _2020_08_13__23_59_59_EDT, false, false),
+            new SameDayTestCase("1:23", _2020_08_14__02_59_59_EDT, false, false),
+
+            // Same-day arrivals
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__03_00_00_EDT, true, true),
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__23_59_59_EDT, true, true),
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__02_59_59_EDT, true, true),
```
evansiroky: Since OTP shouldn't ever return a "arrive by" itinerary that ends after the date + time request parameters, I don't believe such a case would occur. But if it did, I guess this is what I'd expect for being the same day though, so I guess it's fine to keep these test cases.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            this.checkAllDays = checkAllDays;
+            this.dates = dates;
+        }
+    }
+
+    /**
+     * Data structure for the same-day test.
+     */
+    private static class SameDayTestCase {
+        public final boolean isArrival;
```
evansiroky: For consistency, please rename to `isArriveBy`

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+        List<Itinerary> processedItineraries = ItineraryUtils.getSameDayItineraries(itineraries, trip, QUERY_DATE);
+        Assertions.assertEquals(2, processedItineraries.size());
+        Assertions.assertTrue(processedItineraries.contains(itineraries.get(0)));
+        Assertions.assertTrue(processedItineraries.contains(itineraries.get(1)));
+    }
+
+    /**
+     * Helper method to create a bare-bones itinerary with start or end time.
+     */
+    private Itinerary simpleItinerary(Long startTime, boolean isArrival) {
```
evansiroky: Please rename `startTime` to `targetEpochMillis`.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+        public final boolean isArrival;
+        public final boolean shouldBeSameDay;
+        public final String timeOfDay;
+        public final Long tripTime;
+
+        public SameDayTestCase(String timeOfDay, Long tripTime, boolean isArrival, boolean shouldBeSameDay) {
+            this.isArrival = isArrival;
+            this.shouldBeSameDay = shouldBeSameDay;
+            this.timeOfDay = timeOfDay;
+            this.tripTime = tripTime;
```
evansiroky: Please always name a variable to have the specific time units the variable is supposed to represent. Consider naming this `tripTargetTimeEpochMillis`.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            // Not same-day departures
+            new SameDayTestCase(QUERY_TIME, _2020_08_12__23_59_59_EDT, false, false),
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__02_59_59_EDT, false, false),
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__03_00_00_EDT, false, false),
+            new SameDayTestCase("1:23", _2020_08_13__03_00_00_EDT, false, false),
+            new SameDayTestCase("1:23", _2020_08_13__23_59_59_EDT, false, false),
+            new SameDayTestCase("1:23", _2020_08_14__02_59_59_EDT, false, false),
+
+            // Same-day arrivals
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__03_00_00_EDT, true, true),
```
evansiroky: This one is kind of tricky. Since it does end on the desired date, I suppose it makes sense. However, it implies that the trip has started on the previous day. If this ever comes up it's probably due to some really infrequent routes that run at odd times, so I guess it's fine to keep for now.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__23_59_59_EDT, true, true),
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__02_59_59_EDT, true, true),
+            new SameDayTestCase("1:23", _2020_08_12__03_00_00_EDT, true, true),
+            new SameDayTestCase("1:23", _2020_08_12__23_59_59_EDT, true, true),
+            new SameDayTestCase("1:23", _2020_08_13__02_59_59_EDT, true, true),
```
evansiroky: Since OTP shouldn't ever return a "arrive by" itinerary that ends after the date + time request parameters, I don't believe such a case would occur. But if it did, I guess this is what I'd expect for being the same day though, so I guess it's fine to keep these test cases.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+
+            // Not same-day arrivals
+            new SameDayTestCase(QUERY_TIME, _2020_08_12__23_59_59_EDT, true, false),
+            new SameDayTestCase(QUERY_TIME, _2020_08_13__02_59_59_EDT, true, false),
+            new SameDayTestCase(QUERY_TIME, _2020_08_14__03_00_00_EDT, true, false),
```
evansiroky: Since OTP shouldn't ever return a "arrive by" itinerary that ends after the date + time request parameters, I don't believe such a case would occur. But if it did, I guess this is what I'd expect for being the same day though, so I guess it's fine to keep these test cases.

evansiroky: This comment applies to the remaining 3 cases as well.

evansiroky (changes requested):
Looking good, just a few changes requested:

- some minor formatting items
- renaming some variables to note the type of time representation
- updating to the timezone finding methodology that was changed in the latest edits of #75

evansiroky (approved):
Good to go after addressing https://github.com/ibi-group/otp-middleware/pull/77#discussion_r502725596.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: This was addressed in https://github.com/ibi-group/otp-middleware/pull/77/commits/78e31e7b50bcda047f65ddf1cbbbe506dbb8e455#diff-5a730e4259529c8d307574db916d01f209537dd2f60746a86a6e9b022ce8cca4R286.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: Okay, this was the comment I missed. Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/36221dbf5c6ad2a52748e2995e4eb06c00be790a.

src/main/java/org/opentripplanner/middleware/controllers/api/ItineraryCheckController.java
```diff
@@ -0,0 +1,120 @@
...
+import static com.beerboy.ss.descriptor.MethodDescriptor.path;
+import static org.opentripplanner.middleware.utils.HttpUtils.JSON_ONLY;
+import static org.opentripplanner.middleware.utils.JsonUtils.getPOJOFromRequestBody;
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+/**
+ * This endpoint checks is used by web clients or other APIs to let users know
+ * whether a given itinerary exists on each day of the week.
+ */
+public class ItineraryCheckController implements Endpoint {
```
landonreed: This design seems a little clunky to me. This entire controller does not need to exist. Here's what I propose:
1. In the UI, when a user clicks save trip, we send a request to createMonitoredTrip.
2. As part of the creation of that monitored trip, we kick off a checkItinerary job (Runnable), that gets tracked in a status object as part of the MonitoredTrip it sends back in the create trip response.
3. Meanwhile in the UI, we process the response from create trip and keep polling the get monitored trip endpoint to check the check itinerary status until it is complete. As itinerary existence is confirmed on the various days, that status gets updated.
4. Once the check is complete, the monitored trip status is fully updated and the user can proceed with confirming the trip/notification details.

landonreed (changes requested):
I'm proposing a different way to to handle this itinerary check (see comment).

src/main/java/org/opentripplanner/middleware/controllers/api/ItineraryCheckController.java
landonreed: Oh, actually the check already happens in `initializeTripAndSetVerifiedItinerary`. I need to dig into this a little more to understand why the separate controller exists then.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /**
+     * Class to pass the results of the OTP itinerary checks.
+     */
+    public static class Result {
+        /** Whether all itineraries checked exist. */
+        public final boolean allItinerariesExist;
+        /**
+         * A map with the same keys as the input from checkAll,
```
landonreed: What is `checkAll`? I need to read more, but I'm a bit suspect of this labeledQueries approach.

src/main/java/org/opentripplanner/middleware/controllers/api/ItineraryCheckController.java
binh-dam-ibigroup: This controller is for the UI to show which days are possible or not.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
binh-dam-ibigroup: `checkAll` determines whether all days of the week are checked. The UI needs to check all days to report which days are not available, whereas `monitoredtrip` PUT/POST only needs to checks trip existence on the days selected for monitoring.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -0,0 +1,166 @@
...
+        return checkItineraryExistence(trip, checkAllDays,false);
+    }
+
+    /**
+     * Variant of checkItineraryExistence used for tests,
+     * to ensure that the order of the dates to check matches the order of the mock OTP responses for those dates.
+     */
+    public static ItineraryExistence checkItineraryExistenceOrdered(MonitoredTrip trip, boolean checkAllDays) throws URISyntaxException {
+        return checkItineraryExistence(trip, checkAllDays,true);
+    }
```
evansiroky: This doesn't seem particularly helpful to make two new methods just to set a boolean argument. These could probably be removed.

evansiroky (approved):
Refactor looks good. There's just one item I don't feel strongly about.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
binh-dam-ibigroup: Probably the `ordered` variant doesn't need to be here. I'd rather keep the other one to hide the sorting argument, though.

landonreed (changes requested):
See suggested changes in #93.

evansiroky (approved):

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -0,0 +1,255 @@
...
+            ? resultForDay.itineraries.get(0)
+            : null;
+    }
+
+    /**
+     * @return A collection of days of week (and first date found) for which the trip is not possible.
+     */
+    @JsonIgnore
+    @BsonIgnore
+    public Collection<String> getInvalidDaysOfWeek() {
```
landonreed: Since this is no longer returning formatted date strings, it might make more sense just to return a string message instead of a collection of Strings. In other words, the join the strings once again. I originally wrote it to return a collection of dates in case that was needed for some other purpose, but now it seems more specialized for just writing out a message.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -0,0 +1,255 @@
...
+        /**
+         * Holds a list of valid dates for which an itinerary exists.
+         */
+        public List<String> validDates = new ArrayList<>();
+
+        /**
+         * Holds any matching itineraries (sorted by date) for the applicable day of the week.
+         * TODO: remove itineraries from list of returned fields in swagger.
+         */
+        public transient List<Itinerary> itineraries = new ArrayList<>();
```
landonreed, line 235: Does this need to be public?

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -0,0 +1,255 @@
...
+         * True if and only if an itinerary is available for all dates tested for existence.
+         */
+        @JsonProperty
+        public boolean isValid() {
+            return invalidDates.size() == 0;
+        }
+
+        /**
+         * Method and argument solely needed for accepting request contents that include a `valid` field
+         * (e.g. for checking a monitored trip with previous existence check record).
```
landonreed: This doesn't make a ton of sense to me.

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -859,6 +878,21 @@ definitions:
...
+    properties:
+      invalidDates:
+        type: "array"
+        items:
+          type: "string"
+      validDates:
+        type: "array"
+        items:
+          type: "string"
+      itineraries:
```
landonreed, line 892: Not a big deal for me, but the `isValid` JSON prop is not showing up here.

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -964,6 +1005,36 @@ definitions:
...
         format: "int32"
+  ItineraryExistence:
+    type: "object"
+    properties:
+      otpRequests:
+        type: "array"
+        items:
+          $ref: "#/definitions/OtpRequest"
+      referenceItinerary:
+        $ref: "#/definitions/Itinerary"
```
landonreed, line 1016: More non-blocking spark-swagger weirdness: doesn't seem like private fields should be shown.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -0,0 +1,241 @@
...
+    @AfterEach
+    public void tearDownAfterTest() {
+        TestUtils.resetOtpMocks();
+    }
+
+    /**
+     * Test case in which all itineraries exist and result.allCheckedDatesAreValid should be true.
+     */
+    @Test
+    public void testAllItinerariesExist() throws URISyntaxException {
```
landonreed: Just a minor comment on test names: we don't usually call them `testMethodName`. I would prefer all these test names match our other syntax. E.g., `canCheckValidItineraries` or something.

landonreed (changes requested):
Looking great. Just a couple of final comments.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
binh-dam-ibigroup: This needs to be public for the tests in `ItineraryUtilsTest`.
Swagger will publish all fields in a type regardless of scope according to https://github.com/manusant/spark-swagger/blob/master/src/main/java/io/github/manusant/ss/factory/DefinitionsFactory.java#L36, so we can either leave it listed in the swagger doc for future use, or remove it during post-processing of the generated swagger. Which one should we do?

src/main/resources/latest-spark-swagger-output.yaml
binh-dam-ibigroup: Same reason as for the `itineraries` field. Although there is a getter and setter,`valid` is not a field, so spark-swagger just ignores it.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/a655afcdf5b8e76d68cfecf0e6ba3edb2a5d43fa.

src/main/resources/latest-spark-swagger-output.yaml
binh-dam-ibigroup: I added code to remove these fields in the public docs in https://github.com/ibi-group/otp-middleware/pull/77/commits/d26baf619770dce9e11667961912b80261de090c.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
binh-dam-ibigroup: I added code to remove these fields in the public docs in https://github.com/ibi-group/otp-middleware/pull/77/commits/d26baf619770dce9e11667961912b80261de090c.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/d26baf619770dce9e11667961912b80261de090c..

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/d26baf619770dce9e11667961912b80261de090c.

src/main/resources/latest-spark-swagger-output.yaml
binh-dam-ibigroup: I added the field in the public docs in https://github.com/ibi-group/otp-middleware/pull/77/commits/d26baf619770dce9e11667961912b80261de090c.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
landonreed: Forgot the javadoc (!)

binh-dam-ibigroup: Despair... Updated in https://github.com/ibi-group/otp-middleware/pull/77/commits/99a34ede08b01c3684b17a33315869c1e36ce3d4.

landonreed (approved):
Let's go ahead and merge. Future testing might suggest further changes, but I think is good for now!

PR #78, opened 2020-09-18
Unit test tidy

br648:
### Checklist

### Description

Updated unit tests to remove all DB objects after tests have completed. Also adddressed a few other items that came to light when I started looking into this.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/models/ApiUser.java
+++ b/src/main/java/org/opentripplanner/middleware/models/ApiUser.java
@@ -40,13 +40,16 @@ public class ApiUser extends AbstractUser {
     public String name;
 
     /**
-     * Delete API user's API keys (from AWS) and self (from Mongo).
+     * Delete API user details including Auth0 user.
      */
     @Override
     public boolean delete() {
         return delete(true);
     }
 
+    /**
+     * Delete API user's API keys (from AWS), self (from Mongo). Optionally delete user from Auth0.
+     */
     public boolean delete(boolean deleteAuth0User) {
         for (ApiKey apiKey : apiKeys) {
             if (!ApiGatewayUtils.deleteApiKey(apiKey)) {
--- a/src/main/java/org/opentripplanner/middleware/models/ApiUser.java
+++ b/src/main/java/org/opentripplanner/middleware/models/ApiUser.java
@@ -44,19 +44,26 @@ public class ApiUser extends AbstractUser {
      */
     @Override
     public boolean delete() {
+        return delete(true);
+    }
+
+    public boolean delete(boolean deleteAuth0User) {
         for (ApiKey apiKey : apiKeys) {
             if (!ApiGatewayUtils.deleteApiKey(apiKey)) {
                 LOG.error("Could not delete API key for user {}. Aborting delete user.", apiKey.keyId);
                 return false;
             }
         }
-        boolean auth0UserDeleted = super.delete();
-        if (auth0UserDeleted) {
-            return Persistence.apiUsers.removeById(this.id);
-        } else {
-            LOG.warn("Aborting user deletion for {}", this.email);
-            return false;
+
+        if (deleteAuth0User) {
+            boolean auth0UserDeleted = super.delete();
+            if (!auth0UserDeleted) {
+                LOG.warn("Aborting user deletion for {}", this.email);
+                return false;
+            }
         }
+
+        return Persistence.apiUsers.removeById(this.id);
     }
 
     public void createApiKey(String usagePlanId, boolean persist) throws CreateApiKeyException {
--- a/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
+++ b/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
@@ -56,14 +56,15 @@ public class ApiKeyManagementTest extends OtpMiddlewareTest {
     }
 
     /**
-     * Remove the users {@link AdminUser} and {@link ApiUser} and any remaining API keys. Note: apiUser.delete() cannot
-     * be used because no Auth0 user is created and it fails on attempting to delete this account from Auth0.
+     * Remove the users {@link AdminUser} and {@link ApiUser} and any remaining API keys.
      */
     @AfterAll
     public static void tearDown() {
         assumeTrue(isEndToEndAndAuthIsDisabled());
-        // Delete admin user.
-        Persistence.adminUsers.removeById(adminUser.id);
+        // Note: apiUser.delete() cannot be used because no Auth0 user is created. The method fails on attempting to
+        // delete the random value assigned to the apiUser on initialization. This in turn prevents the apiUser from
+        // being delete from Mongo.
+
         // Refresh api keys for user.
         apiUser = Persistence.apiUsers.getById(apiUser.id);
         // remove remaining api keys if present
@@ -74,6 +75,8 @@ public class ApiKeyManagementTest extends OtpMiddlewareTest {
         }
         // Delete api user.
         Persistence.apiUsers.removeById(apiUser.id);
+        // Delete admin user.
+        Persistence.adminUsers.removeById(adminUser.id);
     }
 
     /**
--- a/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
+++ b/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
@@ -6,7 +6,6 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.opentripplanner.middleware.models.AbstractUser;
 import org.opentripplanner.middleware.models.AdminUser;
-import org.opentripplanner.middleware.models.ApiKey;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.persistence.Persistence;
 import org.opentripplanner.middleware.persistence.PersistenceUtil;
@@ -23,7 +22,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.opentripplanner.middleware.TestUtils.isEndToEnd;
-import static org.opentripplanner.middleware.TestUtils.isEndToEndAndAuthIsDisabled;
 import static org.opentripplanner.middleware.TestUtils.mockAuthenticatedRequest;
 import static org.opentripplanner.middleware.auth.Auth0Connection.getDefaultAuthDisabled;
 import static org.opentripplanner.middleware.auth.Auth0Connection.setAuthDisabled;
@@ -63,21 +61,8 @@ public class ApiKeyManagementTest extends OtpMiddlewareTest {
      */
     @AfterAll
     public static void tearDown() {
-        assumeTrue(isEndToEndAndAuthIsDisabled());
-        // Note: apiUser.delete() cannot be used because no Auth0 user is created. The method fails on attempting to
-        // delete the random value assigned to the apiUser on initialization. This in turn prevents the apiUser from
-        // being delete from Mongo.
-
-        // Refresh api keys for user.
-        apiUser = Persistence.apiUsers.getById(apiUser.id);
-        // remove remaining api keys if present
-        if (!apiUser.apiKeys.isEmpty()) {
-            for (ApiKey apiKey : apiUser.apiKeys) {
-                deleteApiKeyRequest(apiUser.id, apiKey.keyId, adminUser);
-            }
-        }
-        // Delete api user.
-        Persistence.apiUsers.removeById(apiUser.id);
+        setAuthDisabled(getDefaultAuthDisabled());
+        apiUser.delete(false);
         // Delete admin user.
         Persistence.adminUsers.removeById(adminUser.id);
     }
--- a/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
+++ b/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
@@ -62,7 +62,11 @@ public class ApiKeyManagementTest extends OtpMiddlewareTest {
     @AfterAll
     public static void tearDown() {
         setAuthDisabled(getDefaultAuthDisabled());
-        if (apiUser != null) apiUser.delete(false);
+        if (apiUser != null) {
+            // refresh API key(s)
+            apiUser = Persistence.apiUsers.getById(apiUser.id);
+            apiUser.delete(false);
+        }
         if (adminUser != null) Persistence.adminUsers.removeById(adminUser.id);
     }
 
--- a/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
+++ b/src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
@@ -6,6 +6,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.opentripplanner.middleware.models.AbstractUser;
 import org.opentripplanner.middleware.models.AdminUser;
+import org.opentripplanner.middleware.models.ApiKey;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.persistence.Persistence;
 import org.opentripplanner.middleware.persistence.PersistenceUtil;
@@ -55,7 +56,8 @@ public class ApiKeyManagementTest extends OtpMiddlewareTest {
     }
 
     /**
-     * Remove the users {@link AdminUser} and {@link ApiUser} and any remaining API keys.
+     * Remove the users {@link AdminUser} and {@link ApiUser} and any remaining API keys. Note: apiUser.delete() cannot
+     * be used because no Auth0 user is created and it fails on attempting to delete this account from Auth0.
      */
     @AfterAll
     public static void tearDown() {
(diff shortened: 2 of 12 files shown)


evansiroky (approved):

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
```diff
@@ -27,66 +29,74 @@
  * {@link org.opentripplanner.middleware.controllers.api.ApiController} are not themselves tested here.
  */
 public class TripHistoryPersistenceTest extends OtpMiddlewareTest {
+    private static final int limit = 3;
```
binh-dam-ibigroup: These constants should be all caps.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
```diff
@@ -27,66 +29,74 @@
...
+        Persistence.tripSummaries.removeById(tripSummary.id);
+        TripSummary deletedTripSummary = Persistence.tripSummaries.getById(tripSummary.id);
+        assertNull(deletedTripSummary, "Deleted trip summary should no longer exist in database (should return as null).");
     }
 
     @Test
     public void canGetFilteredTripRequestsWithFromAndToDate() {
-        int limit = 3;
         String TRIP_REQUEST_DATE_CREATED_FIELD_NAME = "dateCreated";
         String TRIP_REQUEST_USER_ID_FIELD_NAME = "userId";
```
binh-dam-ibigroup: Move these constants to class-level as they are repeated in other tests.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -64,7 +66,14 @@ public static void tearDown() {
         Persistence.adminUsers.removeById(adminUser.id);
         // Refresh api keys for user.
         apiUser = Persistence.apiUsers.getById(apiUser.id);
-        apiUser.delete();
```
binh-dam-ibigroup, line 70: Does this function not do the same as below? If not, then it would be good to have to have a comment why `apiUser.delete` should not be used.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -64,7 +66,14 @@ public static void tearDown() {
         Persistence.adminUsers.removeById(adminUser.id);
```
binh-dam-ibigroup: Some tests in this class leave AWS API keys behind. I think we need the admin user to still exist in the database in order to delete the keys, so this line should be moved to after the API user is deleted.

binh-dam-ibigroup (changes requested):
Looks good with minor suggestions, but I'm still seeing 3 AWS API keys left behind from `ApiKeyManagementTest`. In that class, I think the admin user should be deleted last.

binh-dam-ibigroup:
Oh and the merge conflicts should be resolved too.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
```diff
@@ -27,66 +29,74 @@
...
+        user = createUser(TEST_EMAIL);
+        tripRequest = createTripRequest(userId);
+        tripSummary = createTripSummary();
+        tripSummaryWithError = createTripSummaryWithError();
+        tripRequests = createTripRequests(limit, user.id);
+    }
+
+    @AfterAll
+    public static void tearDown() {
+        if (user != null) Persistence.otpUsers.removeById(user.id);
```
binh-dam-ibigroup: Could we use `user.delete()` here? Or add a comment if we should not use `user.delete()`.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: You are of course correct! I have updated. Thanks.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
br648: I've added a comment, same issue as apiUser.delete(). Both fail if the Auth0 user is not deleted.

binh-dam-ibigroup: Or, should there be an `apiUser.delete(boolean deleteAuth0User)` to deal with both the tests and the regular code? (EDIT: and likewise for `OtpUser`)

br648: That would make sense, depends how much we want to update the code to accomodate tests, @landonreed thoughts?

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -61,13 +63,23 @@ public static void setUp() throws IOException, InterruptedException {
      */
     @AfterAll
     public static void tearDown() {
-        assumeTrue(isEndToEnd);
-        setAuthDisabled(getDefaultAuthDisabled());
-        // Delete admin user.
-        Persistence.adminUsers.removeById(adminUser.id);
+        assumeTrue(isEndToEndAndAuthIsDisabled());
```
landonreed: Using something like `isEndToEndAndAuthIsDisabled` no longer makes sense here. `authDisabled` is now set manually in this test class' `setUp` method. Also, `setAuthDisabled(getDefaultAuthDisabled())` should be called here.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -61,13 +63,23 @@ public static void setUp() throws IOException, InterruptedException {
...
+        // Note: apiUser.delete() cannot be used because no Auth0 user is created. The method fails on attempting to
+        // delete the random value assigned to the apiUser on initialization. This in turn prevents the apiUser from
+        // being delete from Mongo.
```
landonreed: I think we should just overload the method with a boolean to optionally skip deleting on Auth0 user rather than copy-pasta delete code all over the place.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
```diff
@@ -27,66 +28,77 @@
...
 
-    OtpUser user = null;
-    TripRequest tripRequest = null;
-    TripSummary tripSummary = null;
-    List<TripRequest> tripRequests = null;
+    @AfterAll
+    public static void tearDown() {
+        // Note: otpUser.delete() cannot be used because no Auth0 user is created. The method fails on attempting to
+        // delete the random value assigned to the otpUser on initialization. This in turn prevents the otpUser from
+        // being delete from Mongo.
```
landonreed: Again, this just seems like an opportunity to overload the delete method, rather than duplicate a bunch of code.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -35,19 +35,29 @@
...
- * The following environment variable must be set for these tests to run: - RUN_E2E=true
+ * AUTH0_DOMAIN set to a valid Auth0 domain.
+ *
+ * AUTH0_API_CLIENT set to a valid Auth0 application client id.
+ *
+ * AUTH0_API_SECRET set to a valid Auth0 application client secret.
+ *
+ * DEFAULT_USAGE_PLAN_ID set to a valid usage plan id (AWS requires this to create an api key).
+ *
+ * OTP_API_ROOT set to a live OTP instance (e.g. http://otp-server.example.com/otp).
```
landonreed, line 51: This is no longer accurate... is this the result of a bad merge?

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -80,23 +80,16 @@ public static void tearDown() {
...
-        // Submit a query to the OTP server.
-        // From P&R to Downtown Orlando
-        OtpDispatcherResponse otpDispatcherResponse = OtpDispatcher.sendOtpPlanRequest(
-            "28.45119,-81.36818",
-            "28.54834,-81.37745"
-        );
-        // Construct a monitored trip from it.
-        MonitoredTrip monitoredTrip = new MonitoredTrip(otpDispatcherResponse)
-            .updateAllDaysOfWeek(true);
+        MonitoredTrip monitoredTrip = TestUtils.constructedMonitoredTripFromOtpResponse();
```
landonreed: This method name does not seem right. Also, I find the combination of the OTP plan request and construction of MonitoredTrip into this method not entirely useful. To me it just obfuscates what's happening without really providing much benefit. I think a cleaner refactor would be to encapsulate the test plan request with sample coordinates into its own method (e.g., `TestUtils.sendSamplePlanRequest`) and then just create the monitored trip from the response directly in this test method's body (as was previously done).

landonreed (changes requested):
A couple things could use refactoring for a cleaner tidying in my opinion.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
landonreed: Oh, interesting, I didn't see this before my review, but came to the same conclusion. I think there are cases in the regular code where we may not want to delete Auth0User anyways, so the overloaded delete method is the best course of action in my view.

br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/persistence/TripHistoryPersistenceTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
br648: @landonreed yes I think so. I had updated so the tests would be against a live OTP server and removed the `TestUtils.mockOtpServer()` call from the `setUp()` method (which is now back). This was in response to "TODO: Run a live OTP instance?".

Which is the ideal approach, mock or live OTP server?

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -61,13 +61,10 @@ public static void setUp() throws IOException, InterruptedException {
      */
     @AfterAll
     public static void tearDown() {
-        assumeTrue(isEndToEnd);
         setAuthDisabled(getDefaultAuthDisabled());
+        apiUser.delete(false);
```
binh-dam-ibigroup, line 66: This line is causing a CI test failure: https://travis-ci.org/github/ibi-group/otp-middleware/builds/729933382#L389

br648: @binh-dam-ibigroup ... but works locally (I know!). I can't see how env variables are impacting this, but I can't think of anything else. Any ideas?

br648: @binh-dam-ibigroup I am actually get an error message but I don't think it is related. "Must be an admin to delete an API key", I'll need to update to delete the user as an admin.

br648: @binh-dam-ibigroup Ignore that it is expected as part of the delete tests. I'm having a good day!

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -115,31 +123,26 @@ public static void tearDown() {
      * records. This also includes Auth0 users if auth is enabled.
      */
     @Test
-    public void canSimulateApiUserFlow() throws IOException {
+    public void canSimulateApiUserFlow()  {
```
binh-dam-ibigroup: Extra space

binh-dam-ibigroup (commented):
Changes look good so far. There are 2 API keys created and not deleted by the E2E tests. @br648 are you able to confirm?

binh-dam-ibigroup (changes requested):
These tests in `ApiKeyManagementTest` still create keys without deleting them:
- `canCreateApiKeyForSelf`
- `adminCanCreateApiKeyForApiUser`

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
br648: Fixed.

br648:
@binh-dam-ibigroup you are correct, I can confirm that API keys are remaining. Reviewing...

br648:
@binh-dam-ibigroup thanks for spotted. Had to refresh the api user to get the latest API keys.

binh-dam-ibigroup (commented):
One of the CI tests failed...

br648:
@binh-dam-ibigroup CI is failing because the API key is already subscribed to the usage plan. The API key is generated by the API gateway so I'm not sure how I can influence this? If I'm correct this is a concern.

landonreed:
@br648 and @binh-dam-ibigroup, the pagination PR has a fix for the API key issue seen in the CI failure. I think it has to do with the improper use of the `customerId` field. I have replaced this with a tag instead. See https://github.com/ibi-group/otp-middleware/pull/72/commits/a2146c3ce5561ad986683eb5f3241be52e989b2b

br648:
@landonreed Thanks. I've re-run the test and it passed this time! I'll take the customer id update and apply it to this PR.

binh-dam-ibigroup (approved):
Should be good to go now. Thanks for the changes.

src/main/java/org/opentripplanner/middleware/models/ApiUser.java
```diff
@@ -44,19 +44,26 @@
      */
     @Override
     public boolean delete() {
+        return delete(true);
+    }
+
+    public boolean delete(boolean deleteAuth0User) {
```
landonreed, line 53: javadoc could use an update (this method probably deserves the main javadoc and the one above should describe it as a default shorthand that deletes the Auth0 user).

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -166,16 +166,16 @@ private static String mockOtpPlanResponse(Request request, Response response) th
...
      */
-    public static MonitoredTrip constructedMonitoredTripFromOtpResponse() {
+    public static OtpDispatcherResponse sendSamplePlanRequest() {
         // Submit a query to the OTP server.
         // From P&R to Downtown Orlando
         OtpDispatcherResponse otpDispatcherResponse = OtpDispatcher.sendOtpPlanRequest(
             "28.45119,-81.36818",
             "28.54834,-81.37745"
         );
         // Construct a monitored trip from it.
```
landonreed: Comment is out of date. You could also just remove the otpDispatcherResponse variable and return directly the sendOtpPlanRequest result.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
```diff
@@ -61,13 +61,13 @@ public static void setUp() throws IOException, InterruptedException {
      */
     @AfterAll
     public static void tearDown() {
-        assumeTrue(isEndToEnd);
```
landonreed, line 64: I think tearDown should `assumeTrue(isEndToEnd)`, right? That's how our other test classes are constructed I believe. I think that would allow you to remove the conditional checks for null apiUser and adminUser.

landonreed (approved):
A couple of minor changes needed still, but otherwise looks pretty good.

src/main/java/org/opentripplanner/middleware/models/ApiUser.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/TestUtils.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/ApiKeyManagementTest.java
br648: @landonreed I think so. I have updated, it runs fine locally so will keep an eye on CI.

PR #82, opened 2020-10-02
Refactor api user check

br648:
### Checklist

### Description

First off, this is very much WIP. I was hoping that I wouldn't have to touch many classes, but it's ended up being a few. The initial focus was to satisfy the ApiUserFlowTest unit tests based on authenticating with an API key instead of Auth0. The API key authentication mirrors that of Auth0 and expects an 'x-api-key' header value. I think I have removed any Auth0 stuff that is related to Api users (no Auth0 account creating/deleting), but by no means done exhaustive testing.

If an Otp user created by an Api user attempts to access the otp-middleware it will fail because auth will go down the Auth0 route.

1) Does an API key represent the key id or value? The two appear to be interchangeable at the moment so this needs addressing.
2) I've made the assumption that if an Otp user does not have an application Id then they will have an Auth0 id when it comes to deletion, this might not be correct?

Still need to consider:

For monitored trips, third pary users cannot receive notifications via SMS/email (we need to add an HTTP endpoint whereby API users can check notification status for a monitored trip on behalf of this user class).

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
@@ -14,6 +14,7 @@ import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.models.AbstractUser;
 import org.opentripplanner.middleware.models.ApiUser;
 import org.opentripplanner.middleware.models.OtpUser;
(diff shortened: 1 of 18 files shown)


br648:
Blocked as additional Auth0 authorization is being added to a child branch. The likelihood is that the child branch will supercede this one.

PR #83, opened 2020-10-08
Change in persisting phone number.

### LINKED ISSUE #79: Do not require persisting user in order to verify phone number
## Observed behavior

The phone number validation flow looks a bit awkward as it requires extra /api/secure/user PUT operations in order to get the phone number validated. See https://github.com/ibi-group/otp-middleware/pull/33/files#r483187826. (current phone validation PR https://github.com/opentripplanner/otp-react-redux/pull/224 implements a workaround.)

## Expected behavior

There shouldn't be a need to PUT new user data just to verify a phone number.

binh-dam-ibigroup:
### Checklist

### Description

This PR fixes #79.

Per team discussion, the `/api/secure/user/<id>/verify_sms` GET endpoint (for requesting an SMS) now takes a phone number as path parameter and persists that number as the user's phone number. It also sets `isPhoneNumberVerified = false` ~and `notificationChannel = "sms"` (let me know if we should not set the notification channel)~. A test has been added to check these assignments.

The `/api/secure/user/<id>/verify_sms` POST endpoint now sets `isPhoneNumberVerified = true` ~and `notificationChannel = "sms"`~ upon receiving an "approved" phone verification status.


src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
```diff
@@ -191,4 +194,48 @@ public void canSimulateApiUserFlow() {
         ApiUser deletedApiUser = Persistence.apiUsers.getById(apiUser.id);
         assertNull(deletedApiUser);
     }
+
+    /**
+     * Tests for the chain of API calls to verify a user's phone number.
+     */
+    @Test
+    public void canSimulatePhoneVerificationFlow() {
```
landonreed: I don't think this test really belongs here. This test class was intended to model how a third party API user interacts with otp-middleware. And an API user (eventually) won't be able to create an OtpUser that can register a phone number for SMS notifications.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -125,7 +141,17 @@ public VerificationResult verifyPhoneWithCode(Request req, Response res) {
...
-        // If the check is successful, status will be "approved"
-        return new VerificationResult(check);
+        // If the check is successful, status will be "approved".
+        VerificationResult verificationResult = new VerificationResult(check);
+        if (verificationResult.status.equals("approved")) {
+            // If the check is successful, update the OtpUser
+            // (set isPhoneNumberVerified and notificationChannel).
+            // TODO: Figure out adding a test for this update to the OtpUser record.
+            otpUser.isPhoneNumberVerified = true;
+            otpUser.notificationChannel = "sms";
```
landonreed: I don't think we should be setting channel here. I think that should be independent of this action (especially considering that we intend to allow a "both" option for notification channel).

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,13 +91,24 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
+        String phoneNumber = req.params(PHONE_PARAM);
+        if (StringUtils.isBlank(phoneNumber)) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "A phone number must be provided.");
         }
-        Verification verification = NotificationUtils.sendVerificationText(otpUser.phoneNumber);
+
+        // Update OtpUser before submitting SMS request.
+        otpUser.phoneNumber = phoneNumber;
+        otpUser.isPhoneNumberVerified = false;
+        otpUser.notificationChannel = "sms";
```
landonreed: As mentioned below, I'm not sure if we want to set this automatically (considering the "both" option)?

landonreed (changes requested):
Looks good. Just a couple of comments (location of test and automatically setting notification channel).

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Removed. I was on the fence about that.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Removed.

src/test/java/org/opentripplanner/middleware/ApiUserFlowTest.java
binh-dam-ibigroup: I moved and changed the test in https://github.com/ibi-group/otp-middleware/pull/83/commits/44864ac4f129dd0e1bc766444a85c73776583201.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,23 +91,35 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
+        Persistence.otpUsers.replace(otpUser.id, otpUser);
+/*
+        Verification verification = NotificationUtils.sendVerificationText(phoneNumber);
         if (verification == null) {
             logMessageAndHalt(req, HttpStatus.INTERNAL_SERVER_ERROR_500, "Unknown error sending verification text");
         }
         // Verification result will show "pending" status if verification text is successfully sent.
         return new VerificationResult(verification);
+ */
+        return null;
```
landonreed: I'm not sure what's happening here... returning null doesn't seem too useful in this scenario! I'm guessing this is just a temporary comment that accidentally got checked in?

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,23 +91,35 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
+     * Before sending a SMS request, this endpoint also saves the submitted phone number, marks it as not verified.
      */
     public VerificationResult sendVerificationText(Request req, Response res) {
         OtpUser otpUser = getEntityForId(req, res);
-        if (otpUser.phoneNumber == null) {
-            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, "User must have valid phone number to verify.");
+        // Get phone number from the path param.
+        String phoneNumber = req.params(PHONE_PARAM);
+        if (StringUtils.isBlank(phoneNumber)) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "A phone number must be provided.");
```
landonreed: After further inspection, this should probably validate the phone number against a regex (or some other validator) instead of just checking whether it's blank.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -125,7 +140,16 @@ public VerificationResult verifyPhoneWithCode(Request req, Response res) {
                 "Unknown error encountered while checking SMS verification code"
             );
         }
-        // If the check is successful, status will be "approved"
-        return new VerificationResult(check);
+        // If the check is successful, status will be "approved".
+        VerificationResult verificationResult = new VerificationResult(check);
+        if (verificationResult.status.equals("approved")) {
+            // If the check is successful, update the OtpUser
+            // (set isPhoneNumberVerified and notificationChannel).
```
landonreed: Probably ok to remove the second line of this comment.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -58,20 +60,21 @@ OtpUser preCreateHook(OtpUser user, Request req) {
...
+        LOG.info("Registering path {}/{}.", ROOT_ROUTE, VERIFY_PATH);
 
         // Add the api key route BEFORE the regular CRUD methods
         ApiEndpoint modifiedEndpoint = baseEndpoint
-            .get(path(ROOT_ROUTE + ID_PATH + VERIFY_ROUTE)
+            .get(path(ROOT_ROUTE + String.format(VERIFY_ROUTE_TEMPLATE, ID_PARAM, VERIFY_PATH, PHONE_PARAM))
                     .withDescription("Request an SMS verification to be sent to an OtpUser's phone number.")
-                    .withPathParam().withName(ID_PARAM).withDescription("The id of the OtpUser.").and()
+                    .withPathParam().withName(ID_PARAM).withRequired(true).withDescription("The id of the OtpUser.").and()
+                    .withPathParam().withName(PHONE_PARAM).withRequired(true).withDescription("The phone number to validate, in E.164 format.").and()
```
landonreed: While precise, this isn't super helpful for most people. It would be good to add an example to this description (e.g., `+15555551234`).

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,23 +91,35 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
-            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, "User must have valid phone number to verify.");
+        // Get phone number from the path param.
+        String phoneNumber = req.params(PHONE_PARAM);
+        if (StringUtils.isBlank(phoneNumber)) {
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "A phone number must be provided.");
         }
-        Verification verification = NotificationUtils.sendVerificationText(otpUser.phoneNumber);
+
+        // Update OtpUser before submitting SMS request.
+        otpUser.phoneNumber = phoneNumber;
```
evansiroky: This approach overwrites the existing phone number before it is verified. If we do things this way, it will be impossible to send off alerts to the phone until it is verified. And if the user decides they actually don't want to verify after all and keep their old number, we will have overwritten it and they would have to go back in and reconfirm their existing phone number.

The temporary phone number should be stored in a temporary field until it is verified.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Yes... thanks for catching that.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -125,7 +140,16 @@ public VerificationResult verifyPhoneWithCode(Request req, Response res) {
...
         }
-        // If the check is successful, status will be "approved"
-        return new VerificationResult(check);
+        // If the check is successful, status will be "approved".
+        VerificationResult verificationResult = new VerificationResult(check);
+        if (verificationResult.status.equals("approved")) {
+            // If the check is successful, update the OtpUser
+            // (set isPhoneNumberVerified and notificationChannel).
+            // TODO: Figure out adding a test for this update to the OtpUser record.
+            otpUser.isPhoneNumberVerified = true;
```
evansiroky, line 161: Only at this point should the `otpUser.phoneNumber` be updated.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,23 +91,35 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
      * during user signup (or if a user wishes to change their notification preferences to use a new un-verified phone
      * number).
+     * Before sending a SMS request, this endpoint also saves the submitted phone number, marks it as not verified.
      */
     public VerificationResult sendVerificationText(Request req, Response res) {
         OtpUser otpUser = getEntityForId(req, res);
-        if (otpUser.phoneNumber == null) {
-            logMessageAndHalt(req, HttpStatus.NOT_FOUND_404, "User must have valid phone number to verify.");
+        // Get phone number from the path param.
+        String phoneNumber = req.params(PHONE_PARAM);
```
landonreed, line 105: I'm wondering now for before we send the verification text, if there is a Twilio service we can call to check whether the phone number is already verified by Twilio?

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: See the Revert Number button in https://github.com/opentripplanner/otp-react-redux/pull/224 for part of this concern.

landonreed (changes requested):
Just a few more items I ran across while testing otp-react-redux.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: I think Twilio throws an error if the number is not in E.164 (e.g. doesn't have the +1 prefix for US numbers).

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
evansiroky: I disagree with the approach of having a revert option due to some of the concerns shared above. Additionally, it is more code to maintain than keeping an existing phone number intact until verification and requiring the user to reverify an original number if they want to change it back.

evansiroky (changes requested):
Please refactor so that an existing confirmed phone number does not get overwritten. See https://github.com/ibi-group/otp-middleware/pull/83#discussion_r502023720

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -157,5 +165,77 @@ public static boolean sendEmailViaSendGrid(String toEmail, String subject, Strin
         }
 
     }
+
+    /**
+     * Ensures that the provided phone number is a domestic mobile number.
+     * Returns a 400-Bad request status if that is not the case.
+     */
+    public static com.twilio.rest.lookups.v1.PhoneNumber ensureDomesticPhoneNumber(spark.Request req, String phoneNumberString) {
```
landonreed: This method needs a lot of work before being ready to ship:

- First, I don't think it should take a request as a parameter. Take a look at `ApiGatewayUtils#createApiKey` which throws a `CreateApiKeyException` for a pattern that could be mimicked here.

https://github.com/ibi-group/otp-middleware/blob/692c66afa6e2af09e415154039d870a556ff2260/src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java#L64

- Also, it appears to contain some strange logic for fake numbers, which could cause issues in a production environment.
- It also uses a Twilio service (lookup API) to perform this lookup. I think a simpler (and less costly) approach would just to use a regex as I described in a previous comment.

In summary, I don't think this method needs to exist at all. Please just implement a regex check as mentioned previously: https://stackoverflow.com/a/23299989/915811

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: This item is unresolved.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -88,13 +92,25 @@ protected OtpUser getUserProfile(Auth0UserProfile profile) {
...
+            logMessageAndHalt(req, HttpStatus.BAD_REQUEST_400, "A phone number must be provided.");
         }
-        Verification verification = NotificationUtils.sendVerificationText(otpUser.phoneNumber);
+
+        PhoneNumber validNumber = NotificationUtils.ensureDomesticPhoneNumber(req, phoneNumber);
+
+        // Update OtpUser.pendingPhoneNumber before submitting SMS request.
+        String phoneNumberToSubmit = validNumber.getPhoneNumber().toString();
+        otpUser.pendingPhoneNumber = phoneNumberToSubmit;
+        otpUser.pendingPhoneNumberFormatted = validNumber.getNationalFormat();
```
landonreed: This is redundant information.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -113,19 +130,29 @@ public VerificationResult verifyPhoneWithCode(Request req, Response res) {
...
-        // If the check is successful, status will be "approved"
-        return new VerificationResult(check);
+
+        // If the check is successful, status will be "approved".
+        VerificationResult verificationResult = new VerificationResult(check);
+        if (verificationResult.status.equals("approved")) {
+            // If the check is successful, update the OtpUser's phoneNumber and erase pendingPhoneNumber.
+            // TODO: Figure out adding a test for this update to the OtpUser record.
+            otpUser.phoneNumber = otpUser.pendingPhoneNumber;
+            otpUser.pendingPhoneNumber = null;
```
landonreed: We did not agree to implement this change from storing a single phone number to pending + phone number.

landonreed (changes requested):
We did not agree to make the change to OtpUser to add a `pendingPhoneNumber` field. This also adds what appears to be an overly-complex validation method to NotificationUtils that should likely be removed.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
binh-dam-ibigroup: I see three big advantages for using Twilio's lookup API over a regex, if cost is not too much of a concern:
- No need to configure a regex for each phone number format we encounter (the country code is enough)
- We can accept phone numbers from the UI in any format, and Twilio will convert it into both an E.164 format (+15555550123) to use to send trip monitoring notifications, and a localized format (555) 555-0123 that is sent back to the UI as of writing. That also means we don't have to implement the logic/configurations to strip the extra characters or figure out country prefixes in order to create an E.164 number for notifications.
- We can exclude international numbers from notifications (for US-formatted numbers, that including Canadian and non-US numbers that have the +1 prefix), and potentially landlines.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: This is to not overwrite an existing confirmed number.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Twilio offers [caller ID management for outgoing calls](https://support.twilio.com/hc/en-us/articles/223180048-Adding-a-Verified-Phone-Number-or-Caller-ID-with-Twilio), but that is not quite what you are talking about.

binh-dam-ibigroup: Actually it is, "outgoing" meaning the destination number a message goes out to... Do you want to address it in this PR or another one?

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Fixed. in https://github.com/ibi-group/otp-middleware/pull/83/commits/e1e2cd22c54f645d30c6662dbf1e8e366797c6ce

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: @landonreed and @evansiroky please agree on the above comment. https://github.com/ibi-group/otp-middleware/pull/83/commits/5621a131b26f784d618e0a8cb228ba3e3700c787 does per above, https://github.com/ibi-group/otp-middleware/pull/83/commits/d88766150d215e265639e0525afce51be7c6ec66 does not.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: Sorry, I didn't mean to remove that it should be E.164 (which is nice because it is precise), just that we should add the example (for folks who don't immediately know what that means). To further clarify: we should be picky about the format used for this phone number field. If it doesn't match e.164, we should reject the request.

It looks like you have added a configurable regex for making this format flexible. I disagree with this change. We should just use a hard-coded regex and enforce a single format in requests.

configurations/default/env.yml.tmp
```diff
@@ -17,6 +17,11 @@ AWS_API_SERVER: aws-api-id.execute-api.us-east-1.amazonaws.com
 AWS_API_STAGE: stage-name
 DEFAULT_USAGE_PLAN_ID: usage-plan-id
 
+### Optional phone number prefix (defaults to "+1" [North America])
+# PHONE_COUNTRY_PREFIX: "+1"
+### Optional phone number regexp to enable phone number checking.
+# PHONE_NUMBER_REGEXP: ^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$
```
landonreed: This should be hardcoded and check for E.164 format

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -98,16 +98,19 @@ public VerificationResult sendVerificationText(Request req, Response res) {
...
+                "Phone number is invalid."
+            );
+        }
 
-        // Update OtpUser.pendingPhoneNumber before submitting SMS request.
-        String phoneNumberToSubmit = validNumber.getPhoneNumber().toString();
-        otpUser.pendingPhoneNumber = phoneNumberToSubmit;
-        otpUser.pendingPhoneNumberFormatted = validNumber.getNationalFormat();
+        // Update OtpUser.phoneNumber before submitting SMS request.
+        String phoneNumberToSubmit = NotificationUtils.getRawPhoneNumber(phoneNumber);
```
landonreed: We should expect clients/requests to contain the country code. Your inclination to make this E.164 format was correct, I'm not sure why you deviated from this. I don't think we really need the config value for country code in this case.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: Using `isPhoneNumberVerified` is how we originally designed things and in my opinion it works just fine. If there needs to be further discussion or an alternative approach, that should be handled outside of this PR.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -28,23 +31,10 @@
...
-     */
-    public String pendingPhoneNumber;
-    /**
-     * Locally-formatted pendingPhoneNumber for display.
-     */
-    public String pendingPhoneNumberFormatted;
-
-    /**
-     * Verified phone number for SMS notifications, in +15551234 format (E.164 format, includes country code, no spaces).
+     * Phone number for SMS notifications.
```
landonreed: Old comment about E.164 format is correct and should be reverted. We want a predictable consistent way for clients to interact with this API. And enforcing that format is an important part of that in my view.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -28,23 +31,10 @@
...
+     * Must include country code, e.g., +15551234
      */
     public String phoneNumber;
-    /**
-     * Locally-formatted phoneNumber for display.
-     */
-    public String phoneNumberFormatted;
```
landonreed: Thanks for removing. If we want some special formatting, we should just handle on the client side with an npm lib.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -167,75 +168,26 @@ public static boolean sendEmailViaSendGrid(String toEmail, String subject, Strin
...
-            }
-        } catch (Exception e) {
-            logMessageAndHalt(
-                req,
-                HttpStatus.INTERNAL_SERVER_ERROR_500,
-                "Error validating phone number format."
-            );
-        }
+        // If no regex string/pattern is configured, assume phone number is valid.
+        if (PHONE_PATTERN == null) return true;
```
landonreed: This will go away once we hardcode the regex.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -167,75 +168,26 @@ public static boolean sendEmailViaSendGrid(String toEmail, String subject, Strin
...
+        return m.matches();
+    }
 
-        return null;
+    /**
+     * @return A phone number in +155555555 format (E.164 format) for passing to Twilio
+     * by stripping non-digit characters and prefixing with the configured country code.
+     */
+    public static String getRawPhoneNumber(String formattedNumber) {
+        return PHONE_COUNTRY_PREFIX + formattedNumber.replaceAll("\\D", "");
```
landonreed: I think this can go away once we enforce that clients pass phone numbers in E.164.

landonreed (changes requested):
Thanks for the changes, @binh-dam-ibigroup. I think the only thing left is to enforce E.164 format in requests. I'm not sure if something I said caused you to deviate from how you had originally designed things. If so, I apologize -- I think your instincts were correct!

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        return cases.entrySet();
+    }
+
+    @Test
+    public void getE164Number() {
+        assumeTrue(testsShouldRun());
+        String formattedNumber = "(800) 555-0123";
+        assertEquals("+18005550123", NotificationUtils.getRawPhoneNumber(formattedNumber));
+    }
```
evansiroky: These 2 tests are for methods entirely contained within the NotificationUtils class, so they should be placed in a the NotificationUtilsTest class or it should be created if it doesn't already exist.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
evansiroky: I still disagree for reasons mentioned in company chat, but don't care enough to block progress.

evansiroky (approved):
Contingent on https://github.com/ibi-group/otp-middleware/pull/83#discussion_r505918524 and @landonreed's approval following E.164 regex hardcoding check.

Still disagree about overwritten a confirmed number with an unconfirmed one, but don't care enough to block progress.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: I can make the UI add the country prefix if that simplifies the backend. It is odd to ask regular users to enter a +1 every time.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
binh-dam-ibigroup: I don't think we want to hardcode the regex.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
binh-dam-ibigroup: I moved the check to `OtpUserController` in https://github.com/ibi-group/otp-middleware/pull/83/commits/a9386a6e2f46bc9529e14d51ff9e51106ec87ba6

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/83/commits/a9386a6e2f46bc9529e14d51ff9e51106ec87ba6.

configurations/default/env.yml.tmp
binh-dam-ibigroup: Removed in https://github.com/ibi-group/otp-middleware/pull/83/commits/a9386a6e2f46bc9529e14d51ff9e51106ec87ba6

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
binh-dam-ibigroup: Reverted in https://github.com/ibi-group/otp-middleware/pull/83/commits/74075ba31332186ab2906e45a2b02c3ff7906448

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: Yep, I think it's more appropriate for the UI to supply this.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
landonreed: I was just curious what was out there, no need to address.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -130,11 +140,11 @@ public VerificationResult verifyPhoneWithCode(Request req, Response res) {
         if (code == null) {
             logMessageAndHalt(req, 400, "Missing code from verify request.");
         }
-        if (otpUser.pendingPhoneNumber == null) {
-            logMessageAndHalt(req, 404, "User must have valid phone number for SMS verification.");
+        if (StringUtils.isBlank(otpUser.phoneNumber)) {
```
landonreed: Is there a good reason to use `isBlank` instead of `isPhoneNumberValidE164`?

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -60,69 +61,70 @@ public static void tearDown() {
...
+     * the request is 400-bad request/500-error, and that the user phone number is unchanged.
      */
-    @Test
-    public void smsRequestShouldSetPendingPhoneNumberOnly() {
-        final String PHONE_NUMBER_TO_VERIFY = "+15555550321";
-        final String PHONE_NUMBER_TO_VERIFY_FORMATTED = "(555) 555-0321";
+    @ParameterizedTest
+    @MethodSource("createBadPhoneNumbers")
+    public void invalidNumbersShouldProduceBadRequest(Map.Entry<String, Integer> testCase) {
+        assumeTrue(testsShouldRun());
```
landonreed: I feel like there is a lot of churn happening on this auth disabled stuff, but I'm pretty sure in some PR we started to use the pattern of not checking `DISABLE_AUTH=false` but rather setting it with Auth0Connection#setAuthDisabled and reverting the value to getDefaultAuthDisabled in tearDown. Please follow that pattern here. Also, does e2e actually need to be enabled for these tests to work?

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -60,69 +61,70 @@ public static void tearDown() {
...
-        OtpUser otpUserWithPhone = JsonUtils.getPOJOFromJSON(otpUserWithPhoneResponse.body(), OtpUser.class);
+        OtpUser otpUserWithPhone = JsonUtils.getPOJOFromJSON(otpUserWithPhoneRequest.body(), OtpUser.class);
         assertEquals(INITIAL_PHONE_NUMBER, otpUserWithPhone.phoneNumber);
-        assertEquals(PHONE_NUMBER_TO_VERIFY, otpUserWithPhone.pendingPhoneNumber);
-        assertEquals(PHONE_NUMBER_TO_VERIFY_FORMATTED, otpUserWithPhone.pendingPhoneNumberFormatted);
+        assertTrue(otpUserWithPhone.isPhoneNumberVerified);
+    }
+
+    private static Set<Map.Entry<String, Integer>> createBadPhoneNumbers() {
+        HashMap<String, Integer> cases = new HashMap<>();
```
landonreed: I kind of like the parameterized pattern that uses `Arguments`. That approach allows for the test args to be `String phoneNumber, int statusCode`.

```java
private static Stream<Arguments> createBadPhoneNumbers() {
        return Stream.of(
            Arguments.of("5555555", HttpStatus.BAD_REQUEST_400),
            Arguments.of("+15555550001", HttpStatus.INTERNAL_SERVER_ERROR_500)
        );
    }
```

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -60,69 +61,70 @@ public static void tearDown() {
...
-    private static List<String> createRejectedNumbers() {
-        return List.of(
-            // "Famous" old parisian print shop number (https://fr.wikipedia.org/wiki/Jean_Mineur)
-            "+33142250001",
+    private static Set<Map.Entry<String, Boolean>> createPhoneNumberTestCases() {
+        HashMap<String, Boolean> cases = new HashMap<>();
+        cases.put("+15555550123", true);
+        cases.put("+1 5555550123", false); // no spaces allowed.
+        cases.put("(555) 555,0123", false);
+        cases.put("555555", false);
```
landonreed: Same comment about Arguments.

landonreed (approved):
Just a couple of comments that could be addressed, but otherwise looks great.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: It's evolution!

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
binh-dam-ibigroup: Good point. e2e is not required.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
binh-dam-ibigroup: Good find, it's a cleaner syntax. Updated in https://github.com/ibi-group/otp-middleware/pull/83/commits/8ec525aabdaac519033d9918bca89ad483ecaa7f.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/83/commits/8ec525aabdaac519033d9918bca89ad483ecaa7f.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/83/commits/8ec525aabdaac519033d9918bca89ad483ecaa7f.

PR #97, opened 2020-11-07
Notification thresholds

evansiroky:
### Checklist

### Description

This PR refactors the delay notifications. It does a few things:

- Saves the original scheduled arrival and departure times for the current journey being monitored
- Saves a current baseline time that is used to check if a threshold for delay has been met. This is then updated for subsequent checks for meeting a new delay threshold.
- Refactors the delay notification methods into a single method that reuses some common logic.
- Adds tests for a few use cases


src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -185,21 +188,58 @@ public TripMonitorNotification checkTripForNewAlerts() {
...
+        );
+
+        // check if threshold met
+        if (deviationInMinutes >= trip.departureVarianceMinutesThreshold) {
+            // threshold met, set new baseline time
+            if (delayType == NotificationType.DEPARTURE_DELAY) {
+                journeyState.baselineDepartureTimeEpochMillis = matchingItineraryTargetTime.getTime();
+            } else {
+                journeyState.baselineArrivalTimeEpochMillis = matchingItineraryTargetTime.getTime();
+            }
```
binh-dam-ibigroup: If we are distinguishing between departure and arrival here, is the "target time" concept still necessary as an intermediate?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -290,7 +330,7 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
         journeyState = trip.retrieveJourneyState();
         if (
             journeyState.matchingItinerary == null ||
-                journeyState.matchingItinerary.endTime.before(new Date(journeyState.lastCheckedMillis))
+                journeyState.matchingItinerary.endTime.before(new Date(journeyState.lastCheckedEpochMillis))
```
binh-dam-ibigroup: We don't need the extra indent here.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -336,6 +376,35 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
                 if (!calculateNextItinerary()) return true;
             }
             LOG.info("Next itinerary happening on {}.", targetDate);
+
+            // update journey state with baseline departure and arrival times which are the last known departure/arrival
+            journeyState.baselineDepartureTimeEpochMillis = matchingItinerary.startTime.getTime();
+            journeyState.baselineArrivalTimeEpochMillis = matchingItinerary.endTime.getTime();
+
+            // update journey state with the original (scheduled departure and arrival times). Calculate these by
```
binh-dam-ibigroup: Typo: move `)` before "times"?

binh-dam-ibigroup (approved):
LGTM, with just some minor comments.

br648 (approved):
LGTM

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -82,6 +78,13 @@ public CheckMonitoredTrip(MonitoredTrip trip) {
         this.trip = trip;
     }
 
+    /**
+     * For testing purposes only!
+     */
+    public void setJourneyState(JourneyState journeyState) {
+        this.journeyState = journeyState;
+    }
+
```
landonreed: I'm not sure this is the PR to do this in, but I've been thinking that since we're exposing JourneyState (or some portion of it) in the getMonitoredTrip endpoint, it probably does make sense to just remove the collection and have the JourneyState object as a field on MonitoredTrip.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -185,21 +188,58 @@ public TripMonitorNotification checkTripForNewAlerts() {
...
+            } else {
+                journeyState.baselineArrivalTimeEpochMillis = matchingItineraryTargetTime.getTime();
+            }
+
+            // create and return notification
+            return TripMonitorNotification.createDelayNotification(
+                TimeUnit.MINUTES.convert(
+                    matchingItineraryTargetTime.getTime() - originalTargetTimeEpochMillis,
+                    TimeUnit.MILLISECONDS
+                ),
```
landonreed: Please consider using a variable here. It's not super easy to read like this.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -336,6 +376,35 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+                        leg.arrivalDelay,
+                        TimeUnit.SECONDS
+                    );
+                    break;
+                }
+            }
+
             // save journey state with updated matching itinerary and target date
             journeyState.update(this);
         } else {
```
landonreed: I realize it's not quite in scope, but this else clause feels very naked.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -336,6 +376,35 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+            for (int i = matchingItinerary.legs.size() - 1; i >= 0; i--) {
+                Leg leg = matchingItinerary.legs.get(i);
+                if (leg.transitLeg) {
+                    journeyState.originalArrivalTimeEpochMillis -= TimeUnit.MILLISECONDS.convert(
+                        leg.arrivalDelay,
+                        TimeUnit.SECONDS
+                    );
+                    break;
+                }
+            }
```
landonreed: Feels like there is an opportunity to encapsulate this into a method. Maybe something like `public static long getTransitStartOrEnd(boolean start)`? Having two back-to-back for loops like this kind of breaks the flow when reading.

Also, does this only need to happen once? I'm having trouble understanding why this gets executed each time the trip is checked.

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
```diff
@@ -39,22 +42,43 @@ public static TripMonitorNotification createAlertNotification(
...
+
         notification.body = String.format(
-            "The %s time for your itinerary was delayed by %d minutes (your threshold is currently set to %d minutes).",
+            "The %s time for your itinerary is now %s (%s) (your threshold is currently set to %d minutes).",
             delayType == NotificationType.ARRIVAL_DELAY ? "arrival" : "departure",
-            delayInMinutes,
-            delayThreshold
+            ZonedDateTime
+                .ofInstant(targetDatetime.toInstant(), DateTimeUtils.getOtpZoneId())
+                .format(DateTimeFormatter.ofPattern("HH:mm")),
```
landonreed: This time format should be read from a config value. Perhaps it could be called `NOTIFICATION_TIME_FORMAT`?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -185,21 +188,58 @@ public TripMonitorNotification checkTripForNewAlerts() {
...
 
-    public TripMonitorNotification checkTripForDepartureDelay() {
-        long departureDelayInMinutes = TimeUnit.SECONDS.toMinutes(matchingItinerary.legs.get(0).departureDelay);
-        // First leg departure time should not exceed variance allowed.
-        if (departureDelayInMinutes >= trip.departureVarianceMinutesThreshold) {
-            return TripMonitorNotification.createDelayNotification(departureDelayInMinutes, trip.departureVarianceMinutesThreshold, DEPARTURE_DELAY);
+    /**
+     * Checks whether the trip is beginning or ending at a time greater than the allowable variance. This will check
+     * whether the departure or arrival time of the whole journey has deviated to the point where the absolute value of
+     * the variance has changed more than the trip's variance. If it has, a notification is generated.
```
landonreed: I find this javadoc to be a bit hard to follow. Suggestion:

> Check whether the trip arrives or departs on time. If there is a delay for the trip that exceeds the allowable variance (in either direction -- early or late), a notification is generated.

I'm not sure if that exactly captures what it currently does because it's kind of hard for me to interpret the second sentence 😅

Also, I think it would improve the javadoc by adding some concreteness with an example.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -185,21 +188,58 @@ public TripMonitorNotification checkTripForNewAlerts() {
...
+     */
+    public TripMonitorNotification checkTripForDelay(NotificationType delayType) {
+        // the target time for trip depending on notification type
+        Date matchingItineraryTargetTime;
+
+        // the current baseline epoch millis to check if a new threshold has been met
+        long baselineItineraryTargetEpochMillis;
```
landonreed, line 210: "baseline" feels too close to "original" to me. Maybe "latest"?

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -122,36 +124,130 @@ public void canMonitorTrip() throws URISyntaxException {
...
+    private static List<DelayNotificationTestCase> createDelayNotificationTestCases () throws URISyntaxException {
+        List<DelayNotificationTestCase> testCases = new ArrayList<>();
+
+        // doesn't create departure/arrival notification for on-time trip
+        CheckMonitoredTrip onTimeTrip = createCheckMonitoredTrip();
+        onTimeTrip.setJourneyState(createDefaultJourneyState());
+        testCases.add(new DelayNotificationTestCase(
+            onTimeTrip,
+            NotificationType.DEPARTURE_DELAY,
+            "doesn't create departure notification for on-time trip"
```
landonreed: For all of these test messages, is "shouldn't/should" a better word to use?

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -122,36 +124,130 @@ public void canMonitorTrip() throws URISyntaxException {
...
+        ));
+        testCases.add(new DelayNotificationTestCase(
+            twentyMinutesLateTimeTrip,
+            NotificationType.ARRIVAL_DELAY,
+            "The arrival time for your itinerary is now 09:18 (20 minutes late) (your threshold is currently set to 15 minutes).",
+            "creates arrival notification for 20 minute late trip"
+        ));
 
+        // doesn't create departure notification for 20 minute late trip w/ 15 minute threshold
+        // doesn't create arrival notification for 20 minute late trip w/ 15 minute threshold
```
landonreed: Am I missing something? Why don't we want to create a notification for a 20 minute late trip when the threshold is 15 minutes? Please enhance comments to help the reader.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -431,12 +570,22 @@ public String toString() {
...
-        mockItinerary.endTime = new Date(mockItinerary.endTime.getTime() + offset);
+        offsetItineraryTime(
+            mockItinerary,
+            baseZonedDateTime.toEpochSecond() * 1000 - mockItinerary.startTime.getTime()
+        );
+    }
+
+    /**
+     * Offsets the itinerary's timing by adding the given offset to the overall start/end time and each leg start/end
+     * times.
```
landonreed, line 580: Is this used to inject mock delays? If so, the javadoc should state that.

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
```diff
@@ -39,22 +42,43 @@ public static TripMonitorNotification createAlertNotification(
...
+            delayHumanTime = "about on time";
+        } else if (delayInMinutes > 0) {
+            delayHumanTime = String.format("%d minute%s late", delayInMinutes, delayInMinutes > 1 ? "s" : "");
+        } else {
+            delayHumanTime = String.format("%d minute%s early", delayInMinutes, delayInMinutes < -1 ? "s" : "");
+        }
+
         notification.body = String.format(
-            "The %s time for your itinerary was delayed by %d minutes (your threshold is currently set to %d minutes).",
+            "The %s time for your itinerary is now %s (%s) (your threshold is currently set to %d minutes).",
```
landonreed: I think it might be helpful to include the actual time in this message. Here are some suggestions for how this text could read:

> Your trip is set to arrive on time (9:38am).
> Your trip is set to depart 5 minutes late (8:55am). 
> Your trip is set to arrive 10 minutes late (9:55am).

This might need further wordsmithing though. I'm not sure that we should necessarily be tying the unit tests to these strings at this point. Instead, I would rather have one test verify that the string is rendered correctly, and maybe the rest check that the notification gets sent and what the delay value is.

Also, I don't think we should necessarily be telling the user about their threshold -- I don't think that will make any sense in the context of an SMS message.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -185,21 +188,58 @@ public TripMonitorNotification checkTripForNewAlerts() {
...
-        return null;
-    }
 
-    public TripMonitorNotification checkTripForArrivalDelay() {
-        Leg lastLeg = matchingItinerary.legs.get(matchingItinerary.legs.size() - 1);
-        long arrivalDelayInMinutes = TimeUnit.SECONDS.toMinutes(lastLeg.arrivalDelay);
-        // Last leg arrival time should not exceed variance allowed.
-        if (arrivalDelayInMinutes >= trip.arrivalVarianceMinutesThreshold) {
-            return TripMonitorNotification.createDelayNotification(arrivalDelayInMinutes, trip.arrivalVarianceMinutesThreshold, ARRIVAL_DELAY);
+        // calculate deviation from threshold in minutes
```
landonreed: Is this the deviation from threshold or the deviation from the scheduled arrival/departure?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
landonreed: I find the result of combining the logic for arrival and departure delay checks to ultimately make things harder to read. I think I would ultimately prefer two separate checks rather than a single method that accepts the NotificationType arg. Maybe there's some shared logic that could be used by both methods, but I find the multiple checks in the method and the generic language to be somewhat hard to follow.

landonreed (changes requested):
I have a few comments on variable names and the notification language that's used. Structurally, I also think having two separate methods (that perhaps rely on shared logic) would be a bit more readable than how it currently stands.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I believe this improves readability as this line immediately follows a new operator. We commonly indent after a new logical operator like this in JS, so I think this is ok.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: `latest` doesn't make sense to me, because this isn't always the most recently obtained time. `baseline` on the other has a definition of "a specific value or values that can serve as a comparison or control" which seems very appropriate here.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: No, delays are stored in a separate field within the legs. This changes the overall time and would preserve any existing delay information.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Clarified comment in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
evansiroky: The tests developed already check all these things at once and also for when notifications shouldn't be sent. There might be some value into breaking up the tests into more fine-grained unit tests to be able to test things in isolation, but I'm inclined to keep the test structure as is since it gets a lot of bang-for-the-buck by checking the exact string because in doing so it checks other timing aspects.

I did go ahead and change the message as you recommended in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Added some more comments and changed some timing to note an 18 minute late baseline in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Used "should" / "should not" in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Examples added in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
evansiroky: Fixed in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: This isn't executed each time a trip is checked since it only occurs if there isn't yet a matching itinerary or if the matching itinerary is no longer active or upcoming. I went ahead an created some methods in the Itinerary class in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: It does kinda come out of nowhere. Refactored in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in a8d90ab7a8ea11d1ed9c3c64e90e122e4a8a6d14

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in 6fb9ef205237507ad27eaab1306f5bfa7e9284a4

evansiroky:
Ready for review again

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
landonreed: There's still not a great definition of/distinction between "original" vs. "baseline" in the javadoc. If we're keeping those variable names, I think they need better descriptions both here and in journey state fields. Maybe it makes perfect sense to use these names, but I'm still finding this somewhat hard to follow.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
landonreed: Thanks for the examples, but I'm still finding the main javadoc very difficult to comprehend.

landonreed (changes requested):
Thanks for the changes! I'm still finding the javadoc for checkTripForDelay hard to comprehend. It might also help to add some more clarification there for what baseline vs. original actually mean.

binh-dam-ibigroup (approved):
Approval stands. The changes to JourneyState look good so far.

evansiroky:
@landonreed, I renamed all "original" variables to "scheduled" and added more docs. Lmk if this doesn't clear things up.

landonreed (approved):

PR #105, opened 2020-11-16
Trip monitoring no itinerary match handling

evansiroky:
### Checklist

### Description

This PR refactors checking monitored trips to make less OTP requests and better handle when itineraries aren't found. Most of the changes are in the CheckMonitoredTrip and CheckMonitoredTripTest classes as follows:

### `shouldSkipMonitoredTripCheck` refactor 

The previous code would always make an OTP request when advancing to the next possible date that the trip occurs. This code does away with that and instead updates the matching itinerary directly on the anticipated time of the next journey. Also, there have been some additional checks for whether an itinerary isn't possible either for the current day or any day and if that is the case, the trip check will be skipped.

### `calculateOtpResponse` and `updateMatchingItinerary` refactor

These two methods have been combined together into `makeOTPRequestAndUpdateMatchingItinerary`. Also, some additional code has been added to `makeOTPRequestAndUpdateMatchingItinerary` that will:

- set the current trip status
- advance the matching itinerary to the next possible time once a trip has completed 
- handle cases where no matching itinerary is found a little better. When no match is found a few checks will be made to see if the trip has been checked for matching itineraries on other days of the week and if it looks like there have been no matches on any active days for a week, then the trip will be set to be no longer possible. Also, a notification indicating whether the trip is no longer possible for the current day or any day will be sent.

### `runCheckLogic`  refactor

The previous code took into account whether the `shouldSkipMonitoredTripCheck` had already calculated and saved a matching itinerary. Since `shouldSkipMonitoredTripCheck` doesn't make any OTP requests anymore, the request to OTP will occur only when the trip is within the threshold for checking. As such, the `calculateOtpResponse` 

### `canMonitorTrip` test case

This was updated to use a fixed time and also use the itineraryExistence close to how it'd be used in real life.

### `testSkipMonitoredTripCheck` parameterized test

- The `testSkipMonitoredTripCheck` parameterized test has had nearly all of the preparation calculations moved to the `ShouldSkipTripTestCase#generateCheckMonitoredTrip` method.
- The `ShouldSkipTripTestCase` class has been refactored to have a single constructor. 
- All stuff related to setting up OTP mocks has been removed since the `shouldSkipMonitoredTripCheck` method no longer makes OTP requests
- The `ShouldSkipTripTestCase#lastCheckedTime` field is now public. If it is not set, there is an assumption for the test case that the trip has never been checked before
- The `ShouldSkipTripTestCase#trip` field is now public. If it is not set, there is an assumption for the test case that the next trip occurs on June 6, 2020 at 840am.

### `canMakeOTPRequestAnd*` tests

The `canMakeOTPRequestAndUpdateMatchingItineraryForPreviouslyUnmatchedItinerary`, `canMakeOTPRequestAndResolveUnmatchedItinerary` and `canMakeOTPRequestAndResolveNoLongerPossibleTrip` test cases have all been added that will check whether after executing the `CheckMonitoredTrip.makeOTPRequestAndUpdateMatchingItinerary()` method that the following is done correctly:

- The trip status is set to the correct value
- The itinerary existence has been updated to the correct value
- The trip's `journeyState#noLongerPossible` filed has been set to the correct value


src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
```diff
@@ -71,6 +68,10 @@
      */
     public String targetDate;
 
+    public TripStatus tripStatus;
+
+    public boolean noLongerPossible = false;
```
landonreed: These guys need javadoc.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -139,18 +129,125 @@ private void runCheckLogic() {
...
+
+            // building a new list by copying all values, except for the date which is set to the target date
+            List<NameValuePair> newParams = new ArrayList<>();
+            newParams.add(new BasicNameValuePair("date", targetDate));
+            for (Map.Entry<String, String> param : params.entrySet()) {
+                if (!param.getKey().equals("date")) {
+                    newParams.add(new BasicNameValuePair(param.getKey(), param.getValue()));
+                }
+            }
+            otpDispatcherResponse = OtpDispatcher.sendOtpPlanRequest(URLEncodedUtils.format(newParams, UTF_8));
```
landonreed: This seems very similar to code in `ItineraryUtils#getOtpRequestsForDates`. Things could probably be DRYed out a bit.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -139,18 +129,125 @@ private void runCheckLogic() {
...
+        } catch (Exception e) {
+            // TODO: report bugsnag
+            LOG.error("Encountered an error while making a request ot the OTP server. error={}", e);
```
landonreed: When logging the exception you shouldn't need the formatter notation `{}`.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -139,18 +129,125 @@ private void runCheckLogic() {
...
+        OtpResponse otpResponse = otpDispatcherResponse.getResponse();
         for (int i = 0; i < otpResponse.plan.itineraries.size(); i++) {
             Itinerary candidateItinerary = otpResponse.plan.itineraries.get(i);
-            if (ItineraryUtils.itinerariesMatch(candidateItinerary, trip.itinerary)) {
+            if (ItineraryUtils.itinerariesMatch(trip.itinerary, candidateItinerary)) {
+                // matching itinerary found!
+                LOG.info("Found matching itinerary!");
+
+                // Set the matching itinerary.
                 matchingItinerary = candidateItinerary;
```
landonreed, line 182: might be helpful to add `this` qualifier.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -139,18 +129,125 @@ private void runCheckLogic() {
...
+
+                // set the status according to whether the current itinerary occurs in the past, present or future
+                updateTripStatus();
+
+                // update the trip's itinerary existence data so that any invalid dates are cleared (thus resulting in
+                // that day of week saying that it is a valid day of the week).
+                trip.itineraryExistence.getResultForDayOfWeek(targetZonedDateTime.getDayOfWeek()).invalidDates =
+                    new ArrayList<>();
```
landonreed: This is kind of hard to read. Might be helpful to inject an intermediate variable (best place might be above the start of the for loop block).

```java
ItineraryExistence.ItineraryExistenceResult itinExistenceTargetDay = trip.itineraryExistence
            .getResultForDayOfWeek(targetZonedDateTime.getDayOfWeek());
itinExistenceTargetDay.invalidDates = new ArrayList<>();
```

evansiroky:
This isn't working as expected, so I'm reassigning back to myself.

landonreed (changes requested):
I had started (but not finished) a review. Since you're fixing a few things, here is what I came across so far.

evansiroky:
Ready for review again

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
```diff
@@ -75,17 +67,16 @@
     public JourneyState() {}
 
     /**
-     * Update journey state based on results from {@link CheckMonitoredTrip}.
-     * TODO: This may need some tweaking depending on whether a check was successfully completed or not.
-     *   E.g., should a previous journey state be overwritten by a failed check?
+     * Clone this object.
+     * NOTE: This is used primarily during testing and only clones certain needed items so not all entities are
+     * deep-cloned. Implement this further if additional items should be deep-cloned.
```
landonreed: This is no longer just used in testing, right? It looks like this gets called in the CheckMonitoredTrip constructor now.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -211,6 +211,20 @@ public void checkExistence() {
         }
     }
 
+    /**
+     * Checks whether there is at least one day of the week where the trip is still possible. If there is, then true is
+     * returned.
+     */
+    public boolean isStillPossibleOnAtLeastOneMonitoredDayOfTheWeek(MonitoredTrip trip) {
```
binh-dam-ibigroup: Suggestion: remove `still`

src/main/java/org/opentripplanner/middleware/tripMonitor/TripStatus.java
```diff
@@ -0,0 +1,8 @@
+package org.opentripplanner.middleware.tripMonitor;
+
+public enum TripStatus {
```
binh-dam-ibigroup: I would add JavaDoc describing each of these entries.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -46,29 +48,49 @@
...
-    public OtpResponse otpResponse;
+    public Itinerary matchingItinerary;
+
     /**
      * Tracks the time the notification was sent to the user.
      */
     public long notificationTimestampMillis = -1;
 
+    /**
+     * The journey state that was calculated from the prevous run of this job.
```
landonreed: typo: prevous -> previous

landonreed:
I have a few minor changes (from previous review and from today). I still need to look further at this to better understand what's going on under the hood.

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
```diff
@@ -81,10 +81,17 @@ public static TripMonitorNotification createDelayNotification(
...
+     * Creates a notification that the itinerary was not found on either the current day or any day of the week.
+     */
+    public static TripMonitorNotification createItineraryNotFoundNotification(
+        boolean stillPossibleOnOtherMonitoredDaysOfTheWeek
+    ) {
         TripMonitorNotification notification = new TripMonitorNotification();
         notification.type = NotificationType.ITINERARY_NOT_FOUND;
-        notification.body = "Your itinerary was not found in trip planner results";
+        notification.body = stillPossibleOnOtherMonitoredDaysOfTheWeek
+            ? "Your itinerary was not found in trip planner results for today! Please check realtime conditions and plan a new trip."
```
binh-dam-ibigroup, line 93: I would add a TODO comment for improving this message. The wording looks very complicated for an average user.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -46,29 +48,49 @@
...
+
+    /**
+     * An updated journey state that will be saved to the MonitoredTrip record after this check.
+     */
+    public JourneyState journeyState;
+
     public String targetDate;
 
-    private ZonedDateTime targetZonedDateTime;
+    protected ZonedDateTime targetZonedDateTime;
```
binh-dam-ibigroup: I would just make this field package-visible (there are no subclasses for this job, correct?).

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -451,86 +579,91 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+                newEndTime = newEndTime.plusDays(1);
+            } while (newEndTime.plusDays(1).isBefore(targetZonedDateTime));
+            offsetMillis = newEndTime.toInstant().toEpochMilli() - matchingItinerary.endTime.getTime();
+        } else {
+            // find the first itinerary start time that does exceeds the target zoned date time
+            ZonedDateTime newStartTime = DateTimeUtils.makeOtpZonedDateTime(matchingItinerary.startTime);
+            do {
+                newStartTime = newStartTime.plusDays(1);
+            } while (newStartTime.isBefore(targetZonedDateTime));
+            offsetMillis = newStartTime.toInstant().toEpochMilli() - matchingItinerary.startTime.getTime();
```
binh-dam-ibigroup: Almost duplicate code!

binh-dam-ibigroup (commented):
I did a first read and have a few minor comments. I need to play more with it.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I'm not sure how it'd be helpful here and not all the other places this field is used.

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
evansiroky: I imagine all of our messaging will get updated over time, so I think it's fine to leave as it is without a TODO comment for now.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Changed in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6. I'll admit I don't really know much about the differences between protected vs package-visible.

src/main/java/org/opentripplanner/middleware/tripMonitor/TripStatus.java
evansiroky: Added in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
evansiroky: Changed in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
evansiroky: Removed note about testing in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6. The deep-cloning part of the note is still relevant.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Used `ItineraryUtils#toQueryString` in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
evansiroky: Added in 74d09031e8b0f80d0116218d4cb6cfb4d509d6a6.

evansiroky:
Ready for review again.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -338,90 +332,306 @@ void testSkipMonitoredTripCheck(ShouldSkipTripTestCase testCase) throws Exceptio
...
+
+        // execute makeOTPRequestAndUpdateMatchingItinerary method and verify the expected outcome
+        assertEquals(false, mockCheckMonitoredTrip.makeOTPRequestAndUpdateMatchingItinerary());
+
+        // fetch updated trip from persistence
+        MonitoredTrip updatedTrip = Persistence.monitoredTrips.getById(mockTrip.id);
+
+        // verify that status is active
+        assertEquals(
+            TripStatus.NEXT_TRIP_NOT_POSSIBLE,
```
binh-dam-ibigroup: I think there should also be an assertion to check that a notification is generated for this new trip status.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -338,90 +332,306 @@ void testSkipMonitoredTripCheck(ShouldSkipTripTestCase testCase) throws Exceptio
...
+
+        // execute makeOTPRequestAndUpdateMatchingItinerary method and verify the expected outcome
+        assertEquals(false, mockCheckMonitoredTrip.makeOTPRequestAndUpdateMatchingItinerary());
+
+        // fetch updated trip from persistence
+        MonitoredTrip updatedTrip = Persistence.monitoredTrips.getById(mockTrip.id);
+
+        // verify that status is active
+        assertEquals(
+            TripStatus.NO_LONGER_POSSIBLE,
```
binh-dam-ibigroup: There should be an assertion to check that a notification is generated regarding the new trip status.

binh-dam-ibigroup (changes requested):
I think that the tests where the trip status changes from possible to NO_LONGER_POSSIBLE or NEXT_TRIP_NOT_POSSIBLE should check that a notification is generated (see comments).

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -159,19 +154,11 @@ public boolean makeOTPRequestAndUpdateMatchingItinerary() {
...
-            // building a new list by copying all values, except for the date which is set to the target date
-            List<NameValuePair> newParams = new ArrayList<>();
-            newParams.add(new BasicNameValuePair("date", targetDate));
-            for (Map.Entry<String, String> param : params.entrySet()) {
-                if (!param.getKey().equals("date")) {
-                    newParams.add(new BasicNameValuePair(param.getKey(), param.getValue()));
-                }
-            }
-            otpDispatcherResponse = OtpDispatcher.sendOtpPlanRequest(URLEncodedUtils.format(newParams, UTF_8));
+            params.put("date", targetDate);
```
landonreed: We should probably use `ItineraryUtils#DATE_PARAM` here.

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
landonreed: Is this `noLongerPossible` field any different than just checking that `tripStatus === NO_LONGER_POSSIBLE`? I'm wondering if we need both? Or if we're looking for a quick boolean check, maybe we could just replace this field with a method that handles the tripStatus check?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -332,19 +440,39 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
         // get the configured timezone that OTP is using to parse dates and times
         ZoneId targetZoneId = DateTimeUtils.getOtpZoneId();
 
+        // Make sure journey state indicates that the trip should still possible on at least one day of the week as of
+        // the last time a matching itinerary was attempted to be found. This field in the journey state is used instead
+        // of the trip's itinerary existence data in order to not skip trips that are monitored only one day of the week
+        // and may have not had a matching itinerary on that single monitored day even though it could be possible the
+        // next week.
```
landonreed: Might be simpler to say:

```
        // If trip is no longer possible, no further checking is needed. We should not check itinerary existence data
        // here to avoid incorrectly skipping trips that are monitored on a single day of the week, but which may have
        // not had a matching itinerary on that day for one week (even though the trip is possible the next week).
```

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -332,19 +440,39 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-        boolean matchingItineraryActiveOrUpcoming = latestMatchingItinerary != null &&
-            latestMatchingItinerary.endTime.after(new Date(trip.journeyState.lastCheckedEpochMillis));
+        boolean matchingItineraryActiveOrUpcoming = previousMatchingItinerary != null &&
+            previousMatchingItinerary.endTime.after(new Date(previousJourneyState.lastCheckedEpochMillis));
         if (matchingItineraryActiveOrUpcoming) {
-            matchingItinerary = latestMatchingItinerary;
-            targetDate = trip.journeyState.targetDate;
+            // skip checking the trip the rest of the time that it is active if it had its trip status marked as the
+            // next trip not being possible
+            if (previousJourneyState.tripStatus == TripStatus.NEXT_TRIP_NOT_POSSIBLE) return true;
```
landonreed, line 450: I find it somewhat unclear how this trip status value gets updated if the next trip is not possible. Could that be mentioned here as part of the above comment?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -332,19 +440,39 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+            if (previousMatchingItinerary == null) {
+                // clone the trip's itinerary just in case the code attempts to save the trip (and thus the itinerary)
+                matchingItinerary = trip.itinerary.clone();
+            } else {
+                matchingItinerary = previousMatchingItinerary.clone();
+            }
+
             // calculate target time for the next trip plan request
             // find the next possible day the trip is active by initializing the the appropriate target time. Start by
             // checking today's date at the earliest in case the user has paused trip monitoring for a while
```
landonreed, line 472: On line 479: opportunity to reuse: 
```java
targetZonedDateTime = trip.tripZonedDateTime(DateTimeUtils.nowAsLocalDate());
```
?

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -371,41 +499,29 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-            if (!calculateNextItinerary()) return true;
-
-            // check if the matching itinerary has already ended. This could occur on the very first run of
-            // CheckMonitoredTrip for itineraries that occur today, but has already ended. Up until this point, the
-            // matchingItinerary wasn't calculated, so it wasn't known when the end time was.
-            if (matchingItinerary.endTime.before(DateTimeUtils.nowAsDate())) {
-                // itinerary is done today, advance to the next day and recheck
+            // Check if the CheckMonitoredTrip is being ran for the first time for this trip and if the trip's saved
+            // itinerary has already ended for the day. In that case, advance until the next day.
+            if (previousMatchingItinerary == null && trip.itinerary.endTime.after(DateTimeUtils.nowAsDate())) {
```
landonreed: It might be helpful to add some convenience methods to Itinerary to help with the readability of some of these endTime checks. Instead of

```
trip.itinerary.endTime.after(DateTimeUtils.nowAsDate())
```
We could just have:
```
trip.itinerary.hasEnded()
```

There are a few other places where endTime is checked. And there's at least one place (`updateTripStatus`) where a check like `trip.itinerary.inProgress()` could be useful.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -451,86 +567,91 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-     */
-    private boolean calculateNextItinerary() {
-        // double check that this trip is in fact active and throw an error if it is not
-        if (trip.isInactive()) {
-            throw new RuntimeException("An attempt was made to find the next possible active date of an inactive monitored trip.");
-        }
-
-        // if the trip is not active on the current zoned date time, advance until a day is found when the trip is
-        // active.
+    private void advanceToNextActiveTripDate() throws URISyntaxException {
```
landonreed, line 573: Woops, missed a spot that might benefit from some javadoc. Perhaps just bubbling up some of the inlined comments could be useful.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -451,86 +567,91 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+        for (int i = matchingItinerary.legs.size() - 1; i >= 0; i--) {
+            Leg leg = matchingItinerary.legs.get(i);
+            if (leg.transitLeg) {
+                journeyState.scheduledArrivalTimeEpochMillis -= TimeUnit.MILLISECONDS.convert(
+                    leg.arrivalDelay,
+                    TimeUnit.SECONDS
+                );
+                break;
+            }
         }
```
landonreed: This seems very familiar to `Itinerary#getScheduledEndTimeEpochMillis`. Opportunity for DRYing out?

EDIT: it looks like that method is no longer used in this branch, so maybe this should just be replaced with the use of that method (also getScheduledStartTimeEpochMillis)?

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
```diff
@@ -1,13 +1,8 @@
 package org.opentripplanner.middleware.tripMonitor;
```
landonreed: Regarding a previous PR discussion about package naming, might be a good time to change the package name to all lower case.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -338,90 +332,306 @@ void testSkipMonitoredTripCheck(ShouldSkipTripTestCase testCase) throws Exceptio
...
+
+        // verify that trip is no longer possible overall
+        assertEquals(
+            true,
+            updatedTrip.journeyState.noLongerPossible,
+            "updated Trip should indicate it is no longer possible"
+        );
+    }
+
     private static class DelayNotificationTestCase {
```
landonreed: This file is becoming massive. Might be good to move these TestCase classes into their own files. Also, consider what other utils/non-test methods could be move out of here.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -46,29 +43,49 @@
...
+
+    /**
+     * An updated journey state that will be saved to the MonitoredTrip record after this check.
+     */
+    public JourneyState journeyState;
+
     public String targetDate;
 
-    private ZonedDateTime targetZonedDateTime;
+    ZonedDateTime targetZonedDateTime;
```
landonreed, line 84: Are both `targetDate` and `targetZonedDateTime` necessary? Might be helpful to have some javadoc explaining what they're used for/why the distinction exists.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -451,86 +567,91 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+                newEndTime = newEndTime.plusDays(1);
+            } while (newEndTime.plusDays(1).isBefore(targetZonedDateTime));
+            offsetMillis = newEndTime.toInstant().toEpochMilli() - matchingItinerary.endTime.getTime();
+        } else {
+            // find the first itinerary start time that does exceeds the target zoned date time
+            ZonedDateTime newStartTime = DateTimeUtils.makeOtpZonedDateTime(matchingItinerary.startTime);
+            do {
+                newStartTime = newStartTime.plusDays(1);
+            } while (newStartTime.isBefore(targetZonedDateTime));
+            offsetMillis = newStartTime.toInstant().toEpochMilli() - matchingItinerary.startTime.getTime();
```
landonreed: I think this code block to get the offset could be placed in a new method to improve readability of code.

EDIT: expanding on this... things could be rearranged to look like this (after using the Itinerary methods mentioned below and creating an `Itinerary#offsetTimes` method):

```java
        matchingItinerary.offsetTimes(getNextTripOffsetMillis());

        // update journey state with baseline departure and arrival times which are the last known departure/arrival
        journeyState.baselineDepartureTimeEpochMillis = matchingItinerary.startTime.getTime();
        journeyState.baselineArrivalTimeEpochMillis = matchingItinerary.endTime.getTime();

        // update journey state with the original (scheduled departure and arrival times).
        journeyState.scheduledDepartureTimeEpochMillis = matchingItinerary.getScheduledStartTimeEpochMillis();
        journeyState.scheduledArrivalTimeEpochMillis = matchingItinerary.getScheduledEndTimeEpochMillis();
```

landonreed (changes requested):
A few more comments to improve readability/flow of the code.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: I agree that the `Itinerary#offsetTimes` method is useful and implemented it in 1987213b6754ffa98f4748820b628be64f9443f4. However, creating a new method for `getNextTripOffsetMillis` doesn't seem quite worth it as that logic is quite central to the method it is already in, so I think it'd be better for it to stay in this method.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Huh, yeah, `targetDate` isn't necessary as a field in this class. Removed in 1987213b6754ffa98f4748820b628be64f9443f4.

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Moved some stuff around in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
evansiroky: Renamed in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Not sure why I didn't use that. Fixed in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Added in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Refactored in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Updated in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Updated in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Updated in 1987213b6754ffa98f4748820b628be64f9443f4

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in 1987213b6754ffa98f4748820b628be64f9443f4

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Fixed in 1987213b6754ffa98f4748820b628be64f9443f4

src/test/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Good call. In doing so I discovered that the improper notification body was being generated! See 1987213b6754ffa98f4748820b628be64f9443f4.

src/main/java/org/opentripplanner/middleware/tripMonitor/JourneyState.java
evansiroky: Huh, you're right, the trip status can be used instead. I removed this field in 1987213b6754ffa98f4748820b628be64f9443f4.

evansiroky:
Ready for review again.

src/main/java/org/opentripplanner/middleware/tripmonitor/JourneyState.java
```diff
@@ -60,31 +55,30 @@
...
+    public TripStatus tripStatus;
+
     public JourneyState() {}
 
     /**
-     * Update journey state based on results from {@link CheckMonitoredTrip}.
-     * TODO: This may need some tweaking depending on whether a check was successfully completed or not.
-     *   E.g., should a previous journey state be overwritten by a failed check?
+     * Clone this object.
+     * NOTE: This is only clones certain needed items so not all entities are deep-cloned. Implement this further if
```
br648: Remove 'is'.

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -0,0 +1,600 @@
...
+                .withHour(8)
+                .withMinute(45)
+        );
+
+        // Next, run a monitor trip check from the new monitored trip using the simulated response.
+        CheckMonitoredTrip checkMonitoredTrip = new CheckMonitoredTrip(monitoredTrip);
+        checkMonitoredTrip.run();
+        // Assert that there is one notification generated during check.
+        // TODO: Improve assertions to use snapshots.
+        Assertions.assertEquals(1, checkMonitoredTrip.notifications.size());
```
br648, line 129: canMonitorTrip() test fails:

org.opentest4j.AssertionFailedError: 
Expected :1
Actual   :3

br648 (commented):
One minor item, plus one of the unit tests fails for me.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -191,10 +193,7 @@ public boolean makeOTPRequestAndUpdateMatchingItinerary() {
                 // If the updated trip status is upcoming and the end time of the current matching itinerary is in the
                 // past, this means the trip has completed and the next possible time the trip occurs should be
                 // calculated
-                if (
-                    journeyState.tripStatus == TripStatus.TRIP_UPCOMING &&
-                        matchingItinerary.endTime.before(DateTimeUtils.nowAsDate())
-                ) {
+                if (journeyState.tripStatus == TripStatus.TRIP_UPCOMING && !matchingItinerary.hasEnded()) {
                     LOG.info("Matching Itinerary has concluded, advancing to next possible trip date.");
```
binh-dam-ibigroup: The if condition and log message contradict each other.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -246,4 +246,31 @@ public long getScheduledEndTimeEpochMillis() {
         }
         return endTimeEpochMillis;
     }
+
+    /**
+     * Returns true if the current time falls between the start and end time of the itinerary
+     */
+    public boolean isActive() {
```
landonreed, line 255: We might need the Json/Bson Ignore annotations here. (I don't think we need them on `hasEnded`.)

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -191,10 +193,7 @@ public boolean makeOTPRequestAndUpdateMatchingItinerary() {
                 // If the updated trip status is upcoming and the end time of the current matching itinerary is in the
                 // past, this means the trip has completed and the next possible time the trip occurs should be
                 // calculated
-                if (
-                    journeyState.tripStatus == TripStatus.TRIP_UPCOMING &&
-                        matchingItinerary.endTime.before(DateTimeUtils.nowAsDate())
-                ) {
+                if (journeyState.tripStatus == TripStatus.TRIP_UPCOMING && !matchingItinerary.hasEnded()) {
```
landonreed: Shouldn't this just be `matchingItinerary.hasEnded()` (without the "not").

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -440,23 +435,27 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-        // and may have not had a matching itinerary on that single monitored day even though it could be possible the
-        // next week.
-        if (previousJourneyState.noLongerPossible) return true;
+        // If trip is no longer possible, no further checking is needed. The itinerary existence data should not be
+        // checked here to avoid incorrectly skipping trips that are monitored on a single day of the week, but which
+        // may have not had a matching itinerary on that day for one week (even though the trip could be possible the
+        // next week).
+        if (
+            previousJourneyState.tripStatus != null &&
+                previousJourneyState.tripStatus.equals(TripStatus.NO_LONGER_POSSIBLE)
```
binh-dam-ibigroup: `previousJourneyState.tripStatus == TripStatus.NO_LONGER_POSSIBLE`? (enum value equality)

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -440,23 +435,27 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-        // of the trip's itinerary existence data in order to not skip trips that are monitored only one day of the week
-        // and may have not had a matching itinerary on that single monitored day even though it could be possible the
-        // next week.
-        if (previousJourneyState.noLongerPossible) return true;
+        // If trip is no longer possible, no further checking is needed. The itinerary existence data should not be
+        // checked here to avoid incorrectly skipping trips that are monitored on a single day of the week, but which
+        // may have not had a matching itinerary on that day for one week (even though the trip could be possible the
+        // next week).
+        if (
+            previousJourneyState.tripStatus != null &&
```
landonreed: Is the null check necessary? I think we can just use `==` for better readability.

src/test/java/org/opentripplanner/middleware/testutils/OtpTestUtils.java
```diff
@@ -34,6 +36,33 @@
...
+     * - toPlace: Uncharted Realities, SW 3rd Ave, Downtown - Portland 97204::45.51639151281627,-122.67681483620306
+     * - first itinerary end time: 8:58:44am
+     */
+    public static final OtpDispatcherResponse OTP_DISPATCHER_PLAN_RESPONSE =
+        initializeMockPlanResponse("otp/response/planResponse.json");
+
+    /** Contains an OTP response with no itinerary found. */
+    public static final OtpDispatcherResponse OTP_DISPATCHER_PLAN_ERROR_RESPONSE =
+        initializeMockPlanResponse("otp/response/planErrorResponse.json");
+
```
binh-dam-ibigroup, line 65: +1

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/ShouldSkipTripTestCase.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public MonitoredTrip trip;
+
+    ShouldSkipTripTestCase(String message, ZonedDateTime mockTime, boolean shouldSkipTrip) {
+        this.message = message;
+        this.mockTime = mockTime;
+        this.shouldSkipTrip = shouldSkipTrip;
+    }
+
+    @Override public String toString() {
```
binh-dam-ibigroup: Don't we normally place annotations on the preceding line?

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
binh-dam-ibigroup: I see that too. The test doesn't run in CI, so the test failure it doesn't show there. I think you may want to clear all alerts in the simulated itinerary before adding the fake one.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -600,40 +601,16 @@ private void advanceToNextActiveTripDate() throws URISyntaxException {
         }
```
landonreed, line 612: If we're not moving this to its own method, I think it could still use some work. Mainly I would like to see more comments peppering the while checks just to give more context to why there is a difference with the `plusDays` -- that's not immediately clear from the code. Also, the large comment block above could maybe use an edit for a simplified description of what's happening.

landonreed (changes requested):
Just a couple more items. I, too, had some issues running tests locally. Also, I noticed that `OtpMiddlewareTest#canRegisterRedirect` is now run on every test class that extends middleware test -- I forget exactly where this change was made, but we should fix that (doesn't necessarily need to be here if you deem it out of scope).

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -371,41 +495,29 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
-            if (!calculateNextItinerary()) return true;
-
-            // check if the matching itinerary has already ended. This could occur on the very first run of
-            // CheckMonitoredTrip for itineraries that occur today, but has already ended. Up until this point, the
-            // matchingItinerary wasn't calculated, so it wasn't known when the end time was.
-            if (matchingItinerary.endTime.before(DateTimeUtils.nowAsDate())) {
-                // itinerary is done today, advance to the next day and recheck
+            // Check if the CheckMonitoredTrip is being ran for the first time for this trip and if the trip's saved
+            // itinerary has already ended for the day. In that case, advance until the next day.
+            if (previousMatchingItinerary == null && trip.itinerary.endTime.after(DateTimeUtils.nowAsDate())) {
```
binh-dam-ibigroup: The `endTime` comparison is incorrect. Use instead: `trip.itinerary.hasEnded()`.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -452,85 +564,71 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+            ZonedDateTime newEndTime = DateTimeUtils.makeOtpZonedDateTime(matchingItinerary.endTime);
+            do {
+                newEndTime = newEndTime.plusDays(1);
+            } while (newEndTime.plusDays(1).isBefore(targetZonedDateTime));
+            offsetMillis = newEndTime.toInstant().toEpochMilli() - matchingItinerary.endTime.getTime();
+        } else {
+            // find the first itinerary start time that does exceeds the target zoned date time
+            ZonedDateTime newStartTime = DateTimeUtils.makeOtpZonedDateTime(matchingItinerary.startTime);
+            do {
+                newStartTime = newStartTime.plusDays(1);
```
binh-dam-ibigroup, line 609: I tried the following on FDOT and here is what I observed:
- Plan a new rail trip "leaving now" (rail trips have a mock alert set up)
- Find and save an itinerary that leaves within an hour.
- This line gets executed.

So does that mean I will not get the notifications for today (for the trip I am interested in), and will only start getting notifications tomorrow?

binh-dam-ibigroup (changes requested):
There are still issues in the trip checking logic as pointed by @landonreed and me. Also please advise regarding https://github.com/ibi-group/otp-middleware/pull/105/files#diff-19407efb25515c956e1de5427204c08d0c8b92744881a9eeb2c4a29561a0b17aR598, this is preventing notifications for new saved trips departing soon.

The refactoring of the tests are very nice though.

evansiroky:
> Just a couple more items. I, too, had some issues running tests locally. Also, I noticed that `OtpMiddlewareTest#canRegisterRedirect` is now run on every test class that extends middleware test -- I forget exactly where this change was made, but we should fix that (doesn't necessarily need to be here if you deem it out of scope).

Fixed in 4922ec4b7c782fae709aecafdaaff77d08bce88d.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: That's a bug. Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/ShouldSkipTripTestCase.java
evansiroky: Reformatted in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Code was wrong. Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Comments refactored in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Refactored as suggested in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: Yep, fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
evansiroky: Added in d95910c71758df3212e5f878a91a509f02839dd0.

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
evansiroky: Bug introduced in 1987213b6754ffa98f4748820b628be64f9443f4. Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

src/main/java/org/opentripplanner/middleware/tripmonitor/JourneyState.java
evansiroky: Fixed in d95910c71758df3212e5f878a91a509f02839dd0.

evansiroky:
Ready for review again.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -452,85 +565,81 @@ public boolean shouldSkipMonitoredTripCheck() throws Exception {
...
+            // zoned date time.
+            //
+            // Example: The new target time is June 15 at 9am and the previous matching itinerary ended on June 13 at
+            // 8:50am. In this case, the matching itinerary will be incremented two days so the updated matching
+            // itinerary ends at 8:50am on June 15.
+            ZonedDateTime newEndTime = DateTimeUtils.makeOtpZonedDateTime(
+                new Date(matchingItinerary.getScheduledEndTimeEpochMillis())
+            );
+            do {
+                newEndTime = newEndTime.plusDays(1);
```
binh-dam-ibigroup, line 595: Same thing as for the "leave now" trip where notifications are not sent.
To reproduce with FDOT:
- Run https://github.com/opentripplanner/otp-react-redux/pull/289
- Plan and save a trip on SunRail using an "arrive by" time today
- The log will say an itinerary happens today.
- This line will execute and the monitoring will be set for the following day.

binh-dam-ibigroup (commented):
Sorry, one more change for notifications not sent on "arrive by" trips planned for today, (see https://github.com/ibi-group/otp-middleware/pull/105/files#r544431570)

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: I believe 0a834f09c6880b1a8d3d57e4870ef7bbf43be0a8 has fixed this.

evansiroky:
Ready for hopefully one last review.

binh-dam-ibigroup (approved):
Let's approve this if the arrive-by-later-today trip generates a notification (SunRail stopped for the day before I got to this review.)

evansiroky:
I hacked together the ability to plan and monitor an arrive-by trip in advance in [this branch](https://github.com/opentripplanner/otp-react-redux/tree/trip-status-ui-with-hacked-timing-form). It seems to work for the day of.

evansiroky:
Further refactoring of the underlying otp-middleware server start for tests is now in #126.

br648 (approved):
LGTM. Good work!

landonreed (approved):

PR #91, opened 2020-10-30
Only persist trip with transit and no-rental itinerary

binh-dam-ibigroup:
### Checklist

- [X] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [X] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR adds a check to `MonitoredTripController` so that only trips with transit and no rentals can be saved/monitored.

To test in OTP-RR, use any branch except https://github.com/opentripplanner/otp-react-redux/pull/266.

PR #90, opened 2020-10-28
Itinerary matching MVP

evansiroky:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR refactors the checks for itinerary matches.

PR #109, opened 2020-11-18
Merge monitorability checks from ItineraryUtils and MonitoredTripController

binh-dam-ibigroup:
### Checklist

### Description

This PR merges the checks regarding whether an itinerary can be monitored that were previously implemented in `MonitoredTripController` (#91) and simply stubbed in `ItineraryUtils` (#90) but not merged.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
@@ -172,7 +172,7 @@ public class MonitoredTripController extends ApiController<MonitoredTrip> {
             logMessageAndHalt(
                 request,
                 HttpStatus.BAD_REQUEST_400,
-                String.format("Trip cannot be monitored due to: %s", reasonsString)
+                String.format("The trip cannot be monitored: %s", reasonsString)
             );
         }
     }
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
@@ -165,7 +165,7 @@ public class MonitoredTripController extends ApiController<MonitoredTrip> {
      */
     private void checkTripCanBeMonitored(MonitoredTrip trip, Request request) {
         Set<InvalidItineraryReason> invalidReasons = trip.itinerary.checkItineraryIsMonitorable();
-        if (invalidReasons.size() > 0) {
+        if (!invalidReasons.isEmpty()) {
             String reasonsString = invalidReasons.stream()
                 .map(InvalidItineraryReason::getMessage)
                 .collect(Collectors.joining(", "));
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
@@ -164,7 +164,7 @@ public class MonitoredTripController extends ApiController<MonitoredTrip> {
      * {@link org.opentripplanner.middleware.otp.response.Itinerary} can be monitored).
      */
     private void checkTripCanBeMonitored(MonitoredTrip trip, Request request) {
-        Set<InvalidItineraryReason> invalidReasons = trip.itinerary.checkItineraryIsMonitorable();
+        Set<InvalidItineraryReason> invalidReasons = trip.itinerary.checkItineraryCanBeMonitored();
         if (!invalidReasons.isEmpty()) {
             String reasonsString = invalidReasons.stream()
                 .map(InvalidItineraryReason::getMessage)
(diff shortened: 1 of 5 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -161,14 +162,12 @@ private void verifyBelowMaxNumTrips(String userId, Request request) {
...
-
-        if (!hasTransit || hasRentalOrRideHail) {
+        Itinerary.ItineraryCanBeMonitored canBeMonitored = trip.itinerary.assessCanBeMonitored();
+        if (!canBeMonitored.overall) {
             String rejectReason = "";
-            if (!hasTransit) rejectReason += "it does not include a transit leg";
-            if (!hasTransit && hasRentalOrRideHail) rejectReason += ", and ";
-            if (hasRentalOrRideHail) rejectReason += "it includes a rental or ride hail";
+            if (!canBeMonitored.hasTransit) rejectReason += "it does not include a transit leg";
+            if (!canBeMonitored.overall) rejectReason += ", and ";
```
evansiroky: This doesn't seem future-proofed in case another rejection reason is added later. It might make sense to make this a List of checks that are then joined together with the ", and " string.

evansiroky (changes requested):
Please address https://github.com/ibi-group/otp-middleware/pull/109#discussion_r527090958. Also, it would be nice to have a unit test for checking the resulting failure messages.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: Addressed, with unit tests for the message, in https://github.com/ibi-group/otp-middleware/pull/109/commits/e85ca83ef2591b8f2834fc3e46a34c3f395803e5.

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -166,4 +181,41 @@ public Itinerary clone() throws CloneNotSupportedException {
         }
         return cloned;
     }
+
+    /**
+     * A class that holds results for method {@link #assessCanBeMonitored()},
+     * including outcomes for the checks regarding whether the itinerary can be monitored.
+     */
+    public class ItineraryCanBeMonitored {
+        public final boolean hasTransit;
```
landonreed: I think it might help the clarity of this class to change `hasTransit` to `missingTransit`. So that a reader can expect any `true` value to be "bad".

src/main/java/org/opentripplanner/middleware/otp/response/Itinerary.java
```diff
@@ -166,4 +181,41 @@ public Itinerary clone() throws CloneNotSupportedException {
...
+        public final boolean hasRentalOrRideHail;
+        // TODO: add other reasons for why a trip cannot be monitored.
+
+        /**
+         * True if the itinerary has transit and no rental/ride hail legs.
+         */
+        public final boolean overall;
```
landonreed: I think a better name for this might be `result`.

landonreed (changes requested):
Looks pretty good, thanks for taking this on. I originally had a couple of comments (now outdated) but ultimately decided to offer changes in a PR: https://github.com/ibi-group/otp-middleware/pull/113

I'm not too keen on tying a lot of tests to string messages, especially messages that aren't actually used in the code (for example, the `This trip can be monitored.` message). In general, I like the approach, but it seemed like the construction of messages would be better served by an enum.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -158,22 +161,18 @@ private void verifyBelowMaxNumTrips(String userId, Request request) {
...
-        boolean hasRentalOrRideHail = trip.itinerary.hasRentalOrRideHail();
-
-        if (!hasTransit || hasRentalOrRideHail) {
-            String rejectReason = "";
-            if (!hasTransit) rejectReason += "it does not include a transit leg";
-            if (!hasTransit && hasRentalOrRideHail) rejectReason += ", and ";
-            if (hasRentalOrRideHail) rejectReason += "it includes a rental or ride hail";
-
+        Set<InvalidItineraryReason> invalidReasons = trip.itinerary.checkItineraryIsMonitorable();
+        if (invalidReasons.size() > 0) {
```
br648: Could be replaced with !invalidReasons.isEmpty()

br648 (approved):
LTGM

landonreed (approved):

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: Replaced in https://github.com/ibi-group/otp-middleware/pull/109/commits/c08f561d0b78149a3203db4dc9ae25744dfa36bc.

evansiroky (approved):

PR #111, opened 2020-11-20
Error notifications

landonreed:
### Checklist

### Description

This PR does a few big things:
- Refactors Bugsnag to no longer poll for all the projects (e.g., datatools-server-prod, commtrans-middleware, etc.) in a Bugsnag org. It now uses a config property `MONITORED_COMPONENTS` to list specific components to monitor (and it requires that you fill in the `bugsnagProjectId`.
- DRYs out/refactors some of the Bugsnag jobs code (NOTE: I struggled to have git recognize the file name change for BugsnagEventJob.java -> BugsnagEventHandlingJob.java, but they are in fact the same file)
- Adds email notifications when new errors are reported from Bugsnag
- Adds context to notifications based on user type (OtpUsers get notifications from "The Trip Planner" AdminUsers from "The Admin Dashboard")
- Adds email subscriptions to AdminUsers (works with changes in https://github.com/ibi-group/otp-admin-ui/pull/29)

Main config changes are:
```yml
# IMPORTANT: this value changes from org name to org id (uuid)
BUGSNAG_ORGANIZATION: abcd123456

# Add components to monitor in middleware
# TODO: Add other services as needed (e.g., cloudwatch, pingdom).
MONITORED_COMPONENTS:
  - bugsnagProjectId: abcd1234
    name: datatools-server
  - bugsnagProjectId: abcd12345
    name: datatools-ui
  - bugsnagProjectId: abcd123456
    name: otp-middleware

# Config settings for linking to the OTP Admin Dashboard.
OTP_ADMIN_DASHBOARD_EMAIL: OTP Admin Dashboard <no-reply@email.com>
OTP_ADMIN_DASHBOARD_NAME: OTP Admin Dashboard
OTP_ADMIN_DASHBOARD_URL: https://otp-admin-ui.now.sh

# Config settings for linking to the OTP UI (trip planner).
OTP_UI_NAME: FDOT Trip Planner
OTP_UI_URL: https://plan.example.com
```


configurations/default/env.yml.tmp
```diff
@@ -33,8 +33,27 @@ OTP_PLAN_ENDPOINT: /plan
 OTP_TIMEZONE: America/Los_Angeles
 MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
 
+# Add components to monitor in middleware
+# TODO: Add other services as needed (e.g., cloudwatch, pingdom).
+MONITORED_COMPONENTS:
+  - bugsnagProjectId: abcd1234
```
binh-dam-ibigroup, line 38: If other services will be added, would it make sense to have items look as follows:
```yml
MONITORED_COMPONENTS
  - type: bugsnagProject
    id: <projectId>
    name: datatools-ui
```

configurations/default/env.yml.tmp
```diff
@@ -33,8 +33,27 @@ OTP_PLAN_ENDPOINT: /plan
...
+    name: datatools-server
+  - bugsnagProjectId: abcd12345
+    name: datatools-ui
+  - bugsnagProjectId: abcd123456
+    name: otp-middleware
+
+# Config settings for linking to the OTP Admin Dashboard.
+OTP_ADMIN_DASHBOARD_EMAIL: OTP Admin Dashboard <no-reply@email.com>
```
binh-dam-ibigroup: I think the nested structure you started with `MONITORED_COMPONENTS` is a good idea... how about an `OTP_ADMIN_DASHBOARD` header?

src/main/java/org/opentripplanner/middleware/bugsnag/EventSummary.java
```diff
@@ -32,15 +33,10 @@
...
-    /** The dashboard URL for the project */
-    public String htmlUrl = "Unknown";
-
-
-    public EventSummary(BugsnagProject project, BugsnagEvent bugsnagEvent) {
-        if (project != null) {
-            this.projectName = project.name;
-            this.projectId = project.id;
-            this.htmlUrl = project.htmlUrl;
+    public EventSummary(MonitoredComponent component, BugsnagEvent bugsnagEvent) {
```
binh-dam-ibigroup, line 35: If `MonitoredComponent` will not necessarily be a Bugsnag project, should BugsnagEvent be included in this constructor? (This is not a blocking issue right now)

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventRequestJob.java
```diff
@@ -3,32 +3,31 @@
...
+import org.slf4j.LoggerFactory;
 
 /**
- * This job is responsible for making event data requests to Bugsnag and storing the response.
+ * This job is responsible for triggering event data requests to Bugsnag and storing the response. These requests happen
+ * asynchronously and must be monitored over time, which is the task of {@link BugsnagEventHandlingJob}.
  *
- * Bugsnag only allows ten API requests per minute (with the current pricing plan). This places too greater limit on
- * confidently being able to extract error and event information via the organization, project, error and event
+ * This architecture is informed by Bugsnag's limit of ten API requests per minute (with the current pricing plan). This
```
binh-dam-ibigroup: Replace "(with the current pricing plan)" with "on some pricing plans"

src/main/java/org/opentripplanner/middleware/controllers/api/ErrorEventsController.java
```diff
@@ -72,18 +68,18 @@ public void bind(final SparkSwagger restApi) {
         int limit = HttpUtils.getQueryParamFromRequest(req, LIMIT_PARAM, 0, DEFAULT_LIMIT, 100);
         int offset = HttpUtils.getQueryParamFromRequest(req, OFFSET_PARAM, 0, 0);
         // Get latest events from database.
-        FindIterable<BugsnagEvent> events = bugsnagEvents.getSortedIterableWithOffsetAndLimit(
+        FindIterable<BugsnagEvent> events = Persistence.bugsnagEvents.getSortedIterableWithOffsetAndLimit(
```
binh-dam-ibigroup, line 69: This is still tied to Bugsnag, so we probably need a TODO comment to remove the Bugsnag references in this class.

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
```diff
@@ -1,5 +1,6 @@
 package org.opentripplanner.middleware.models;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
```
binh-dam-ibigroup: Why this addition?

src/main/java/org/opentripplanner/middleware/models/MonitoredComponent.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class MonitoredComponent extends Model {
+    private static final Logger LOG = LoggerFactory.getLogger(MonitoredComponent.class);
+    /**
+     * The field that maps this component to a project in Bugsnag (this is a string UUID).
+     *
+     * See https://api.bugsnag.com/organizations/<organization_id>/projects for a list of projects.
+     */
+    public String bugsnagProjectId;
```
binh-dam-ibigroup, line 32: See my comment in env.yml for a suggestion to remove the mention of Bugsnag.

src/main/java/org/opentripplanner/middleware/models/MonitoredComponent.java
```diff
@@ -0,0 +1,88 @@
...
+    public static void initializeMonitoredComponentsFromConfig() {
+        List<MonitoredComponent> configComponents;
+        try {
+            configComponents = JsonUtils.getPOJOFromJSONAsList(
+                ConfigUtils.getConfigProperty("MONITORED_COMPONENTS"),
+                MonitoredComponent.class
+            );
+            if (configComponents == null) {
+                throw new IllegalArgumentException("MONITORED_COMPONENTS should not be missing.");
+            }
```
binh-dam-ibigroup, line 50: `MONITORED_COMPONENTS` should be optional IMO and could default to an empty list.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -38,50 +45,37 @@
...
-    public static HttpResponse<String> httpRequestRawResponse(URI uri, int connectionTimeout, REQUEST_METHOD method,
+    public static HttpResponse<String> httpRequestRawResponse(URI uri, int connectionTimeout, HttpMethod method,
                                                               Map<String, String> headers, String bodyContent) {
-
-
         HttpClient client = HttpClient.newHttpClient();
         HttpRequest.Builder httpRequestBuilder = HttpRequest.newBuilder()
             .uri(uri)
             .timeout(Duration.ofSeconds(connectionTimeout))
             .GET();
```
binh-dam-ibigroup: This `GET` is not necessary according to the docs.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -38,50 +45,37 @@
...
-
-        if (method.equals(REQUEST_METHOD.GET)) {
+        if (method.equals(HttpMethod.GET)) {
```
binh-dam-ibigroup: Use `if ... else if ...` or `switch`.

configurations/default/env.yml.tmp
binh-dam-ibigroup: This should be `OTP_ADMIN_DASHBOARD_FROM_EMAIL`.

configurations/test/env.yml
```diff
@@ -15,6 +15,23 @@ BUGSNAG_REPORTING_WINDOW_IN_DAYS: 14
...
+MONITORED_COMPONENTS:
+  - bugsnagProjectId: abcd1234
+    name: datatools-server
+  - bugsnagProjectId: abcd12345
+    name: datatools-ui
+  - bugsnagProjectId: abcd123456
+    name: otp-middleware
+
+# Config settings for linking to the OTP Admin Dashboard.
+OTP_ADMIN_DASHBOARD_EMAIL: OTP Admin Dashboard <no-reply@email.com>
```
binh-dam-ibigroup: This should be `OTP_ADMIN_DASHBOARD_FROM_EMAIL`.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -37,6 +47,9 @@
     public static final String FROM_PHONE = getConfigPropertyAsText("NOTIFICATION_FROM_PHONE");
     private static final String SPARKPOST_KEY = getConfigPropertyAsText("SPARKPOST_KEY");
     private static final String FROM_EMAIL = getConfigPropertyAsText("NOTIFICATION_FROM_EMAIL");
+    private static final String OTP_ADMIN_DASHBOARD_EMAIL = getConfigPropertyAsText("OTP_ADMIN_DASHBOARD_EMAIL");
```
binh-dam-ibigroup: This should be `OTP_ADMIN_DASHBOARD_FROM_EMAIL`.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -104,24 +117,58 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+        // Inject manage subscriptions tag.
+        if (html != null) html.with(manageSubscriptions(OTP_ADMIN_DASHBOARD_URL + "/account"));
+        return sendEmailViaSparkpost(OTP_ADMIN_DASHBOARD_EMAIL, adminUser.email, subject, text, html);
+    }
+
+    /**
+     * @return A "Manage my subscriptions" tag for email notifications.
+     */
+    public static ContainerTag manageSubscriptions(String url) {
+        return p(small(join("Manage subscriptions", a("here").withHref(url), ".")));
```
binh-dam-ibigroup: This looks like a footer... so `return footer(small(...`?

binh-dam-ibigroup: My preference for the inner content: just render `<a href="...">Manage subscriptions</a>`

binh-dam-ibigroup (changes requested):
See my comments so far. My primary recommendation is to remove reference to Bugsnag in `MonitoredComponent`.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -72,6 +76,27 @@ public static void logMessageAndHalt(Request request, int statusCode, String mes
...
 
+    /**
+     * Utility method to parse generic object from HTTP response.
+     */
+    public static <T> T getPOJOFromHttpBody(HttpResponse<String> response, Class<T> clazz) {
+        if (response == null || response.statusCode() >= 400) {
+            String result = response == null ? "bad response!" : response.body();
+            LOG.error("Error found in HTTP response: {}", result);
+            return null;
+        }
```
evansiroky: Can you make a util method for this and make both `getPOJOFromHttpBody` and `getPOJOFromHttpBodyAsList` use it?

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
evansiroky: I do not like this j2html library at all for this purpose. We should use templates instead. Seems like https://freemarker.apache.org/ is the Java go-to.

evansiroky: Ok, just to follow-up, this is definitely a knee-jerk reaction to seeing something so different than using templates. I read up on this library and it seems like it is very much oriented towards building highly-performant code. Maybe we'll reach a point where we might need this, but there are so many tradeoffs that using this library has. Templates will be much much easier to read, edit and track in version control.

src/main/java/org/opentripplanner/middleware/models/MonitoredComponent.java
evansiroky: I agree that if this model is intended to have various monitoring services, then the class should be abstract and there should be an implementing BugsnagMonitoredComponet class.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
```diff
@@ -33,34 +32,24 @@
...
             TimeUnit.MINUTES);
-
-        LOG.debug("Scheduled Bugsnag event job");
-
-        BugsnagProjectJob bugsnagProjectJob = new BugsnagProjectJob();
-
-        Scheduler.scheduleJob(
-            bugsnagProjectJob,
-            0,
-            BUGSNAG_PROJECT_JOB_DELAY_IN_MINUTES,
```
br648, line 61: The parameter BUGSNAG_PROJECT_JOB_DELAY_IN_MINUTES can be removed from this class and config.

configurations/default/env.yml.tmp
br648: I think this param should be called "BUGSNAG_MONITORED_COMPONENTS" as all other Bugsnag parameters start with this. Assuming this is used just for Bugsnag?

src/main/java/org/opentripplanner/middleware/bugsnag/EventSummary.java
```diff
@@ -3,6 +3,7 @@
 import org.opentripplanner.middleware.bugsnag.response.EventException;
 import org.opentripplanner.middleware.models.BugsnagEvent;
 import org.opentripplanner.middleware.models.BugsnagProject;
```
br648: Remove unused BugsnagProject import. I think the BugsnagProject class can be removed.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -0,0 +1,154 @@
...
+        if (!refreshedRequest.status.equalsIgnoreCase("completed")) {
+            // Request not completed by Bugsnag yet. Return and await the next cycle/refresh.
+            // TODO: Update the request data in the database? What if the status has changed since the last fetch?
+            return;
+        }
+        // First, remove "stale" requests (i.e., those that are older than this latest one).
+        Persistence.bugsnagEventRequests.removeFiltered(Filters.lte("dateCreated", request.dateCreated));
+        // Next, get and store the new events from the completed request and notify users.
+        List<BugsnagEvent> newEvents = getNewEvents(refreshedRequest);
+        if (newEvents.size() > 0) {
```
br648, line 79: This could be replaced with `!newEvents.isEmpty()`. What is the preferred approach as I've also noticed Binh using the `size()` method?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -0,0 +1,154 @@
...
+        );
+        if (latestRequest != null) {
+            // Handle the request data if it exists.
+            refreshEventRequest(latestRequest);
+        } else {
+            LOG.debug("No pending event data requests found.");
+        }
+        // FIXME: Do we want to remove these stale events? This could be confusing for users of the system and could cause
+        //  issues tracing down issues over time.
+        removeStaleEvents();
```
br648, line 61: Maybe the removal window could be greater than the reporting window (`BUGSNAG_REPORTING_WINDOW_IN_DAYS`) to give more flexibility?

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
```diff
@@ -1,5 +1,6 @@
...
 import org.opentripplanner.middleware.otp.response.Itinerary;
 import org.opentripplanner.middleware.otp.response.LocalizedAlert;
```
br648: Unused import `LocalizedAlert`

src/test/java/org/opentripplanner/middleware/BugsnagTest.java
```diff
@@ -39,13 +38,6 @@ public void canCreateBugsnagEventRequest() {
         assertEquals(BUGSNAG_EVENT_REQUEST.id, retrieved.id, "Found Bugsnag event request.");
     }
 
-    @Test
```
br648: Remove unused `BugsnagProject` (and `FileUtils`) import.

br648 (changes requested):
LGTM. A few minor comments and the readme needs updating. Should the process to obtain the organization and project ids be included somewhere as they can't simply be retrieved from the Bugsnag UI?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Convenience method to send email notification to all subscribed users.
+     */
+    private void sendEmailForEvents(List<BugsnagEvent> newEvents) {
+        // Construct email content.
+        String dashboardUrl = OTP_ADMIN_DASHBOARD_URL + "?dashboard=errors";
+        String subject = String.format("%d new error events", newEvents.size());
+        String text = String.format("Visit the %s to view new errors: %s", OTP_ADMIN_DASHBOARD_NAME, dashboardUrl);
+        ContainerTag html = div(
```
evansiroky: Same comment as https://github.com/ibi-group/otp-middleware/pull/111#discussion_r529854869.

src/main/java/org/opentripplanner/middleware/controllers/api/ErrorEventsController.java
```diff
@@ -72,18 +68,18 @@ public void bind(final SparkSwagger restApi) {
...
+        // Get Bugsnag projects by id.
+        Map<String, MonitoredComponent> componentsByProjectId = MonitoredComponent.getComponentsByProjectId();
         // Construct event summaries from project map.
         // FIXME: Group by error/project type?
         List<EventSummary> eventSummaries = events
-            .map(event -> new EventSummary(projectsById.get(event.projectId), event))
+            .map(event -> new EventSummary(componentsByProjectId.get(event.projectId), event))
             .into(new ArrayList<>());
-        return new ResponseList<>(EventSummary.class, eventSummaries, offset, limit, bugsnagEvents.getCount());
+        return new ResponseList<>(EventSummary.class, eventSummaries, offset, limit, Persistence.bugsnagEvents.getCount());
```
evansiroky: Line is too long, please split up.

evansiroky (changes requested):
I'm not entirely sure where this MonitoredComponent thing is headed, but I guess it's a good idea if we monitor other aspects of system reliability. The only glaring thing I see is the use of j2html. I think a templating engine is much better suited for creating HTML emails. I believe this will greatly improve the readability and maintainability of the code despite a slight performance sacrifice. It looks like https://freemarker.apache.org/ or http://velocity.apache.org/ might be viable possibilities for this.

configurations/default/env.yml.tmp
landonreed: The idea behind this approach is to make these components extensible. So, if we begin using logrocket to monitor the same component, we can add a logrocketId to an existing component and be able to track logrocket and bugsnag info for the same component (e.g., datatools-ui).

src/main/java/org/opentripplanner/middleware/models/JourneyState.java
landonreed: merge issue

src/main/java/org/opentripplanner/middleware/bugsnag/EventSummary.java
landonreed: I don't fully understand the question. This constructor is just used to supplement bugsnag events with information about the project from a specific MonitoredComponent.

src/main/java/org/opentripplanner/middleware/controllers/api/ErrorEventsController.java
landonreed: These events are not expected to become untied from Bugsnag anytime soon. The only reason I changed the controller class name was to clarify what its primary response was.

src/main/java/org/opentripplanner/middleware/models/MonitoredComponent.java
landonreed: It is optional in the sense that the application doesn't fail if it's missing. This exception is thrown so that a null value will be caught in the same way as if a parsing error were to occur.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
landonreed: Good question. In general I prefer non-negative conditions. So maybe if we were expecting `newEvents.size() == 0` I would be more inclined to use `newEvents.isEmpty()`. I don't feel very strongly about any of this though. For this instance, I might leave as is if that's OK?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
landonreed: I'll update the comment to include this thought.

src/main/java/org/opentripplanner/middleware/bugsnag/EventSummary.java
binh-dam-ibigroup: Hmm sorry, please disregard. I probably thought at reading that `EventSummary` was tied to `MonitoredComponent`.

src/main/java/org/opentripplanner/middleware/models/MonitoredComponent.java
binh-dam-ibigroup: What happens if someone does not enter a `MONITORED_COMPONENTS` in the config file?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
br648: @landonreed yep that is fine. I'll make a note of this for when I come across this scenerio.

configurations/default/env.yml.tmp
landonreed: Doing this appears to cause issues with the templating engine (or at least complicates the approach we've taken). I'm going to create a separate issue for this for now.

landonreed:
OK, this is ready for review again. NOTE: I did not change OTP_ADMIN_DASHBOARD config props to be nested because that complicates things with getting the values for templating, but have logged an issue for future consideration (#122).

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -113,25 +105,20 @@ private void refreshEventRequest(BugsnagEventRequest request) {
...
         String subject = String.format("%d new error events", newEvents.size());
-        String text = String.format("Visit the %s to view new errors: %s", OTP_ADMIN_DASHBOARD_NAME, dashboardUrl);
-        ContainerTag html = div(
-            h1(subject),
-            p(join(
-                "Visit the",
-                a(OTP_ADMIN_DASHBOARD_NAME).withHref(dashboardUrl),
-                "to view new errors."
-            ))
+        Map<String, Object> data = Map.of(
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`.

src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -272,28 +273,26 @@ private void sendNotifications() {
             LOG.info("Last notifications match current ones. Skipping notify.");
             return;
         }
-        String name = trip.tripName != null ? trip.tripName : "Trip for " + otpUser.email;
-        String subject = name + " Notification";
-        StringBuilder body = new StringBuilder();
-        for (TripMonitorNotification notification : notifications) {
-            body.append(notification.body);
-        }
+        Map<String, Object> data = Map.of(
```
binh-dam-ibigroup: Suggestion: rename to `templateFields` or `notificationFields`.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -38,53 +45,49 @@
...
-        if (method.equals(REQUEST_METHOD.DELETE)) {
-            httpRequestBuilder.DELETE();
-        }
-
-        if (method.equals(REQUEST_METHOD.POST)) {
-            httpRequestBuilder.POST(HttpRequest
-                .BodyPublishers
-                .ofString(bodyContent));
+            .timeout(Duration.ofSeconds(connectionTimeout));
+        switch (method) {
```
binh-dam-ibigroup, line 78: I would insert a blank line here.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -72,6 +74,36 @@ public static void logMessageAndHalt(Request request, int statusCode, String mes
         return null;
     }
 
+    /**
+     * Check if an {@link HttpResponse} is OK.
+     */
+    private static boolean isResponseOk(HttpResponse<String> response) {
```
binh-dam-ibigroup, line 81: Suggestion: rename to `isNotError`. Remove redundant comment, or state in comment what constitutes a good vs. bad response.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -37,12 +42,35 @@
...
     private static final String SPARKPOST_KEY = getConfigPropertyAsText("SPARKPOST_KEY");
     private static final String FROM_EMAIL = getConfigPropertyAsText("NOTIFICATION_FROM_EMAIL");
+    public static final String OTP_ADMIN_DASHBOARD_FROM_EMAIL = getConfigPropertyAsText("OTP_ADMIN_DASHBOARD_FROM_EMAIL");
 
     /**
-     * Send a SMS message to the provided phone number
-     * @param toPhone - e.g., +15551234
-     * @param body - SMS message body
-     * @return messageId if message was sucessful (null otherwise)
+     * Send templated SMS to {@link OtpUser}.
```
binh-dam-ibigroup: I would add `to {@link OtpUser}'s verified phone number`.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -37,12 +42,35 @@
...
+     * @param otpUser       target user
+     * @param smsTemplate   template to use for SMS message
+     * @param data          template data
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -37,12 +42,35 @@
...
+        if (!otpUser.isPhoneNumberVerified) {
+            LOG.error("Cannot send SMS to unverified user ({})!", otpUser.email);
+            return null;
+        }
+        try {
+            String body = TemplateUtils.renderTemplate(smsTemplate, data);
+            return sendSMS(otpUser.phoneNumber, body);
+        } catch (TemplateException | IOException e) {
+            // This catch indicates there was an error rendering the template. Note: TemplateUtils#renderTemplate
+            // handles Bugsnag reporting/error logging, so that is not needed here.
```
binh-dam-ibigroup, line 64: The Bugsnag reference should be removed.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -104,24 +136,97 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+    /**
+     * Send notification email to {@link OtpUser}, ensuring the correct from
+     * email address is used (i.e., {@link #FROM_EMAIL}).
+     */
+    public static boolean sendEmail(
+        OtpUser otpUser,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Object data
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -104,24 +136,97 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+    /**
+     * Send notification email to {@link AdminUser}, ensuring the correct from
+     * email address is used (i.e., {@link #OTP_ADMIN_DASHBOARD_FROM_EMAIL}).
+     */
+    public static boolean sendEmail(
+        AdminUser adminUser,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Object data
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -104,24 +136,97 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+     * @param data          template data
+     * @return              whether the email was sent successfully
+     */
+    private static boolean sendEmail(
+        String fromEmail,
+        String toEmail,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Object data
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -104,24 +136,97 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+    ) {
+        try {
+            String text = TemplateUtils.renderTemplate(textTemplate, data);
+            String html = TemplateUtils.renderTemplate(htmlTemplate, data);
+            return sendEmailViaSparkpost(fromEmail, toEmail, subject, text, html);
+        } catch (TemplateException | IOException e) {
+            // This catch indicates there was an error rendering the template. Note: TemplateUtils#renderTemplate
+            // handles Bugsnag reporting/error logging, so that is not needed here.
```
binh-dam-ibigroup, line 191: Remove the Bugsnag reference.

src/main/java/org/opentripplanner/middleware/utils/TemplateUtils.java
```diff
@@ -0,0 +1,61 @@
...
+        } catch (TemplateModelException e) {
+            LOG.error("An error occurred while initializing FreeMarker: ", e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Renders a template given an object.
+     * @param templatePath  path to template file (.ftl that is found in the {@link #BASE_TEMPLATE_PATH} resources
+     *                     directory
```
binh-dam-ibigroup: Is the IDE doing these indent mismatch?

src/main/java/org/opentripplanner/middleware/utils/TemplateUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param data          template data (any kind of public class that has public getXxx/isXxx methods as prescribed
```
binh-dam-ibigroup: Suggestion: rename to `templateFields`. Also, I think we should restrict the type to `Map<String, Object>`, that is the only type passed to this method so far. We can add an overload method if other needs arise.

binh-dam-ibigroup (commented):
I have to play some more, my main comment for now is making the `data -> templateFields` argument of `TemplateIUtils#renderTemplate` a `Map<String, Object>`. There are also renaming and javadoc suggestions.

src/main/java/org/opentripplanner/middleware/utils/TemplateUtils.java
landonreed: Yep. Weird, eh?

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
landonreed: I don't understand why we would remove the Bugsnag reference? We're using Bugsnag to report the error.

binh-dam-ibigroup: This PR is abstracting Bugsnag out, so removing the Bugsnag reference is consistent with that purpose.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagDispatcher.java
```diff
@@ -112,133 +93,62 @@ private static void buildEventRequestFilter() {
...
+        // Create new request if null ID is provided.
+        boolean create = eventDataRequestId == null;
+        URI eventDataRequestUri = HttpUtils.buildUri(
             BUGSNAG_API_URL,
-            "/organizations/" + ORGANIZATION.id + "/projects",
-            null);
-
-        String response = HttpUtils.httpRequest(
-            uri,
+            "organizations", BUGSNAG_ORGANIZATION, "event_data_requests", eventDataRequestId
```
binh-dam-ibigroup, line 126: Is `BUGSNAG_ORGANIZATION` the right parameter to use here? Did the meaning of that parameter change?  Previously we would pass a Bugsnag organization ID.

binh-dam-ibigroup: It looks like the meaning of `BUGSNAG_ORGANIZATION` has changed to be a Bugsnag organization id. README and comments in `env.yml.tmp` and `test/env.yml` should be updated accordingly.

binh-dam-ibigroup (changes requested):
Some README and env.yml comment updates are needed per my comment https://github.com/ibi-group/otp-middleware/pull/111#discussion_r538797179. Also per our chat, please add instructions in README on how to obtain Bugsnag project ids.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
landonreed: Bugsnag is still very much a part of the system though. Nothing about Bugsnag reporting has changed and we're still tracking all of the errors in Bugsnag.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
landonreed: I think I prefer `templateData` here and elsewhere.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
landonreed: Bugsnag is still used to report errors, so I think I'll keep this in place.

src/main/java/org/opentripplanner/middleware/utils/TemplateUtils.java
landonreed: I made this change previously and @evansiroky had some issues with it, so in the interest of moving forward, I think I'll leave as is.

README.md
```diff
@@ -51,6 +51,17 @@ The follow parameters are used to interact with an OTP server.
 | OTP_API_ROOT | This is the address of the OTP server, including the root path to the OTP API, to which all OTP related requests will be sent to. | http://otp-server.example.com/otp | 
 | OTP_PLAN_ENDPOINT | This defines the plan endpoint part of the requesting URL. If a request is made to this, the assumption is that a plan request has been made and that the response should be processed accordingly. | /plan |
 
+### Monitored Components
+
+This application allows you to monitor various system components (e.g., OTP API, OTP UI, Data Tools) that work together
```
br648: Typo (e.g. OTP API, OTP UI & Data Tools)

br648 (commented):
One minor comment, plus the TemplateUtilsTest unit tests don't work on a Windows env (result was \r\n and not \n). Intellij is configured to use Unix and macOS (\n) line separator, but this makes no difference. Obviously this only affects me, but it would be nice for all the unit tests to pass :-)

README.md
```diff
@@ -101,10 +112,11 @@ Bugsnag (https://app.bugsnag.com), clicking on settings (top right hand corner)
 select `Personal auth tokens` and then `Generate new token`.
 
 ##### BUGSNAG_ORGANIZATION
-A bugsnag organization contains all projects which errors/events will be reported on. The organization name and more 
-specifically its ID is the starting point for most Bugsnag API requests. The organization name can be obtained by 
-logging into Bugsnag (https://app.bugsnag.com), clicking on settings (top right hand corner) then `Organization settings`. 
-From here, click on the organization name and then copy the name from the pop-up window.
+A bugsnag organization contains all projects which errors/events will be reported on. The organization ID is the 
+starting point for most Bugsnag API requests. The organization ID can be obtained by opening the Chrome Devtools Network
```
binh-dam-ibigroup: Remove the Chrome reference. How about "by opening the Network tab in your browser's developer tools, ..."

README.md
```diff
@@ -101,10 +112,11 @@ Bugsnag (https://app.bugsnag.com), clicking on settings (top right hand corner)
 select `Personal auth tokens` and then `Generate new token`.
 
 ##### BUGSNAG_ORGANIZATION
-A bugsnag organization contains all projects which errors/events will be reported on. The organization name and more 
-specifically its ID is the starting point for most Bugsnag API requests. The organization name can be obtained by 
-logging into Bugsnag (https://app.bugsnag.com), clicking on settings (top right hand corner) then `Organization settings`. 
-From here, click on the organization name and then copy the name from the pop-up window.
+A bugsnag organization contains all projects which errors/events will be reported on. The organization ID is the 
+starting point for most Bugsnag API requests. The organization ID can be obtained by opening the Chrome Devtools Network 
+tab, then opening the Bugsnag dashboard (https://app.bugsnag.com). Filter the network requests with
```
binh-dam-ibigroup: "then **navigating** to the Bugsnag dashboard..."

binh-dam-ibigroup (approved):
Thanks for the changes. There are some minor documentation tweaks that remain, but good to go otherwise.

PR #114, opened 2020-12-01
Fix ConflictException: API Key already subscribed to usage plan

landonreed:
### Checklist

### Description

Be tolerant of ConflictException encountered while adding API key to usage plan.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
+++ b/src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
@@ -72,6 +72,7 @@ public class ApiGatewayUtils {
         AmazonApiGateway gateway;
         CreateApiKeyResult apiKeyResult = null;
         GetUsagePlanResult usagePlanResult;
+        String keyId = null;
         try {
             gateway = getAmazonApiGateway();
             // Before creating key, verify usage plan exists (if not an exception will be thrown and caught below).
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):

br648 (approved):
LGTM.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
```diff
@@ -86,23 +93,28 @@ public static ApiKey createApiKey(ApiUser user, String usagePlanId) throws Creat
...
+        } catch (Exception e) {
+            CreateApiKeyException createApiKeyException = new CreateApiKeyException(user.id, usagePlanId, e);
 
-            // add API key to usage plan
-            CreateUsagePlanKeyRequest usagePlanKeyRequest = new CreateUsagePlanKeyRequest();
-            usagePlanKeyRequest
+            BugsnagReporter.reportErrorToBugsnag("Error creating API key", keyId, createApiKeyException);
+            throw createApiKeyException;
+        }
+        try {
```
evansiroky: I don't quite understand why there are 2 try/catch blocks here. It seems like there should be just one try/catch and the `catch (Exception e)` block should be retained in almost the same position where it was, only this time it'd be included right after the `catch (ConflictException e)` block. This way it seems that any Exception generated from the `gateway.createUsagePlanKey(usagePlanKeyRequest)` call that isn't a `ConflictException` will be properly caught.

evansiroky (changes requested):
Please address https://github.com/ibi-group/otp-middleware/pull/114#discussion_r534357759.

src/main/java/org/opentripplanner/middleware/utils/ApiGatewayUtils.java
landonreed: I don't know if ConflictExceptions can be thrown by `AmazonApiGateway#createApiKey`. There may be ConflictExceptions that we want to cause a `CreateApiKeyException` rather than just continue with CreateUsagePlanKeyRequest. I'd rather err on the side of caution and not try to bundle these calls up into the same block.

evansiroky: Ok, that reasoning sounds good enough to keep the 2 try/catch blocks. However, I think 2 improvements should be made:

1. Add some comments about this decision
2. Add another `catch(Exception e)` block to the 2nd try/catch block to make sure that any other funky AWS client error is properly caught and reported.

landonreed: See nested solution in dc6630e8a2470cd01f38a49a62c3d86c554dfde5

evansiroky (approved):

PR #116, opened 2020-12-02
Use Apache FreeMarker as a template engine

evansiroky:
### Checklist

### Description

Following up on https://github.com/ibi-group/otp-middleware/pull/111#discussion_r529854869, this PR replaces j2html with [FreeMarker](https://freemarker.apache.org/). I strongly believe using a templating engine is highly preferable to j2html for the following reasons:

1. To adhere to the [Separation of Concerns](https://deviq.com/separation-of-concerns/) principle.
1. Once we begin to need to make emails that don't consist of really short text, we will need to better style and format the HTML in the emails. I believe this will be much easier to implement and maintain with templates since j2html would require all kinds of method calling and could become quite unreadable quite quickly.
1. It will be much faster to edit HTML in some test editor online and then paste the HTML into a template and then add the necessary FreeMarker markup.
1. Now that this template engine has been setup, it could be used for things other than just making HTML emails.


src/main/java/org/opentripplanner/middleware/tripMonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -240,12 +240,12 @@ private void sendNotifications() {
                 success = NotificationUtils.sendSMS(otpUser.phoneNumber, body.toString()) != null;
                 break;
             case "email":
-                success = NotificationUtils.sendEmail(otpUser, subject, body.toString(), null);
+                success = NotificationUtils.sendTextOnlyEmail(otpUser, subject, body.toString());
```
landonreed: I don't think this method needs to exist. The only reason we're not yet sending html here is that we haven't gotten around to deciding how this should be formatted. If anything we should just add a FIXME about constructing an HTML email.

src/main/resources/templates/AdminUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
+<#include "BaseEmail.ftl">
+
+<#macro email_main>
+<#-- Provide an error message in the default case. If this hasn't been overwritten by an inhereting
```
landonreed: Typo: inheret -> inherit

src/main/resources/templates/AdminUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+        component, then an error has occurred. -->
+    The main content of this email haven't been configured! Sorry about that! Please contact the webmaster.
```
landonreed: haven't -> hasn't

src/main/resources/templates/AdminUserContainer.ftl
landonreed: Also, does this error get thrown as a TemplateException or does it get delivered to the user? I think I would prefer that the email just doesn't go out at all.

src/main/resources/templates/BaseEmail.ftl
```diff
@@ -0,0 +1,11 @@
+<#macro email_body>
+<#-- Provide an error message in the default case. If this hasn't been overwritten by an inhereting
```
landonreed: inheret -> inherit

src/main/resources/templates/OtpUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
+<#include "BaseEmail.ftl">
+
+<#macro email_main>
+<#-- Provide an error message in the default case. If this hasn't been overwritten by an inhereting
```
landonreed: inheret -> inherit

src/main/resources/templates/OtpUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+        component, then an error has occurred. -->
+The main content of this email haven't been configured! Sorry about that! Please contact the webmaster.
```
landonreed: haven't -> hasn't

src/main/resources/templates/OtpUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+</#macro>
+
+<#macro email_body>
+<#-- header -->
+<div>
+    <h2>An email from the ${OTP_UI_NAME}</h2>
+</div>
```
landonreed, line 13: Not really sure this heading should appear in the emails (same goes for AdminUserContainer). If you're looking to include the otp-ui name in the body for snapshotting, I would add it to the subscriptions message:

> Manage ${OTP_UI_NAME} subscriptions <a href="${OTP_UI_URL}/#/account">here</a>.

src/main/resources/templates/OtpUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+<#-- main -->
+<div>
+    <@email_main/>
+</div>
+<#-- footer -->
+<div>
+    <p>
+        <small>
+            Manage Subscriptions <a href="${OTP_UI_URL}/#/account">here</a>.
```
landonreed, line 22: Subscriptions -> subscriptions

src/main/resources/templates/AdminUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+</div>
+<#-- main -->
+<div>
+    <@email_main/>
+</div>
+<#-- footer -->
+<div>
+    <p>
+        <small>
+            Manage Subscriptions <a href="${OTP_ADMIN_DASHBOARD_URL}/account">here</a>.
```
landonreed: Subscriptions -> subscriptions

src/main/resources/templates/AdminUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
...
+<#-- Provide an error message in the default case. If this hasn't been overwritten by an inhereting
+        component, then an error has occurred. -->
+    The main content of this email haven't been configured! Sorry about that! Please contact the webmaster.
+</#macro>
+
+<#macro email_body>
+<#-- header -->
+<div>
+    <h2>An email from the ${OTP_ADMIN_DASHBOARD_NAME}</h2>
+</div>
```
landonreed, line 13: Same comment as for OtpUserContainer (heading doesn't belong here).

landonreed (changes requested):
I was skeptical, but I think you're right on the mark with this approach. I have a handful of changes, but most of them are included in #117. I'll try to go ahead and resolve the ones that are already fixed there.

src/main/resources/templates/OtpUserContainer.ftl
landonreed: This is probably OK for now. Maybe we'll replace it with something else moving forward.

src/main/resources/templates/AdminUserContainer.ftl
landonreed: Probably OK for now.

src/main/resources/templates/AdminUserContainer.ftl
evansiroky: Yeah, it would show up. I removed this in 2321b170be74b6317fc9c2c826158342e4daf32e.

evansiroky:
Ready for review again. I did the following:

- merged in #117
- updated templates to not have default macros
- updated templates to have comments about what they're about
- improved the template testing so that individual template rendering test cases can be ran with proper snapshot matching

src/test/java/org/opentripplanner/middleware/utils/TemplateUtilsTest.java
```diff
@@ -39,30 +47,54 @@ public void canRenderTemplates(TemplateRenderingTestCase testCase) throws Except
...
+            // enforce unique test case names so that snapshotting works properly
+            if (testCaseNames.contains(testCaseName)) {
+                throw new IllegalArgumentException(
+                    String.format(
+                        "A test case with the name %s already exists! Test case names must be unique.",
+                        testCaseName
+                    )
+                );
+            }
+            testCaseNames.add(testCaseName);
```
landonreed, line 90: Non-blocking nit: we could just check `if (!testCaseNames.add(testCaseName))` above rather than a contains and add call. I don't feel strongly about this though (one could argue that the current way is more readable).

src/main/resources/templates/OtpUserContainer.ftl
```diff
@@ -1,10 +1,10 @@
...
 
-<#macro email_main>
-<#-- Provide an error message in the default case. If this hasn't been overwritten by an inheriting
-        component, then an error has occurred. -->
-The main content of this email hasn't been configured! Sorry about that! Please contact the webmaster.
-</#macro>
+<#--
+    This is a template for all emails related to the OTP end-user application. Any templates that
+    use this template must implement a macro called `email_main`. The whole HTML email can then be
+    rendered by the statement <@html_email/>.
```
landonreed: Would it be more clear to call this statement `<@render_html_email/>`? Maybe not -- I'm not sure what the convention is for macro naming. The more I think about this, it seems like maybe the way you have it is best.

landonreed (approved):
Looks great. Thanks for putting this together. A couple of small comments that are non-blocking.

src/main/resources/templates/MonitoredTripSms.ftl
```diff
@@ -0,0 +1,13 @@
...
+-->
+
+Your trip has the following notifications:
+<#list notifications as notification>
+  -${notification}
+</#list>
+
+View trip: ${OTP_UI_URL}/#/savedtrips/${tripId}
+
+To stop receiving notifications, reply STOP.
```
binh-dam-ibigroup, line 13: Unless we are actively working on listening to replies to these SMS, we shouldn't mention this.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -45,7 +45,7 @@ public static void setUp() throws IOException, InterruptedException {
     public static void tearDown() {
         // Delete the users if they were not already deleted during the test script.
         otpUser = Persistence.otpUsers.getById(otpUser.id);
-        if (otpUser != null) otpUser.delete();
+        if (otpUser != null) otpUser.delete(false);
```
binh-dam-ibigroup, line 50: Add a comment to explain the need to preserve the Auth0 account.

src/main/resources/templates/MonitoredTripHtml.ftl
```diff
@@ -0,0 +1,21 @@
+<#include "OtpUserContainer.ftl">
+
+<#--
+    This is a template for an HTML email that gets sent when there are new notifications about an
+    OTP user's monitored trip.
+-->
+
+<#macro email_main>
+    <div>
+        <h2></h2>
```
binh-dam-ibigroup: Not sure why there is this empty tag, and we will need to refine which headings use which tags.

binh-dam-ibigroup (commented):
See my preliminary comments. I have to play more with the PR, but no big issues so far except for the templates we will need to refine later.

src/main/resources/templates/AdminUserContainer.ftl
```diff
@@ -0,0 +1,27 @@
+<#include "BaseEmail.ftl">
+
+<#--
+    This is a template for all emails related to the OTP Admin Dashboard. Any templates that use
+    this template must implement a macro called `email_main`. The whole HTML email can then be
+    rendered by the statement <@html_email/>.
+-->
+
+<#macro email_body>
```
landonreed: Change to capitalized camelcase.

src/main/resources/templates/MonitoredTripSms.ftl
landonreed: @binh-dam-ibigroup, I think this generally comes out-of-the-box with any text messaging service: https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-

Don't worry, I have no intention to communicate with all of our users via SMS :)

PR #117, opened 2020-12-02
refactor(expand use of templates to CheckMonitoredTrip):

landonreed:
### Checklist

### Description

Refactor notification utils a bit and builds on use of FreeMarker to include templates in CheckMonitoredTrip.


src/test/java/org/opentripplanner/middleware/utils/TemplateUtilsTest.java
```diff
@@ -38,18 +38,25 @@ public void canRenderTemplates(TemplateRenderingTestCase testCase) throws Except
...
+        );
+        testCases.add(new TemplateRenderingTestCase(notificationsData, "MonitoredTripSms.ftl"));
+        testCases.add(new TemplateRenderingTestCase(notificationsData, "MonitoredTripText.ftl"));
+        testCases.add(new TemplateRenderingTestCase(notificationsData, "MonitoredTripHtml.ftl"));
         return testCases;
     }
 
     private static class TemplateRenderingTestCase {
-        public final Object data;
+        public final Map data;
```
evansiroky: This should stay as an Object because according to the [FreeMarker docs](https://freemarker.apache.org/docs/pgui_quickstart_createdatamodel.html) it is possible to use...

>  any kind of public class that has public getXxx/isXxx methods as prescribed by the JavaBeans specification

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
      */
-    public static boolean sendEmail(AdminUser adminUser, String subject, String text, String html) {
-        return sendEmailViaSparkpost(OTP_ADMIN_DASHBOARD_EMAIL, adminUser.email, subject, text, html);
+    private static boolean sendEmail(
+        String fromEmail,
+        String toEmail,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Map<String, Object> data
```
evansiroky: This should be an Object because according to the [FreeMarker docs](https://freemarker.apache.org/docs/pgui_quickstart_createdatamodel.html) it is possible to use...

>  any kind of public class that has public getXxx/isXxx methods as prescribed by the JavaBeans specification

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
-     * email address is used (i.e., {@link #FROM_EMAIL}).
-     */
-    public static boolean sendEmail(OtpUser otpUser, String subject, String text, String html) {
-        return sendEmailViaSparkpost(FROM_EMAIL, otpUser.email, subject, text, html);
+    public static boolean sendEmail(
+        AdminUser adminUser,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Map<String, Object> data
```
evansiroky: This should be an Object because according to the [FreeMarker docs](https://freemarker.apache.org/docs/pgui_quickstart_createdatamodel.html) it is possible to use...

>  any kind of public class that has public getXxx/isXxx methods as prescribed by the JavaBeans specification

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
      * email address is used (i.e., {@link #FROM_EMAIL}).
      */
-    public static boolean sendTextOnlyEmail(OtpUser otpUser, String subject, String text) {
-        return sendEmailViaSparkpost(FROM_EMAIL, otpUser.email, subject, text, null);
+    public static boolean sendEmail(
+        OtpUser otpUser,
+        String subject,
+        String textTemplate,
+        String htmlTemplate,
+        Map<String, Object> data
```
evansiroky: This should be an Object because according to the [FreeMarker docs](https://freemarker.apache.org/docs/pgui_quickstart_createdatamodel.html) it is possible to use...

>  any kind of public class that has public getXxx/isXxx methods as prescribed by the JavaBeans specification

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -40,13 +42,26 @@
     public static final String FROM_PHONE = getConfigPropertyAsText("NOTIFICATION_FROM_PHONE");
     private static final String SPARKPOST_KEY = getConfigPropertyAsText("SPARKPOST_KEY");
     private static final String FROM_EMAIL = getConfigPropertyAsText("NOTIFICATION_FROM_EMAIL");
-    private static final String OTP_ADMIN_DASHBOARD_EMAIL = getConfigPropertyAsText("OTP_ADMIN_DASHBOARD_EMAIL");
+    public static final String OTP_ADMIN_DASHBOARD_EMAIL = getConfigPropertyAsText("OTP_ADMIN_DASHBOARD_EMAIL");
 
+    public static String sendSMS(OtpUser otpUser, String smsTemplate, Map<String, Object> data) {
```
evansiroky: Could use JavaDoc.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+        String htmlTemplate,
+        Map<String, Object> data
+    ) {
+        return sendEmail(OTP_ADMIN_DASHBOARD_EMAIL, adminUser.email, subject, textTemplate, htmlTemplate, data);
     }
 
     /**
      * Send notification email for {@link AdminUser}, ensuring the correct from
      * email address is used (i.e., {@link #OTP_ADMIN_DASHBOARD_EMAIL}).
      */
```
evansiroky, line 176: JavaDoc seems like it should be for the above method which is now missing JavaDoc.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+        Map<String, Object> data
+    ) {
+        try {
+            String text = TemplateUtils.renderTemplate(textTemplate, data);
+            String html = TemplateUtils.renderTemplate(htmlTemplate, data);
+            return sendEmailViaSparkpost(fromEmail, toEmail, subject, text, html);
+        } catch (TemplateException | IOException e) {
+            // This catch indicates there was an error rendering the template.
+            return false;
+        }
```
evansiroky, line 193: Haha, I originally did it this way too, but wondered if it was placing too much responsibility in here, but I guess this makes sense if we're heading in the direction of always using templates for emails which seems like a good idea.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -109,34 +128,56 @@ public static VerificationCheck checkSmsVerificationCode(String phoneNumber, Str
...
+        String textTemplate,
+        String htmlTemplate,
+        Map<String, Object> data
+    ) {
+        try {
+            String text = TemplateUtils.renderTemplate(textTemplate, data);
+            String html = TemplateUtils.renderTemplate(htmlTemplate, data);
+            return sendEmailViaSparkpost(fromEmail, toEmail, subject, text, html);
+        } catch (TemplateException | IOException e) {
+            // This catch indicates there was an error rendering the template.
```
evansiroky: Might be useful to note that `TemplateUtils.renderTemplate` handles reporting template rendering errors to Bugsnag, so that isn't done here.

evansiroky (changes requested):
I like the direction this is headed, but have a few minor comments that should probably be addressed.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
landonreed: Yea, I think it makes sense to reinforce the approach by providing a specific method that can more easily get re-used.

evansiroky (approved):

PR #120, opened 2020-12-03
Utilize Auth0 Scope

### LINKED ISSUE #94: Expected immutable parameters can be altered in submitted data
## Observed behavior

Example regarding Monitored Trips, but a similar issue will exists with all submitted data:

Need to check that the requesting user can manage the user id assigned to the entity to update, otherwise there is a possibility that an ApiUser can assign an entity (e.g. a MonitoredTrip) to a user not managed by that ApiUser.

## Expected behavior

The content should be varified to confirm that no immutable parameters have been changed e.g. userId.

br648:
### Checklist

### Description

The process of defining the requesting user has been updated and is now based on the scope provided as part of a user's bearer token. This results in a single user type (otp-user, api-user & admin-user) being defined under the requesting user where previously multiple types could be defined (based on the auth0UserId). 

The bearer token 'scope' claim must contain one of otp-user, api-user or admin-user for the user to be correctly matched to a user held in MongoDB.

I have only tested this using the https://auth0.com/docs/flows/call-your-api-using-resource-owner-password-flow approach (see `Auth0Users.getApiUserAuth0TokenResponse`) but I'm expecting the scope aspect to be managed the same with other auth0 auth methods.

### Required Auth0 Rule

For this to work, the following rule must be added to your Auth0 tenant:

```
function (user, context, callback) {
    const req = context.request;
    // Retrieve scopes either from the parameters or body
    const requestedScopeString = (req.query && req.query.scope) || (req.body && req.body.scope);
    context.accessToken.scope = requestedScopeString;
    return callback(null, user, context);
}
```
I have added this rule (Assign Scope) to the FDOT tenant.

This PR also fixes https://github.com/ibi-group/otp-middleware/issues/94 "Expected immutable parameters can be altered in submitted data"


src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
```diff
@@ -25,37 +31,54 @@
...
-        otpUser = Persistence.otpUsers.getOneFiltered(withAuth0UserId);
-        apiUser = Persistence.apiUsers.getOneFiltered(withAuth0UserId);
-        adminUser = Persistence.adminUsers.getOneFiltered(withAuth0UserId);
+        if (scope == null || scope.isEmpty()) {
+            LOG.error("Required scope claim unavailable");
+            return;
+        }
+
+        if (scope.contains(OtpUser.SCOPE)) {
+            otpUser = (testing)
```
landonreed, line 64: There might be an opportunity to consolidate these user fields into a single `AbstractUser user` field, but I'm not sure about that approach. Just pointing it out in case we want to add a TODO.

src/test/java/org/opentripplanner/middleware/GetMonitoredTripsTest.java
```diff
@@ -148,10 +148,10 @@ public void canGetOwnMonitoredTrips() throws URISyntaxException, JsonProcessingE
         HttpResponse<String> multiTripsResponse = mockAuthenticatedGet(MONITORED_TRIP_PATH, multiOtpUser);
         ResponseList<MonitoredTrip> multiTrips = JsonUtils.getResponseListFromJSON(multiTripsResponse.body(), MonitoredTrip.class);
 
-        // Multi Otp user has 'enhanced' admin credentials both trips will be returned. The expectation here is that the UI
-        // will always provide the user id to prevent this (as with the next test).
+        // Multi Otp user has 'enhanced' admin credentials, still expect only 1 trip to be returned as the scope will
+        // limit the requesting user to a single 'opt-user' user type.
```
landonreed: typo: opt-user

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -231,6 +236,13 @@ public static void setAuthDisabled(boolean authDisabled) {
         Auth0Connection.authDisabled = authDisabled;
     }
 
+    /**
+     * Override the current {@link #scope} value. This is used principally for setting up test environments.
+     */
+    public static void setScope(String scope) {
+        Auth0Connection.scope = scope;
+    }
```
landonreed: In `TestUtils#getMockHeaders` we set the `Authorization` header with the `auth0UserId` and then extract that in `RequestingUser#createTestUser`. I think it might following the same pattern (with a `scope` header) instead of setting `Auth0Connection#scope` might be a good idea, just so we're keeping all of that code in the same place.

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -115,13 +129,24 @@ public static boolean getBooleanEnvVar(String var) {
...
         }
 
         // Otherwise, get a valid oauth token for the user
-        headers.put("Authorization", "Bearer " + Auth0Users.getAuth0AccessToken(requestingUser.email, TEMP_AUTH0_USER_PASSWORD));
+        headers.put("Authorization", "Bearer " + getMockAuth0AccessToken(requestingUser.email, scope));
         return headers;
     }
 
+    private static String getMockAuth0AccessToken(String username, String scope) {
+        HttpResponse<String> response = Auth0Users.getApiUserAuth0TokenResponse(username, TestUtils.TEMP_AUTH0_USER_PASSWORD, scope);
```
landonreed: Hmm, if we're using this method to get a token with different scope values, maybe the method name `Auth0Users#getApiUserAuth0TokenResponse` doesn't quite work. Maybe a better one would be `Auth0Users#getAuth0TokenWithScope`?

src/test/java/org/opentripplanner/middleware/TestUtils.java
```diff
@@ -115,13 +129,24 @@ public static boolean getBooleanEnvVar(String var) {
...
+            scope = OtpUser.SCOPE;
         }
 
         // Otherwise, get a valid oauth token for the user
-        headers.put("Authorization", "Bearer " + Auth0Users.getAuth0AccessToken(requestingUser.email, TEMP_AUTH0_USER_PASSWORD));
+        headers.put("Authorization", "Bearer " + getMockAuth0AccessToken(requestingUser.email, scope));
         return headers;
     }
 
+    private static String getMockAuth0AccessToken(String username, String scope) {
```
landonreed: Method name feels a little off -- We are getting a real Auth0 token here, right? Maybe just `getTestAuth0AccessToken`? Also, could use some javadoc.

landonreed (changes requested):
This looks great. I think a couple of things could be tweaked. I'm going to assign @binh-dam-ibigroup to give it a once over for some of those authorization edge cases he spotted early on to see how this helps.

Oh, and could you also add that Auth0 rule to the README. I know the Auth0 section is still mostly missing, but it would be good to start filling that out.

binh-dam-ibigroup:
>  I'm going to assign @binh-dam-ibigroup to give it a once over for some of those authorization edge cases he spotted early on to see how this helps.

Good call @landonreed. Using the code in this PR, as an API user, I am able to assign a trip of a user I manage to a user I don't manage, which we said should not be permitted.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -44,6 +44,11 @@
      */
     private static boolean authDisabled = getDefaultAuthDisabled();
 
+    /**
+     * The scope to be used when auth is disabled, defaulted to an OTP user scope.
+     */
+    private static String scope = OtpUser.SCOPE;
```
binh-dam-ibigroup: Should `OtpUser.SCOPE` be named `OtpUser.AUTH0_SCOPE`?

binh-dam-ibigroup: And if variable `scope` is used for testing only, it should probably be called `testScope` or `auth0TestScope`.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -11,6 +11,7 @@
 import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.bugsnag.BugsnagReporter;
 import org.opentripplanner.middleware.models.AbstractUser;
+import org.opentripplanner.middleware.models.ApiUser;
```
binh-dam-ibigroup: This import and a few others show as unused in IntelliJ.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: That is my impression too, so a TODO is in order.

src/main/java/org/opentripplanner/middleware/models/AdminUser.java
```diff
@@ -10,6 +10,7 @@
  * Represents an administrative user of the OTP Admin Dashboard (otp-admin-ui).
  */
 public class AdminUser extends AbstractUser {
+    public static final String SCOPE = "admin-user";
```
binh-dam-ibigroup: Should rename as `AUTH0_SCOPE`.

src/main/java/org/opentripplanner/middleware/models/ApiUser.java
```diff
@@ -15,6 +15,7 @@
  * access otp-middleware's endpoints (as well as the geocoding and OTP endpoints).
  */
 public class ApiUser extends AbstractUser {
+    public static final String SCOPE = "api-user";
```
binh-dam-ibigroup: Should rename as `AUTH0_SCOPE`.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -16,6 +16,7 @@
  * can also opt-in to storing their trip planning requests/responses.
  */
 public class OtpUser extends AbstractUser {
+    public static final String SCOPE = "otp-user";
```
binh-dam-ibigroup: Should rename as `AUTH0_SCOPE`.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: In the admin UI, a user could be both an `AdminUser` and an `ApiUser`, how would that work out for them?

binh-dam-ibigroup (commented):
For now it is mostly naming of variables. Pleas provide guidance for the admin UI for users who are both `AdminUser` and `OtpUser`.

Out of scope for this comment is the issue I spotted dealing with an `ApiUser` assigning a trip to another user they don't manage, I need to investigate more.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
br648: @binh-dam-ibigroup it will be the responsibility of the calling service in this case the admin UI to define the scope of the calling user. This would either be `admin-user` or `api-user`. It would be good to test this from a third party source if possible? To date, I've only tested via unit tests with `Auth0Users#getApiUserAuth0TokenResponse`.

src/test/java/org/opentripplanner/middleware/GetMonitoredTripsTest.java
br648: Fixed.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/TestUtils.java
br648: Updated.

src/test/java/org/opentripplanner/middleware/TestUtils.java
br648: Updated.

binh-dam-ibigroup (changes requested):
Please consider the fix from #121 regarding assigning a trip to an other non-managed user, and advise regarding what a user who is both an `AdminUser` and `ApiUser` should see in the admin dashboard. Merge conflicts need to be resolved too. Thanks!

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: Because both kinds of users can log in from the same page, I would guess the admin UI should set both scopes. As a result, I would recommend to split the `else if` into separate `if` blocks.

br648:
> 
> 
> > I'm going to assign @binh-dam-ibigroup to give it a once over for some of those authorization edge cases he spotted early on to see how this helps.
> 
> Good call @landonreed. Using the code in this PR, as an API user, I am able to assign a trip of a user I manage to a user I don't manage, which we said should not be permitted.

@binh-dam-ibigroup I think this is covered in this issue https://github.com/ibi-group/otp-middleware/issues/94 to be addressed in a future PR.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: The scope is now proivded via a header param so this is no longer present.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
br648: Corrected.

src/main/java/org/opentripplanner/middleware/models/AdminUser.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/ApiUser.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
br648: Updated.

README.md
```diff
@@ -23,7 +23,8 @@ OTP Middleware is also suitable for continuous deployment implementations.
 ## Development
 To run otp-middleware in a local, development environment:
 
-```bash
+```
+bash
```
landonreed: This `bash` should be on the same line as the three ticks. It indicates what syntax/language the code block should be rendered with.

README.md
```diff
@@ -43,6 +44,27 @@ java -jar target/otp-middleware.jar configurations/default/env.yml
...
+The requesting user type which determines the level of authorization is based on the scope provided as part of a user's bearer token. 
+The bearer token 'scope' claim must contain one of otp-user, api-user or admin-user for the user to be correctly matched to a user held in the database.
+
+##### Auth0 Scope Rule
+
+A rule must be added to the Auth0 tenant for the scope provided by third parties to be available for authorization within the OTP-middleware.
+This rule takes the scope value provided by the caller and adds it to the access token.  
+
+```
+function (user, context, callback) {
```
landonreed, line 57: Similar to above (bash) this could have `javascript` indicator next to the three ticks.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
```diff
@@ -25,46 +31,70 @@
...
+                : Persistence.adminUsers.getOneFiltered(withAuth0UserId);
+        } else {
+            LOG.error("No user type for scope {} is available", scope);
+        }
     }
 
     /**
      * Utility method for creating a test user. If a Auth0 user Id is defined within the Authorization header param
-     * define test user based on this.
+     * define test user based on this. If the
```
landonreed: Looks like this javadoc didn't get finished.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -84,6 +85,33 @@ MonitoredTrip preCreateHook(MonitoredTrip monitoredTrip, Request req) {
...
+        checkTripCanBeMonitored(monitoredTrip, req);
+        checkTripBelongsToManagedUser(monitoredTrip, req);
+        processTripQueryParams(monitoredTrip, req);
+    }
+
+    /**
+     * Checks that the monitored trip to be written belongs to the requesting user
+     * or to an OtpUser managed by the requesting ApiUser.
+     *
+     * For update (PUT) requests, this check is in addition to MonitoredTrip#canBeManagedBy
```
landonreed: This code ref in javadoc should have an `@link` tag: `{@link MonitoredTrip#canBeManagedBy}`.

Also, I find this javadoc a little confusing... `For update (PUT) requests, this check is in addition...` <-- does this mean the below is redundant? If it's not redundant, how is it different from what's in `MonitoredTrip#canBeManagedBy`?

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
landonreed: @binh-dam-ibigroup, the `else if` blocks were added so that we could avoid cases were a requester is identified as more than one user for a given request. Maybe we can think of a way to keep these in place for the admin UI. How about this:
```java
        Bson withAuth0UserId = Filters.eq("auth0UserId", auth0UserId);
        if (scope == null || scope.isEmpty()) {
            LOG.error("Required scope claim unavailable");
            return;
        }
        // Define only a single user according to the scope. Note: there is an edge case where a user logging in from 
        // the OTP Admin Dashboard may be both an AdminUser and ApiUser, but this code block will force their identity
        // as an AdminUser.
        // TODO: Consider consolidating the user scope fields into a single AbstractUser user field.
        if (scope.contains(OtpUser.AUTH0_SCOPE)) {
            otpUser = (testing)
                ? new OtpUser()
                : Persistence.otpUsers.getOneFiltered(withAuth0UserId);
            return;
        }
        if (scope.contains(AdminUser.AUTH0_SCOPE)) {
            adminUser = (testing)
                ? new AdminUser()
                : Persistence.adminUsers.getOneFiltered(withAuth0UserId);
            return;
        }
        if (scope.contains(ApiUser.AUTH0_SCOPE)) {
            apiUser = (testing)
                ? new ApiUser()
                : Persistence.apiUsers.getOneFiltered(withAuth0UserId);
            return;
        }
        LOG.error("No user type for scope {} is available", scope);
```

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -22,7 +22,7 @@
 
 public class HttpUtils {
     private static final Logger LOG = LoggerFactory.getLogger(HttpUtils.class);
-    public enum REQUEST_METHOD {GET, POST, DELETE}
+    public enum REQUEST_METHOD {GET, POST, PUT, DELETE} // FIXME: use sn existing library enum?
```
landonreed: This has been removed in the error-notifications PR (#111). Depending on which gets merged first we'll hit a conflict. Guess that's unavoidable though.

landonreed (changes requested):
Looking good, just a few small items that need addressing.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: I was thinking something like that. There is a side effect, though, with this change. If you log in as an admin, you will be prompted to create an API user profile, even though you have created one already.

landonreed: I don't understand why that would be the case. If that's due to something weird in the OTP Admin UI, we should just fix it.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: That's right, `MonitoredTrip#canBeManagedBy` must be ran twice. Currently it is ran on the existing entity, but it also need to run on the entity to be written. Perhaps the additional check I introduced could be moved to `ApiController` since there are no trip-specific code (besides the comment).

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
binh-dam-ibigroup: It is. I'm just realizing I left a typo in the FIXME comment.

README.md
```diff
@@ -43,6 +44,27 @@ java -jar target/otp-middleware.jar configurations/default/env.yml
 
 TODO: Add Auth0 setup instructions.
 
+#### Auth0 Scope
+
+The requesting user type which determines the level of authorization is based on the scope provided as part of a user's bearer token.
```
binh-dam-ibigroup: Need some commas, or split the sentence, e.g.:
`The requesting user thype, which determines [...] , is based on the scope ...`

README.md
```diff
@@ -43,6 +44,27 @@ java -jar target/otp-middleware.jar configurations/default/env.yml
 
 TODO: Add Auth0 setup instructions.
 
+#### Auth0 Scope
+
+The requesting user type which determines the level of authorization is based on the scope provided as part of a user's bearer token. 
+The bearer token 'scope' claim must contain one of otp-user, api-user or admin-user for the user to be correctly matched to a user held in the database.
```
binh-dam-ibigroup: Put backticks around scope, api-user, otp-user, admin-user, e.g.  `` `api-user` ``.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
br648: My bad, finished.

README.md
br648: Updated.

README.md
br648: Reverted.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
landonreed: Yes, I think you're right: this probably belongs in ApiController.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
br648: I've removed the comment as it will be addressed when error-notifications PR is merged.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: ~It is happening in MOD UI too.~

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
```diff
@@ -108,6 +108,11 @@ public static boolean isAuthHeaderPresent(Request req) {
         return authHeader != null;
     }
 
+    public static boolean isScopeHeaderPresent(Request req) {
+        final String scope = req.headers("scope");
```
binh-dam-ibigroup: Non-blocking: There is some redundancy (this method is used only once, with a call to `req.headers("scope")` that immediately follows the call to this function. Similar remark applies to method `isAuthHeaderPresent` above.

binh-dam-ibigroup (commented):
See other formatting comments.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
br648: Have updated.

README.md
br648: Updated.

README.md
br648: Updated.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: Disregard my comments regarding admin-UI and MOD-UI, I needed to add the scopes to those.

src/main/java/org/opentripplanner/middleware/auth/Auth0Connection.java
br648: I've removed `isScopeHeaderPresent` and tidied up where is was called in `RequestingUser.createTestUser`. `isAuthHeaderPresent` is used in more than one place so have left this as is.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
binh-dam-ibigroup: This is the middleground I would use as a fix:
```java
        if (scope.contains(OtpUser.AUTH0_SCOPE)) {
            otpUser = (testing)
                ? new OtpUser()
                : Persistence.otpUsers.getOneFiltered(withAuth0UserId);
            return;
        }
        if (scope.contains(AdminUser.AUTH0_SCOPE)) {
            adminUser = (testing)
                ? new AdminUser()
                : Persistence.adminUsers.getOneFiltered(withAuth0UserId);
        }
        if (scope.contains(ApiUser.AUTH0_SCOPE)) {
            apiUser = (testing)
                ? new ApiUser()
                : Persistence.apiUsers.getOneFiltered(withAuth0UserId);
        }
        LOG.error(...)
```

br648: I have updated, but this goes against the idea of limiting request to a single user type.

binh-dam-ibigroup: Apologies, after discussion with @landonreed, please use his suggestion instead. The decision is to not support both `AdminUser` and `ApiUser` on the same Auth0 id.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
```diff
@@ -56,44 +56,38 @@ private void defineUserFromScope(String scope, boolean testing) {
...
-        if (Auth0Connection.isScopeHeaderPresent(req)) {
-            // If the scope header is provided, restrict the requesting user to the type of user defined in the scope.
-            scope = req.headers("scope");
-        }
-
+        // If the auth header has been provided get the Auth0 user id from it. This is different from normal
+        // operation as the parameter will only contain the Auth0 user id and not "Bearer token".
+        String auth0UserId = req.headers("Authorization");
+        // Restrict authorization based on the scope value provided.
+        String scope = req.headers("scope");
```
binh-dam-ibigroup, line 96: +1 better than I thought it would be.

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
landonreed: Let's use this:
```java
// Define only a single user according to the scope. Note: there is an edge case where a user logging in from
// the OTP Admin Dashboard may be both an AdminUser and ApiUser, but this code block will force their identity
// as an AdminUser.
// TODO: Consider consolidating the user scope fields into a single AbstractUser user field.
if (scope.contains(OtpUser.AUTH0_SCOPE)) {
    otpUser = (testing)
        ? new OtpUser()
        : Persistence.otpUsers.getOneFiltered(withAuth0UserId);
    return;
}
if (scope.contains(AdminUser.AUTH0_SCOPE)) {
    adminUser = (testing)
        ? new AdminUser()
        : Persistence.adminUsers.getOneFiltered(withAuth0UserId);
    // Only return at this point if an AdminUser is defined. If not, fall back on defining an ApiUser. If we
    // returned unconditionally, ApiUsers that log in via OTP Admin Dashboard would not be defined below (the
    // final conditional block would never execute.
    if (adminUser != null) return;
}
if (scope.contains(ApiUser.AUTH0_SCOPE)) {
    apiUser = (testing)
        ? new ApiUser()
        : Persistence.apiUsers.getOneFiltered(withAuth0UserId);
    return;
}
LOG.error("No user type for scope {} is available", scope);
```

binh-dam-ibigroup (changes requested):
See https://github.com/ibi-group/otp-middleware/pull/120/commits/baf62e9ce8a0538c43e219f4f0e3262d3d9d9da3#r539588326. Sorry about the confusion.

binh-dam-ibigroup:
> See [baf62e9#r539588326](https://github.com/ibi-group/otp-middleware/commit/baf62e9ce8a0538c43e219f4f0e3262d3d9d9da3#r539588326). Sorry about the confusion.

Also @br648, could you update this PR's description to say it fixes #94, please?

src/main/java/org/opentripplanner/middleware/auth/RequestingUser.java
br648: Updated.

landonreed (approved):

binh-dam-ibigroup (approved):
Good work, LGTM!

PR #125, opened 2020-12-16
Validate YAML config

### LINKED ISSUE #123: Validate YAML config file on startup
There are a number of places where config values are required, but we don't find out they're missing until a code path is invoked. Instead we should validate the YAML when we load it to make sure required fields are all present (and exit the application with an error if they are not).

A good approach might be to use https://github.com/java-json-tools/json-schema-validator to validate the JSON against a schema file when we load the config in startup.

A PR that implements might also want to remove places where we throw errors in the application if config values are missing (because this should no longer be possible).

br648:
### Checklist

### Description

This PR enhances the otp-middleware by validating the env config against a JSON schema. Validation takes place as part of the config load on start-up. The otp-middleware will fail to start if validation fails.

This PR fixes #123

### New Optional Configuration Parameter

VALIDATE_ENVIRONMENT_CONFIG - If set to true validate the environment configuration (env.yml) against the environment schema (env.schema.json).


src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -53,6 +58,40 @@ public static void loadConfig(String[] args) throws IOException {
...
+     * does not match the schema or an exception is thrown, exit the application.
+     */
+    private static void validateConfig() throws IOException {
+        if ("false".equals(getConfigPropertyAsText("VALIDATE_ENVIRONMENT_CONFIG", "true"))) {
+            LOG.warn("Environment configuration schema validation disabled.");
+            return;
+        }
+        if (envConfig == null) {
+            LOG.error("Environment configuration not available to validate!");
+            System.exit(0);
```
landonreed: I think `0` is typically a success/non-error code (maybe we could just use `1` throughout this file). Also, it might be simpler to surround this all in a try/catch and throw an IllegalArgumentException (with each error message you've defined) each place the `exit` method is invoked. Then, in the catch block we could call System#exit once and finally.

br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -18,6 +22,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(ConfigUtils.class);
 
     public static final String DEFAULT_ENV = "configurations/default/env.yml";
+    public static final String DEFAULT_ENV_SCHEMA = "configurations/default/env.schema.json";
```
landonreed: Nit: This can probably just be called `ENV_SCHEMA` since there should only ever be one of them.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -53,6 +58,38 @@ public static void loadConfig(String[] args) throws IOException {
...
+    private static void validateConfig() {
+        if ("false".equals(getConfigPropertyAsText("VALIDATE_ENVIRONMENT_CONFIG", "true"))) {
+            LOG.warn("Environment configuration schema validation disabled.");
+            return;
+        }
+        try {
+            if (envConfig == null) {
+                throw new IllegalArgumentException("Environment configuration not available to validate!");
+            }
+            FileInputStream envSchemaStream = new FileInputStream(new File(DEFAULT_ENV_SCHEMA));
```
landonreed: It looks like here (and in some other spots in this file), creating the FileInputStream with a File arg is redundant (I think we can just pass a String).

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -53,6 +58,38 @@ public static void loadConfig(String[] args) throws IOException {
...
+            }
+            ProcessingReport report = JsonSchemaFactory
+                    .byDefault()
+                    .getValidator()
+                    .validate(envSchema, envConfig);
+            if (!report.isSuccess()) {
+                throw new IllegalArgumentException(report.toString());
+            }
+        } catch (IOException | IllegalArgumentException | ProcessingException e) {
+            LOG.error("Unable to validate environment configuration.", e);
```
landonreed: Here and elsewhere in the log statements, I think we should be more explicit about what we're validating, just so that a user of the software isn't scratching their head about the somewhat vague reference to "environment". In other words, we should just say `env.yml` instead of `environment configuration`. For example: "Unable to validate config file (env.yml)". (Similar to how we refer to it in the loadConfig method)

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
```diff
@@ -154,9 +154,6 @@ public static ZoneId getSystemZoneId() {
      */
     public static ZoneId getOtpZoneId() {
         String otpTzId = getConfigPropertyAsText("OTP_TIMEZONE");
-        if (otpTzId == null) {
-            throw new RuntimeException("OTP_TIMEZONE is not defined in config!");
-        }
```
landonreed, line 159: Might be good to check where the required args are referenced and remove these similar checks. For example, `BugsnagDispatcher#makeEventDataRequest`.

landonreed (changes requested):
Looks good! Just a couple of items. Also, I had the thought that perhaps we could auto-generate a table from the schema for inclusion in the README so that it automatically stays up-to-date. @evansiroky did a similar thing for otp-runner here: https://github.com/ibi-group/otp-runner/blob/e8f33f51323ecbcaa7798e411d3db495bef66961/scripts/readme-utils.js#L4-L10

Not sure how cumbersome that would be to do in this project, but might be valuable to look into.

landonreed:
I just realized that this might also need some changes now that #111 is merged in given that some of the config props might be different from what's specified in this PR.

evansiroky (changes requested):
- **this is the only blocking thing:** When running tests locally (without E2E enabled), there are missing variables in the `configurations/test/env.yml`.
  - One of these missing variables is commented out intentionally... probably should find a way to make non-e2e tests work by either not requiring the field or making the test environment work without the field
- I can now see that the tests are passing in CI because environment variables are being read in instead of config values. So it seems that this new config validation doesn't seem to get hit during testing on CI. For another example, see datatools-server which will build a config file during runtime and then startup the server with this config file. See https://github.com/ibi-group/datatools-server/blob/adb0d27447a94d58a32cda4b767c743dc2cd101d/src/test/java/com/conveyal/datatools/DatatoolsTest.java#L65-L156. This would probably involve a larger refactor, so I don't think I'd request a change like this at this time.
- It's probably out of scope and maybe just a random idea I have, but maybe it's possible to change up the Config so all the config is imported at initialization into final fields so that other parts of the application don't have to keep attempting to get certain config values from the raw JsonNode.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated to just pass a String.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
br648: @landonreed I can't see what needs to be updated in `BugsnagDispatcher#makeEventDataRequest` which is a concern as I'll probably miss other similar updates (I have updated more obvious required args references in other classes/methods).

configurations/default/env.schema.json
```diff
@@ -0,0 +1,172 @@
...
+    },
+    "AUTH0_DOMAIN": {
+      "type": "string",
+      "default": "test.auth0.com",
+      "description": "Auth0 tenant URL."
+    },
+    "AWS_PROFILE": {
+      "type": "string",
+      "default": "default",
+      "description": "AWS credentials"
```
binh-dam-ibigroup: "AWS profile for credentials"

configurations/default/env.schema.json
```diff
@@ -0,0 +1,172 @@
...
+    },
+    "OTP_API_ROOT": {
+      "type": "string",
+      "default": "http://otp-server.example.com/otp",
+      "description": "The URL of an operational OTP server."
+    },
+    "OTP_PLAN_ENDPOINT": {
+      "type": "string",
+      "default": "/routers/default/plan",
+      "description": "The OTP server's endpoint for plan."
```
binh-dam-ibigroup: "The path to the OTP server trip planning endpoint."

configurations/default/env.schema.json
```diff
@@ -0,0 +1,172 @@
...
+    "AUTH0_API_CLIENT",
+    "AUTH0_API_SECRET",
+    "BUGSNAG_API_KEY",
+    "BUGSNAG_ORGANIZATION",
+    "BUGSNAG_PROJECT_NOTIFIER_API_KEY",
+    "DEFAULT_USAGE_PLAN_ID",
+    "OTP_API_ROOT",
+    "OTP_TIMEZONE",
+    "MONGO_DB_NAME",
+    "VALIDATE_ENVIRONMENT_CONFIG"
```
binh-dam-ibigroup: Is this line needed? It is redundant with `getConfigPropertyAsText("VALIDATE_ENVIRONMENT_CONFIG", "true")` in `ConfigUtils#validateConfig`.

Also, This is producing this message which seems to contradict itself:
```log
WARN  2020-12-21T09:51:43.878 [main] o.o.middleware.utils.ConfigUtils(ConfigUtils.java:169)  Config property VALIDATE_ENVIRONMENT_CONFIG not found. Using defaultValue: true.
ERROR 2020-12-21T09:51:44.133 [main] o.o.middleware.utils.ConfigUtils(ConfigUtils.java:90)  Unable to validate environment configuration.
java.lang.IllegalArgumentException: com.github.fge.jsonschema.core.report.ListProcessingReport: failure
--- BEGIN MESSAGES ---
error: object has missing required properties (["VALIDATE_ENVIRONMENT_CONFIG"])

```

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
binh-dam-ibigroup: I think this file should be embedded as a resource (we don't want implementers to modify it).

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -53,6 +58,38 @@ public static void loadConfig(String[] args) throws IOException {
...
+            JsonNode envSchema = yamlMapper.readTree(envSchemaStream);
+            if (envSchema == null) {
+                throw new IllegalArgumentException("Environment configuration schema not available.");
+            }
+            ProcessingReport report = JsonSchemaFactory
+                    .byDefault()
+                    .getValidator()
+                    .validate(envSchema, envConfig);
+            if (!report.isSuccess()) {
+                throw new IllegalArgumentException(report.toString());
```
binh-dam-ibigroup, line 126: If a parameter is missing in `env.yml`, Is it possible to not display the list of required parameters?

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -1,6 +1,10 @@
 package org.opentripplanner.middleware.utils;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.github.fge.jsonschema.core.exceptions.ProcessingException;
+import com.github.fge.jsonschema.core.report.ProcessingReport;
+import com.github.fge.jsonschema.main.JsonSchemaFactory;
+import com.github.fge.jsonschema.main.JsonValidator;
```
binh-dam-ibigroup: This import shows as unused.

binh-dam-ibigroup (changes requested):
Looks good. See my suggestions, my main requests are:
- Make the schema file an embedded resource (or if there is a case against that let me know)
- Exclude `VALIDATE_ENVIRONMENT_CONFIG` from the required list (the code sets a default value already).

I also like what @evansiroky said below but this is also non-blocking:
> * It's probably out of scope and maybe just a random idea I have, but maybe it's possible to change up the Config so all the config is imported at initialization into final fields so that other parts of the application don't have to keep attempting to get certain config values from the raw JsonNode.

binh-dam-ibigroup:
Also @br648, did you push your latest changes?

configurations/default/env.schema.json
br648: Updated.

configurations/default/env.schema.json
br648: Updated.

configurations/default/env.schema.json
br648: I have removed it from the schema's required list.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Removed.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: @binh-dam-ibigroup I'm sure it is possble. Could you explain why you think it should be removed please?

binh-dam-ibigroup: That would simplify the error message IMO.

br648: ok, unless you feel really strongly about this I'm going to leave it as is. The change would require constucting my own report instead of using the default one.

landonreed: I agree that a custom report is not worth building at this time.

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
```diff
@@ -0,0 +1,103 @@
...
+    public static String generateEnvSchemaValuesContent() throws IOException {
+        FileInputStream envSchemaStream = new FileInputStream(ConfigUtils.DEFAULT_ENV_SCHEMA);
+        JsonNode envSchema = yamlMapper.readTree(envSchemaStream);
+        if (envSchema == null) {
+            throw new IllegalArgumentException("env.schema.json not available to update README.md.");
+        }
+        StringBuilder tableData = new StringBuilder("| Key | Type | Required | Default | Description |");
+        tableData.append(System.lineSeparator())
+                .append("| --- | --- | --- | --- | --- |")
+                .append(System.lineSeparator());
```
landonreed: Here and elsewhere (`ReadMeEnvSchemaValuesTest.class`) new lines for continued expressions are indented by two tab widths rather than one (which is the standard in the application). Might need to update your IDE settings (Editor > Code Style > Java > Continuation Indent: `4`).

src/test/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesTest.java
```diff
@@ -0,0 +1,45 @@
...
+     * changes. If there are changes to the env.schema.json file, run {@link ReadMeEnvSchemaValuesUpdater#main} to
+     * update the env.schema.json values section of {@link ReadMeEnvSchemaValuesUpdater#LATEST_README_FILE}.
+     */
+    @Test
+    public void envSchemaValuesAreUpToDate() throws IOException {
+        String expectedEnvSchemaValues = ReadMeEnvSchemaValuesUpdater
+                .generateEnvSchemaValuesContent()
+                .trim();
+        Path readMeFile = new File(LATEST_README_FILE).toPath();
+        String readMeContent = new String(Files.readAllBytes(readMeFile));
```
landonreed: `FileUtils.getFileContents` could be reused here I think. Perhaps this is an artifact of the copying code from the other test class (in which case maybe we should replace whatever is used there as well).

src/test/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesTest.java
```diff
@@ -0,0 +1,45 @@
...
+        int envSchemaJsonTitleIndex = readMeContent.indexOf(ENV_SCHEMA_JSON_TITLE) + ENV_SCHEMA_JSON_TITLE.length();
```
landonreed: Maybe a better name would be `endOfEnvSchemaJsonTitle`.

src/main/resources/env.schema.json
```diff
@@ -0,0 +1,195 @@
...
+    },
+    "TWILIO_AUTH_TOKEN": {
+      "type": "string",
+      "default": "your-auth-token",
+      "description": "Twilio settings available at: https://twilio.com/user/account"
+    },
+    "VALIDATE_ENVIRONMENT_CONFIG": {
+      "type": "boolean",
+      "default": "true",
+      "description": "If set to true the env.yml file is validated against this schema."
```
landonreed: Maybe since this is optional a better description would be "If set to false, the validation of the env.yml file against this schema will be skipped."

src/main/resources/env.schema.json
```diff
@@ -0,0 +1,195 @@
...
+      "type": "integer",
+      "default": "14",
+      "description": "Specifies how far in the past events should be retrieved."
+    },
+    "DEFAULT_USAGE_PLAN_ID": {
+      "type": "string",
+      "default": "usage-plan-id",
+      "description": "AWS API gateway default usage plan used when creating API keys for API users."
+    },
+    "OTP_ADMIN_DASHBOARD_FROM_EMAIL": {
```
landonreed, line 143: I think you've updated most of the config fields since dev was merged in, but I think we're still missing the optional `MONITORED_COMPONENTS`.

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
landonreed: Oh, I think it disappeared with the merge of dev. I think we're probably ok on this front.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -24,11 +24,7 @@
      */
     public static void initializeBugsnagErrorReporting() {
         String apiKey = getConfigPropertyAsText("BUGSNAG_PROJECT_NOTIFIER_API_KEY");
-        if (apiKey != null) {
-            bugsnag = new Bugsnag(apiKey);
-        } else {
-            LOG.warn("Bugsnag project notifier API key not available. Bugsnag error reporting disabled.");
-        }
+        bugsnag = new Bugsnag(apiKey);
```
landonreed: Yea, so for this one, I think we just make the API key optional. Apologies for making the original rec to to make it required.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
```diff
@@ -29,10 +29,6 @@
      * Schedule each Bugsnag job based on the delay configuration parameters
      */
     public static void initialize() {
-        if (BugsnagDispatcher.BUGSNAG_ORGANIZATION == null) {
-            LOG.error("WARNING: Bugsnag organization is not available. Cannot schedule bugsnag jobs.");
-            return;
-        }
```
landonreed: I think we may want to keep this optional actually. I can see some cases we don't want Bugsnag running on middleware (e.g., an instance that is just designed to provide extra features to OTP, but not all of the error event stuff).

landonreed (changes requested):
Sorry, a few more changes. I think one of my comments about removing null checks ended up not being correct because it forces Bugsnag to be essential to the running of the application, but it should probably be optional.

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Updates the {@link #LATEST_README_FILE} file with the latest env.schema.json values. It is expected that the
+     * title for this section be named '### env.schema.json values' and be the last section in the README.md file. Any
+     * sections after this will be removed as part of this update process!
+     */
+    private static void updateEnvSchemaValues() throws IOException {
+        Path readMeFile = new File(LATEST_README_FILE).toPath();
+        String readMeContent = new String(Files.readAllBytes(readMeFile));
```
landonreed: `FileUtils#getFileContents`

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
```diff
@@ -0,0 +1,103 @@
...
+                    .stream()
+                    .anyMatch(propertyName::equals))
+                    ? "Required"
+                    : "Optional";
+            String defaultValue = (property.get("default") != null)
+                    ? property.get("default").toString()
+                    : "";
+            String description = (property.get("description") != null)
+                    ? property.get("description").toString()
+                    : "";
```
landonreed: Might be able to DRY out this code that handles null check for a given property (it's used for `type`, `default`, and `description`).

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
br648: Fixed.

src/test/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesTest.java
br648: Should have copied this part as well. Binh has used `Files.readString` so went with this.

src/test/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesTest.java
br648: Updated.

src/main/resources/env.schema.json
br648: Updated.

src/main/resources/env.schema.json
br648: yeah, I think that is the last one, updated.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
br648: Updated with `Files.readString` to match unit test.

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
br648: yeah good point, updated.

README.md
```diff
@@ -144,4 +144,45 @@ The special E2E client settings should be defined in `env.yml`:
...
+| DEFAULT_USAGE_PLAN_ID | string | Required | usage-plan-id | AWS API gateway default usage plan used when creating API keys for API users. |
+| MAXIMUM_PERMITTED_MONITORED_TRIPS | integer | Optional | 5 | The maximum number of saved monitored trips. |
+| MONGO_DB_NAME | string | Required | otp_middleware | The name of the OTP Middleware Mongo DB. |
+| MONGO_HOST | string | Optional | localhost:27017 | Mongo host address. |
+| MONGO_PASSWORD | string | Optional | (empty string) | Mongo DB password |
+| MONGO_PROTOCOL | string | Optional | mongodb | Mongo DB protocol |
+| MONGO_USER | string | Optional | (empty string | Mongo DB user name |
+| MONITORED_COMPONENTS | array | Optional | n/a | An array of monitored components. |
+| NOTIFICATION_FROM_EMAIL | string | Optional | noreply@email.com | The from email address used in notification emails |
+| NOTIFICATION_FROM_PHONE | string | Optional | +15551234 | The from phone number used in notification SMSs |
```
binh-dam-ibigroup: Phone number values such as `+15551234` are causing a validation error saying it found an integer while a string is expected. Can `integer or string` be specified as a type? Otherwise, the description should indicate that the phone number should be surrounded with quotes.

configurations/test/env.yml
```diff
@@ -46,3 +46,5 @@ MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
 #NOTIFICATION_FROM_EMAIL: noreply@email.com
 #NOTIFICATION_FROM_PHONE: +15551234
```
binh-dam-ibigroup, line 47: See my comment for this param in `README.md`

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -10,7 +10,7 @@
 
 /**
  * Bugsnag util for reporting errors to the project defined by the Bugsnag project notifier API key.
- *
+ * <p>
```
binh-dam-ibigroup: Remove this tag, please.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
```diff
@@ -29,6 +29,7 @@ public static void initializeBugsnagErrorReporting() {
         } else {
             LOG.warn("Bugsnag project notifier API key not available. Bugsnag error reporting disabled.");
         }
+
```
binh-dam-ibigroup: Remove this extra blank line.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -18,6 +25,18 @@
     private static final Logger LOG = LoggerFactory.getLogger(ConfigUtils.class);
 
     public static final String DEFAULT_ENV = "configurations/default/env.yml";
+    public static final String DEFAULT_ENV_SCHEMA = "src/main/resources/env.schema.json";
+    private static JsonNode ENV_SCHEMA = null;
+
+    static {
```
binh-dam-ibigroup, line 32: I would move the constructor after all the field declarations.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -42,17 +61,75 @@ public static boolean getBooleanEnvVar(String var) {
...
-            envConfigStream = new FileInputStream(new File(args[0]));
+            envConfig = yamlMapper.readTree(new FileInputStream(args[0]));
+        }
+        validateConfig();
+    }
+
+    /**
+     * Construct a config file from environment variables. If running in CI, the only way to set up the config is via
+     * environment variables. This allows us to read in these variables from the CI environment and validate it against
+     * the schema file.
```
binh-dam-ibigroup: Suggestion: remove first sentence.

Also: "and validate **them** against ..."

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -154,10 +222,10 @@ public static int getConfigPropertyAsInt(String name, int defaultValue) {
      * (modified from https://stackoverflow.com/questions/14189162/get-name-of-running-jar-or-exe#19045510).
      * TODO: Extract git properties from JAR, see
      * https://github.com/ibi-group/datatools-server/blob/9f74b821cf351efcdaf7c9c93a3ae8b694d3c3b1/src/main/java/com/conveyal/datatools/manager/DataManager.java#L181-L212.
-     *
+     * <p>
      * "/" is used instead of File.separator, see
      * https://stackoverflow.com/questions/24749007/how-to-use-file-separator-for-a-jar-file-resource/24749976#24749976
-     *
+     * <p>
```
binh-dam-ibigroup: Remove tags.

src/main/resources/env.schema.json
```diff
@@ -0,0 +1,225 @@
...
+      "description": "Mongo DB password"
+    },
+    "MONGO_PROTOCOL": {
+      "type": "string",
+      "default": "mongodb",
+      "description": "Mongo DB protocol"
+    },
+    "MONGO_USER": {
+      "type": "string",
+      "default": "(empty string",
```
binh-dam-ibigroup: Typo: extra or missing bracket.

src/test/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesTest.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class ReadMeEnvSchemaValuesTest {
+
+    /**
+     * Verify that {@link ReadMeEnvSchemaValuesUpdater#LATEST_README_FILE} matches the latest env.schema.json file
+     * changes. If there are changes to the env.schema.json file, run {@link ReadMeEnvSchemaValuesUpdater#main} to
+     * update the env.schema.json values section of {@link ReadMeEnvSchemaValuesUpdater#LATEST_README_FILE}.
+     */
+    @Test
+    public void envSchemaValuesAreUpToDate() throws IOException {
```
binh-dam-ibigroup, line 26: Love this

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -120,11 +193,6 @@ public static String getConfigPropertyAsText(String name) {
      * if the config value is not defined (null).
      */
     public static String getConfigPropertyAsText(String name, String defaultValue) {
```
binh-dam-ibigroup, line 194: There are a number of calls to `getConfigPropertyAs[Text|Int]` that specify a default value. Can we get rid of the default value passed to most of these calls, as default values are passed from the schema now? Removing the default value passed when calling these functions would also prevent that and the schema values to contradict each other.

binh-dam-ibigroup (commented):
Some minor changes as indicated, and it looks good overall!
Before I give it a nod, please advise regarding:
- whether to surround the notification phone number with quotes,
- passing default values to `getConfigPropertyAs[Text|Int]` where we already set a default value through the schema.

README.md
br648: Only one type can be specified. I've updated the description as suggested.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Grrrr, I've updated Intellij settings.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagReporter.java
br648: Done.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

src/main/resources/env.schema.json
br648: Updated.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Default values must continue to be provided as the schema default values are not applied and are simply examples. I've noted this in the schema description.

br648:
> 
> 
> Some minor changes as indicated, and it looks good overall!
> Before I give it a nod, please advise regarding:
> 
>     * whether to surround the notification phone number with quotes,
> 
>     * passing default values to `getConfigPropertyAs[Text|Int]` where we already set a default value through the schema.

Updated to surround the notification phone number with quotes so it is correctly parsed.
Schema default values are examples only and are not applied at runtime. Passing default is still required. However, I could update to extract and use the schema default values? @binh-dam-ibigroup @landonreed thoughts on this before I potentiallly implement?

binh-dam-ibigroup:
> [...] However, I could update to extract and use the schema default values? @binh-dam-ibigroup @landonreed thoughts on this before I potentiallly implement?

I really like the idea, but let's hear from the others first.

binh-dam-ibigroup:
@br648 Could you add in the description that `This PR fixes #123` please?

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
landonreed: There's a bit of conflation here between example and default values. Will things break if we remove all of the "example" values (and just move those examples to the description text instead)?

br648: Following the tech call is was decided that the default parameter would be replaced with the example parameter (migrating to draft 7 to allow this) and that specifying a default value in code would remain.

src/main/resources/env.schema.json
```diff
@@ -1,120 +1,120 @@
...
     "MONGO_PROTOCOL": {
       "type": "string",
-      "default": "mongodb",
+      "examples": ["mongodb"],
       "description": "Mongo DB protocol"
     },
     "MONGO_USER": {
       "type": "string",
-      "default": "(empty string)",
+      "examples": ["(empty string)"],
```
landonreed: Might be good to provide better example for this and password (simply `username` and `password` would be fine).

src/main/resources/env.schema.json
```diff
@@ -127,87 +127,87 @@
...
     "OTP_ADMIN_DASHBOARD_NAME": {
       "type": "string",
-      "default": "OTP Admin Dashboard",
+      "examples": ["OTP Admin Dashboard"],
       "description": "Config setting for linking to the OTP Admin Dashboard."
     },
     "OTP_ADMIN_DASHBOARD_URL": {
       "type": "string",
-      "default": "https://otp-admin-ui.now.sh",
+      "examples": ["https://otp-admin-ui.now.sh"],
```
landonreed: Let's use `https://admin.example.com`

src/main/resources/env.schema.json
```diff
@@ -1,120 +1,120 @@
...
     "BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_MINUTES": {
       "type": "integer",
-      "default": "5",
+      "examples": ["5"],
       "description": "Bugsnag event request frequency."
     },
     "BUGSNAG_ORGANIZATION": {
       "type": "string",
-      "default": "bugsnag-organization-id",
+      "examples": ["bugsnag-organization-id"],
```
landonreed: Might be good to use a sample that looks like the ID needed. E.g., `123e4567e89b12d3a4564266`

src/main/resources/env.schema.json
```diff
@@ -1,120 +1,120 @@
...
       "description": "A valid Bugsnag organization id."
     },
     "BUGSNAG_PROJECT_NOTIFIER_API_KEY": {
       "type": "string",
-      "default": "bugsnag-project-api-key-id",
+      "examples": ["bugsnag-project-api-key-id"],
```
landonreed: Might be good to use a sample that looks like the ID needed. E.g., `123e4567e89b12d3a4564266`

src/main/resources/env.schema.json
```diff
@@ -1,120 +1,120 @@
...
     "AWS_API_STAGE": {
       "type": "string",
-      "default": "stage-name",
+      "examples": ["stage-name"],
       "description": "For generating the swagger document at runtime. Can be null, however that will prevent tools such as swagger-UI from submitting test requests to the API server."
     },
     "BUGSNAG_API_KEY": {
       "type": "string",
-      "default": "bugsnag-api-key",
+      "examples": ["bugsnag-api-key"],
```
landonreed: Might be good to use a sample that looks like the ID needed. E.g., `123e4567e89b12d3a456426614174000`

landonreed (approved):
Looks good! Just noticed a couple of example values that could be tweaked.

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
```diff
@@ -60,7 +60,7 @@ public static String generateEnvSchemaValuesContent() throws IOException {
                 .anyMatch(propertyName::equals))
                 ? "Required"
                 : "Optional";
-            String defaultValue = getJSONPropertyFieldAsString(property, "default");
+            String examples = getJSONPropertyFieldAsString(property, "examples");
```
binh-dam-ibigroup: Just curious, would it make sense to resolve `examples` to `String[]` or `List<String>`?

src/main/java/org/opentripplanner/middleware/docs/ReadMeEnvSchemaValuesUpdater.java
```diff
@@ -69,7 +69,9 @@ public static String generateEnvSchemaValuesContent() throws IOException {
                 .append(" | ")
                 .append(required)
                 .append(" | ")
-                .append(defaultValue.replaceAll("\"", ""))
+                .append(examples.replaceAll("\"", "")
```
binh-dam-ibigroup: If `examples` was an array or list, you could simply write `Strings.join(", ", examples)`.

binh-dam-ibigroup (approved):
Just a suggestion to resolve `examples` to an array or list, good to go overall.

evansiroky (approved):

PR #129, opened 2021-01-06
Replace TravisCI with GitHub Actions

evansiroky:
### Checklist

### Description

This PR replaces the TravisCI workflow with GitHub Actions.


br648 (commented):
At a high level this looks good to me. Although I don't see an issue with the updates it would be nice to get an understanding of how this all ties together. Perhaps this could be covered in the Tech call?

scripts/prep-for-deploy-from-github-actions.sh
```diff
@@ -0,0 +1,20 @@
+export BRANCH=$(
+  if [ -n "$GITHUB_HEAD_REF" ]; then
```
binh-dam-ibigroup, line 2: This is breaking the continuous deployment script. The branch name seems to come out in the form `refs_heads_<branch-name>` (e.g. `refs_heads_github-actions`). The continuous deployment script is probing the folder with just the branch name instead.

.github/workflows/maven-ci.yml
```diff
@@ -0,0 +1,71 @@
+name: Maven CI
+
+on: [push, pull_request]
```
binh-dam-ibigroup, line 3: Unlike OTP-RR, we DO want to keep the push event for deploying and testing branch code.

binh-dam-ibigroup (changes requested):
It looks like branch names are prepended with some prefix, which is resulting in folder names different than the one the continuous deployment script is probing. Please advise.

scripts/prep-for-deploy-from-github-actions.sh
evansiroky: Thanks for catching that, fixed in f148f6d.

evansiroky:
Ready for review again. I went ahead and disabled uploading the jar to s3 on pull request actions.

binh-dam-ibigroup (approved):
Seems to work well!. I went ahead and removed the requirement on Travis checks for the repo, so this should be good to go.

br648 (approved):
LGTM

scripts/add-aws-credentials.sh
```diff
@@ -12,5 +12,6 @@ cat > ~/.aws/credentials << EOL
 [default]
```
landonreed, line 11: Nit: this file refs Travis CI in two places above (lines 8 and 3). Since the PR replaces other mentions of Travis, we could do that here also.

landonreed (approved):
Looks good. One minor comment about removing refs to Travis since it is no longer used.

scripts/add-aws-credentials.sh
evansiroky: Fixed in 8162c58

PR #130, opened 2021-01-08
Load schema as resource

br648:
### Checklist

### Description

Loading the JSON schema as a file fails once deployed. This PR addresses that by loading the JSON schema as a resource instead.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
+++ b/src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
@@ -32,7 +32,7 @@ public class ConfigUtils {
     static {
         // Load in env.yml schema file statically so that it is available for populating properties when running CI.
         try {
-            ENV_SCHEMA = yamlMapper.readTree(ConfigUtils.class.getClassLoader().getResourceAsStream(DEFAULT_ENV_SCHEMA));
+            ENV_SCHEMA = loadEnvSchema();
         } catch (IOException e) {
             LOG.error("Could not read env.yml config schema", e);
             System.exit(1);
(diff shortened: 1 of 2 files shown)


src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -25,14 +25,14 @@
...
-    public static final String DEFAULT_ENV_SCHEMA = "src/main/resources/env.schema.json";
+    public static final String DEFAULT_ENV_SCHEMA = "env.schema.json";
     private static JsonNode ENV_SCHEMA = null;
     private static final String JAR_PREFIX = "otp-middleware-";
 
     static {
         // Load in env.yml schema file statically so that it is available for populating properties when running CI.
         try {
-            ENV_SCHEMA = yamlMapper.readTree(new FileInputStream(DEFAULT_ENV_SCHEMA));
+            ENV_SCHEMA = yamlMapper.readTree(ConfigUtils.class.getClassLoader().getResourceAsStream(DEFAULT_ENV_SCHEMA));
```
landonreed: I think this load/read stuff should go into its own method (it's used here and in the Updater class) and the javadoc should state that it must be loaded as a resource. Otherwise, we lose any context as to why this change was made and run the risk of someone changing it back to a `new FileInputStream` because it looks nicer.

landonreed (changes requested):
See comment about adding a method with comments about why this approach is being used.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
br648: Updated.

binh-dam-ibigroup (approved):
LGTM

landonreed (approved):

PR #135, opened 2021-02-11
Http client refactor

br648:
### Checklist

- [X] The description lists all applicable issues this PR seeks to resolve

### Description

This PR seeks to replace the use of `java.net.http.HttpClient` with `org.apache.http.client.HttpClient`. This is due to exceptions being encountered with `java.net.http.HttpClient` which are out of our control. These are:

`CLOSE_WAIT` which is experienced when many `java.net.http.HttpClient`s are used from a static context. This can be fixed by using a single instance of `java.net.http.HttpClient`, but that then leads to the other issue...

`GOAWAY` which is experienced when the maximum number of streams is met for a single connection.

In both cases it appears that the connections are not closed. To compound the issue, `java.net.http.HttpClient` does not provide a method to explicitly close connections.


src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagDispatcher.java
```diff
@@ -126,14 +128,23 @@ public static BugsnagEventRequest makeEventDataRequest(String eventDataRequestId
...
+                BUGSNAG_HEADERS,
+                create ? EVENT_REQUEST_FILTER : null,
+                true
+            );
+            return JsonUtils.getPOJOFromHttpBody(response, BugsnagEventRequest.class);
+        } catch (HttpTimeoutException e) {
+            LOG.error("Request to {} timed out after {} seconds.", eventDataRequestUri, CONNECTION_TIMEOUT_IN_SECONDS, e);
+        } catch (IOException e) {
+            BugsnagReporter.reportErrorToBugsnag("Error requesting data from URI", eventDataRequestUri, e);
+        }
```
landonreed: Any reason not to have the `httpRequestRawResponse` handle the IOException similar to Data Tools:
https://github.com/ibi-group/datatools-server/blob/6829065126adbd7689614ddf5d52b5efb736ed86/src/main/java/com/conveyal/datatools/manager/utils/HttpUtils.java#L82-L88

It looks like we're reporting the Bugsnag error in multiple places, when maybe we should just do this once in the util class. I know we've gone back and forth a bit on utils like this (e.g. JSON parsing), so I'm just raising this to make sure we have a consistent approach.

landonreed (changes requested):
Looks good. Just one question raised re: IOException handling.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagDispatcher.java
br648: @landonreed this also raises another issue, we should probably be using the CloseableHttpClient and try with resources to guarantee the connection is closed. However, this will have an impact on calling services. As the connection will be closed the response body would need to be extracted here and provided with the status code in a bespoke class (probably using a Response Handler).

The thinking behind reporting Bugsnag errors in multiple places was to log any time outs, but not to report them to Bugsnag. Perhaps this is better off in HttpUtils and is used across the board?

landonreed: Yep, that all sounds good:
- using closeablehttpclient in try/catch with resources
- adding new class for response handler
- handle logging timeouts/bugsnag logging in httputils

br648: This approach has now been implemented.

binh-dam-ibigroup (approved):
I love how clean this refactor is. Very nice.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -111,18 +136,21 @@ public static String httpRequest(URI uri, int connectionTimeout, HttpMethod meth
...
-        HttpRequest request = httpRequestBuilder.build();
+        CloseableHttpClient httpClient = allowRedirects
+            ? HttpClientBuilder.create().build()
+            : HttpClientBuilder.create().disableRedirectHandling().build();
 
-        try {
-            return client.send(request, java.net.http.HttpResponse.BodyHandlers.ofString());
-        } catch (InterruptedException | IOException e) {
+        try  {
+            return httpClient.execute(httpUriRequest, new HttpResponseHandler());
```
landonreed: Is it possible to pass the `httpUriRequest` to the `HttpResponseHandler` constructor here? That way we could extract the URI directly within `HttpResponseHandler` (to keep from doing it in OtpDispatcherResponse) and that design may help with extracting other information down the road (if needed).

landonreed (changes requested):
Looks great. Just one question about constructing HttpResponseHandler.

src/main/java/org/opentripplanner/middleware/utils/HttpResponseHandler.java
```diff
@@ -0,0 +1,35 @@
+package org.opentripplanner.middleware.utils;
+
+import org.apache.http.HttpResponse;
+import org.apache.http.client.ResponseHandler;
+import org.apache.http.util.EntityUtils;
```
evansiroky: Is this a transitive dependency? Maybe the Apache HttpCore should be included as a dependency in the pom.xml.

src/main/java/org/opentripplanner/middleware/utils/HttpResponseHandler.java
```diff
@@ -0,0 +1,35 @@
...
+    private static String getResponseBodyAsString(HttpResponse response) {
+        String responseBody = null;
+        try {
+            responseBody = EntityUtils.toString(response.getEntity());
+            EntityUtils.consume(response.getEntity());
+        } catch (IOException e) {
+            LOG.error("Unable to get response body", e);
+        }
+        return responseBody;
+    }
```
evansiroky: Is it required to parse the response body and consume it here, or can that be done as an instance method in `HttpResponseValues`? The reason I ask is that sometimes with large response bodies, it may be more efficient to read the response body as a stream instead of first reading all of it into a string and then doing further processing.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
```diff
@@ -90,18 +90,18 @@ private static boolean isResponseOk(HttpResponse<String> response) {
     /**
      * Utility method to parse generic object from HTTP response.
      */
-    public static <T> T getPOJOFromHttpBody(HttpResponse<String> response, Class<T> clazz) {
+    public static <T> T getPOJOFromHttpBody(HttpResponseValues response, Class<T> clazz) {
         return isResponseOk(response)
-            ? getPOJOFromJSON(response.body(), clazz)
+            ? getPOJOFromJSON(response.responseBody, clazz)
```
evansiroky, line 95: Like right here, this calls `getPOJOFromJSON` which merely calls `mapper.readValue(json, clazz)`. The `mapper` can also take in an InputStream, so technically it'd be possible to read the response body just once using the mapper instead of first reading it into a string and then parsing the string.

evansiroky (approved):
Probably fine for now especially if it helps reduce the amount of HTTP connection problems we're having. My only concern is about potentially double-reading the response bodies instead of having an ObjectMapper read from the HTTP response stream. I'm not sure if it even adds much performance improvements, but maybe it'd be good to test out different implementations or reading massive OTP responses just in case.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
br648: @landonreed that is not possible. The `httpUriRequest` is required by `httpClient.execute`. Also the connection is closed after this point so any attempts to down the response body would fail.

src/main/java/org/opentripplanner/middleware/utils/HttpResponseHandler.java
br648: Unfortunately not, the connection is closed after this point so any attempts to down the response body would fail.

src/main/java/org/opentripplanner/middleware/utils/JsonUtils.java
br648: The response body has to be completely consumed in the response handler. After this the connection will be closed.

src/main/java/org/opentripplanner/middleware/utils/HttpResponseHandler.java
br648: It is not. A number of dependencies include it as a sub-dependency. It is now a top level dependency.

landonreed:
@br648 see changes ini #137.

landonreed (approved):

PR #145, opened 2021-06-23
Raise exception instead of returning null after failed JSON > POJO parse

evansiroky:
### Checklist

### Description

Previously if a parse exception occurred, the code reported the exception to Bugsnag, but then returned `null`. Methods that called this method are mostly test methods, but some other non-test code that made this call would then not properly do a null check which could result in `NullPointerException`s. Therefore, this refactors the `JSONUtils#getPOJOFromJSON` method so that it throws a `JsonProcessingException`.

The affected methods that now need to catch this exception have been modified to better handle this exception and provide more relevant information to Bugsnag and avoid the possibility of NullPointerExceptions.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
@@ -4,6 +4,7 @@ import com.beerboy.ss.SparkSwagger;
 import com.beerboy.ss.descriptor.EndpointDescriptor;
 import com.beerboy.ss.descriptor.ParameterDescriptor;
 import com.beerboy.ss.rest.Endpoint;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.auth.Auth0Connection;
 import org.opentripplanner.middleware.auth.RequestingUser;
@@ -125,7 +126,14 @@ public class OtpRequestProcessor implements Endpoint {
         }
         // If the request path ends with the plan endpoint (e.g., '/plan' or '/default/plan'), process response.
         if (otpRequestPath.endsWith(OtpDispatcher.OTP_PLAN_ENDPOINT) && otpUser != null) {
-            handlePlanTripResponse(request, otpDispatcherResponse, otpUser);
+            if(!handlePlanTripResponse(request, otpDispatcherResponse, otpUser)) {
+                logMessageAndHalt(
+                    request,
+                    HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    "Failed to save trip history."
+                );
+                return null;
+            }
         }
         // provide response to requester as received from OTP server
         response.type(MediaType.APPLICATION_JSON);
@@ -135,10 +143,10 @@ public class OtpRequestProcessor implements Endpoint {
 
     /**
      * Process plan response from OTP. Store the response if consent is given. Handle the process and all exceptions
-     * seamlessly so as not to affect the response provided to the requester.
+     * seamlessly so as not to affect the response provided to the requester. Returns false if there was an error.
      */
-    private static void handlePlanTripResponse(Request request, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
-
+    private static boolean handlePlanTripResponse(Request request, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
+        boolean result = true;
         String batchId = request.queryParams("batchId");
         if (batchId == null) {
             //TODO place holder for now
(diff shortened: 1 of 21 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -135,10 +143,10 @@ private static String proxy(Request request, spark.Response response) {
 
     /**
      * Process plan response from OTP. Store the response if consent is given. Handle the process and all exceptions
-     * seamlessly so as not to affect the response provided to the requester.
+     * seamlessly so as not to affect the response provided to the requester. Returns false if there was an error.
```
landonreed: Might be better to just move that last sentence to an `@return`.

br648 (approved):
LGTM.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -181,7 +186,12 @@ public void checkExistence() {
             }
             // Send off each plan query to OTP.
             OtpDispatcherResponse response = OtpDispatcher.sendOtpPlanRequest(otpRequest);
-            TripPlan plan = response.getResponse().plan;
+            TripPlan plan = null;
+            try {
+                plan = response.getResponse().plan;
+            } catch (JsonProcessingException e) {
+                LOG.error("Encountered a parse exception for otpRequest {}", otpRequest, e);
```
landonreed:
```suggestion
                LOG.error("Could not parse plan response for otpRequest {}", otpRequest, e);
```

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -70,9 +77,15 @@ public void setResponse(OtpResponse response) {
...
         // Only include the plan response if requestUri.path ends with OTP_PLAN_ENDPOINT.
         // Without this check, we are sending valid responses from non-plan OTP endpoints to Bugsnag as errors.
-        String planResponse = requestUri.getPath().endsWith(OTP_PLAN_ENDPOINT)
-                ? ", response=" + getResponse()
-                : "";
+        String planResponse = null;
+        try {
+            planResponse = requestUri.getPath().endsWith(OTP_PLAN_ENDPOINT)
+                    ? ", response=" + getResponse()
+                    : "";
```
landonreed:
```suggestion
                ? ", response=" + getResponse()
                : "";
```

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -160,19 +162,31 @@ public boolean makeOTPRequestAndUpdateMatchingItinerary() {
...
+                null
+            );
+            return false;
+        }
+        OtpResponse otpResponse = null;
+        try {
+            otpResponse = otpDispatcherResponse.getResponse();
+        } catch (JsonProcessingException e) {
+            // don't report to Bugsnag since the getResponse method will already have reported to Bugsnag.
+            LOG.error("Unable to parse OTP response!", e);
```
landonreed, line 187: Does this mean we're reporting the error twice? (line 175 and in `getResponse`)? I suppose if that only happens if `getResponse` encounters a parse exception, this should be fine.

src/test/java/org/opentripplanner/middleware/testutils/ApiTestUtils.java
```diff
@@ -90,7 +90,7 @@
      * Attempt to get Auth0 token from Auth0 tenant for the given username (email) and scope. If the response from Auth0
      * is ok, extract the access token from the token holder response and return to caller.
      */
-    private static String getTestAuth0AccessToken(String username, String scope) {
+    private static String getTestAuth0AccessToken(String username, String scope) throws Exception {
```
landonreed, line 93: Do we want this to throw an exception or should we simply swallow the exception generated by:
```
TokenHolder token = JsonUtils.getPOJOFromJSON(response.responseBody, TokenHolder.class);
```
I suppose the effect of throwing the exception would be that the test would automatically fail? That seems like acceptable behavior to me, so perhaps we keep the exception. I will say though that the appearance of a generic exception on all of these methods rather than `JsonProcessingException` had me scratching my head for a bit.

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
```diff
@@ -92,8 +91,9 @@
     );
 
     /** Contains the verified itinerary set for a trip upon persisting. */
-    private static final Itinerary DEFAULT_ITINERARY =
-        OtpTestUtils.OTP_DISPATCHER_PLAN_RESPONSE.getResponse().plan.itineraries.get(0);
+    public static Itinerary getDefaultItinerary() throws Exception {
```
landonreed, line 94: Why not the more precise `JsonProcessingException`?

landonreed (approved):
Looks good. A couple of non-blocking items re: exception types.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcherResponse.java
```diff
@@ -70,9 +77,15 @@ public void setResponse(OtpResponse response) {
...
+        } catch (JsonProcessingException e) {
+            LOG.error("Encountered exception wile parsing OTP_PLAN_ENDPOINT response", e);
```
binh-dam-ibigroup: Typo
```suggestion
            LOG.error("Encountered exception while parsing OTP_PLAN_ENDPOINT response", e);
```

binh-dam-ibigroup (approved):
LGTM, but I'm wondering whether there should be a test against the failed JSON > POJO parsing?

src/test/java/org/opentripplanner/middleware/utils/ItineraryUtilsTest.java
evansiroky: Cause I was lazy and this is in testing code.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
evansiroky: No, line 175 is immediately followed by a return statement, so there wouldn't be double reporting since the getResponse call wouldn't occur as the method would've already have been returned from.

src/test/java/org/opentripplanner/middleware/testutils/ApiTestUtils.java
evansiroky: Yeah, the tests would just fail. I started putting up a bunch of JsonProcessingExceptions, but sometimes those got appended to other exceptions that it just seemed like adding a generic Exception would just catch any error and fail the test. Of course, the exceptions should be handled differently in non-test code, but this seemed appropriate to me to liberally add generic Exceptions being thrown from various test methods.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
evansiroky: Added in 19fb070

PR #147, opened 2021-07-16
Bugsnag improvements

evansiroky:
### Checklist

### Description

This PR adds a few things to help with error reporting:

- Adds the ability to customize Bugsnag reporting data such that it includes version, app type and release stage.
- Specifically catches the java.net.SocketTimeoutException and does not create a Bugsnag error because of this.
- Makes sure every possible place that ends up calling HttpUtils.httpRequestRawResponse is able to handle a null response from this method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,10 +156,12 @@ The special E2E client settings should be defined in `env.yml`:
 | AWS_API_SERVER | string | Optional | aws-api-id.execute-api.us-east-1.amazonaws.com | For generating the swagger document at runtime. Can be null, however that will prevent tools such as swagger-UI from submitting test requests to the API server. |
 | AWS_API_STAGE | string | Optional | stage-name | For generating the swagger document at runtime. Can be null, however that will prevent tools such as swagger-UI from submitting test requests to the API server. |
 | BUGSNAG_API_KEY | string | Required | 123e4567e89b12d3a4564266 | A valid Bugsnag authorization token. |
+| BUGSNAG_APP_TYPE | string | Optional | my-application | An identifier of the app type of this otp-middleware instance. This can be used to segment errors within the Bugsnag dashboard. |
 | BUGSNAG_EVENT_JOB_DELAY_IN_MINUTES | integer | Optional | 1 | Bugsnag event job frequency. |
 | BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_MINUTES | integer | Optional | 5 | Bugsnag event request frequency. |
 | BUGSNAG_ORGANIZATION | string | Optional | 123e4567e89b12d3a4564266 | A valid Bugsnag organization id. |
 | BUGSNAG_PROJECT_NOTIFIER_API_KEY | string | Optional | 123e4567e89b12d3a4564266 | A valid Bugsnag project API key. |
+| BUGSNAG_RELEASE_STAGE | string | Optional | staging | An identifier of the release stage of this otp-middleware instance. This can be used to segment errors within the Bugsnag dashboard. |
 | BUGSNAG_REPORTING_WINDOW_IN_DAYS | integer | Optional | 14 | Specifies how far in the past events should be retrieved. |
 | DEFAULT_USAGE_PLAN_ID | string | Required | 123e45 | AWS API gateway default usage plan used when creating API keys for API users. |
 | MAXIMUM_PERMITTED_MONITORED_TRIPS | integer | Optional | 5 | The maximum number of saved monitored trips. |
--- a/src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
+++ b/src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
@@ -20,6 +20,7 @@ import spark.Request;
 import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.http.HttpTimeoutException;
 import java.time.LocalDate;
@@ -147,7 +148,7 @@ public class HttpUtils {
         try  {
             // Extract required information from the response and return to caller. The connection is closed once complete.
             return httpClient.execute(httpUriRequest, new HttpResponseHandler(httpUriRequest));
-        } catch (HttpTimeoutException e) {
+        } catch (HttpTimeoutException | SocketTimeoutException e) {
             LOG.error("Request to {} timed out after {} seconds.", uri, timeoutInSeconds, e);
         } catch (IOException e) {
             BugsnagReporter.reportErrorToBugsnag("Error requesting data from URI", uri, e);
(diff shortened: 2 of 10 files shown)


miles-grant-ibigroup (approved):
Looks good to me given my experience with otp-middleware.

br648 (approved):
LGTM.

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -147,7 +148,7 @@ public static HttpResponseValues httpRequestRawResponse(URI uri, int timeoutInSe
         try  {
             // Extract required information from the response and return to caller. The connection is closed once complete.
             return httpClient.execute(httpUriRequest, new HttpResponseHandler(httpUriRequest));
-        } catch (HttpTimeoutException e) {
+        } catch (HttpTimeoutException | SocketTimeoutException e) {
```
landonreed, line 151: Do you have a proposal for how we are alerted about request timeouts? It feels kind of sketchy to just ignore timeouts, given that they alerted us to some major issues with our ATL implementation of OTP.

EDIT: It looks like ultimately this might end up triggering a bugsnag report via this line: 

https://github.com/ibi-group/otp-middleware/blob/19fb070e89bdb709862dcbe105f44bd4255ff1c8/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java#L124

But it might be good to make note of that either here, at OtpDispatcher.java line 94, or in both places. It's a little opaque otherwise, especially since we're weirdly choosing to log errors on IOExceptions? I don't fully understand why we would choose to log errors for IO but not timeouts (and we're effectively swallowing the timeout exception).

landonreed (changes requested):
I'm not fully on board with swallowing the timeout exceptions. I don't really know why we ever built that initial catch clause that avoids bugsnag reporting. This timeout report in bugsnag helped us identify an important issue with our OTP setup, why would we choose to disable this moving forward?

evansiroky:
> I'm not fully on board with swallowing the timeout exceptions. I don't really know why we ever built that initial catch clause that avoids bugsnag reporting. This timeout report in bugsnag helped us identify an important issue with our OTP setup, why would we choose to disable this moving forward?

I have some reasoning in my head that I didn't share here, sorry about that. My thinking is that Bugsnag error reports should only be sent for problems related to the code not working in an expected way given all reasonable implementations for fault tolerance. What this means here is that network problems are inevitable problems and should be tolerated and handled accordingly within this application, but tracked and monitored by something other than Bugsnag.

For example, In the event of a large-scale internet outage, our applications will of course be affected, but not due to some problem with our coding. It's the internet outage that is the problem and that happening is something that I would expect otp-middleware to gracefully handle. I'm definitely open to receiving reviews if there are some problems that might occur with this implementation that handles network timeouts. As can be seen with a few other pieces of code that I added in this PR, I believe this scenario is now tolerated in a way that the application should recover once a more reliable connection occurs. 

It would be way too much noise and not providing good insight to be sending along potentially thousands of errors to Bugsnag indicating that the internet is down. That being said, in our reliability infrastructure, we absolutely should be monitoring for network issues with something better suited for that task such as [UptimeRobot](https://uptimerobot.com/) or [Pingdom](https://www.pingdom.com/).

README.md
```diff
@@ -156,10 +156,12 @@ The special E2E client settings should be defined in `env.yml`:
 | AWS_API_SERVER | string | Optional | aws-api-id.execute-api.us-east-1.amazonaws.com | For generating the swagger document at runtime. Can be null, however that will prevent tools such as swagger-UI from submitting test requests to the API server. |
 | AWS_API_STAGE | string | Optional | stage-name | For generating the swagger document at runtime. Can be null, however that will prevent tools such as swagger-UI from submitting test requests to the API server. |
 | BUGSNAG_API_KEY | string | Required | 123e4567e89b12d3a4564266 | A valid Bugsnag authorization token. |
+| BUGSNAG_APP_TYPE | string | Optional | my-application | An identifier of the app type of this otp-middleware instance. This can be used to segment errors within the Bugsnag dashboard. |
```
binh-dam-ibigroup, line 159: Would you like to put examples as you did in your technical doc about this?

README.md
```diff
@@ -156,10 +156,12 @@ The special E2E client settings should be defined in `env.yml`:
...
 | BUGSNAG_EVENT_JOB_DELAY_IN_MINUTES | integer | Optional | 1 | Bugsnag event job frequency. |
 | BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_MINUTES | integer | Optional | 5 | Bugsnag event request frequency. |
 | BUGSNAG_ORGANIZATION | string | Optional | 123e4567e89b12d3a4564266 | A valid Bugsnag organization id. |
 | BUGSNAG_PROJECT_NOTIFIER_API_KEY | string | Optional | 123e4567e89b12d3a4564266 | A valid Bugsnag project API key. |
+| BUGSNAG_RELEASE_STAGE | string | Optional | staging | An identifier of the release stage of this otp-middleware instance. This can be used to segment errors within the Bugsnag dashboard. |
```
binh-dam-ibigroup, line 164: Would you like to put examples as you did in your technical doc about this?

binh-dam-ibigroup (approved):
Maybe add examples of appType and release stages in the docs. Otherwise LGTM with what @landonreed said.

landonreed:
@evansiroky, all of that sounds fine, but the issues we're encountering are not strictly "network issues," which I'm happy to monitor with Uptime Robot or Pingdom. I'm just asking for a specific proposal (whether it resides within the OTP codebase or here) that will help detect these unexpected timeouts -- which are not a result of bad network connectivity, but due to some yet-to-be-resolved OTP performance issues.

evansiroky:
> @evansiroky, all of that sounds fine, but the issues we're encountering are not strictly "network issues," which I'm happy to monitor with Uptime Robot or Pingdom. I'm just asking for a specific proposal (whether it resides within the OTP codebase or here) that will help detect these unexpected timeouts -- which are not a result of bad network connectivity, but due to some yet-to-be-resolved OTP performance issues.

I, unfortunately, don't have an exact solution in mind, but whatever that solution is, the origin of the problem is with the external service. Therefore it doesn't make sense to me to have errors originating from otp-middleware about problems happening elsewhere.

README.md
evansiroky: I think this is technical enough information that if someone wants to enter in this information they can do some additional research to figure out what this means and if they want to set these values to something.

README.md
evansiroky: I think this is technical enough information that if someone wants to enter in this information they can do some additional research to figure out what this means and if they want to set these values to something.

landonreed:
Just to follow up on this, I think there needs to be a specific proposal to fill the gap here before we can merge. Also, looks like there are now some conflicts from @br648's PR.

evansiroky:
I didn't get around to fully resolving this matter, so it might be worth looking into again if better Bugsnag reports of exceptions caught in otp-middleware are needed in the future.

PR #148, opened 2021-08-09
Bugsnag refactor

br648:
### Checklist

### Description

Refactor to address issues raised by the Bugsnag time. Organization event data requests every five minutes is having a negative impact on the performance of their service. To address the following updates have been made:

1) Organization wide event data requests have been drop in favour of project specific event data requests (based on monitored components).
2) A on start-up process to seed the database with the up to the previous two weeks of project event data has been implemented.
3) A daily process to update the last 24 hours of project event data has been implemented.
4) A Bugsnag webhook end point has been implemented to process project errors once processed by Bugsnag.

### Config changes

Added: BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_HOURS
Added: BUGSNAG_WEBHOOK_PERMITTED_IPS: 104.196.245.109, 104.196.254.247
Deleted: BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_MINUTES
Deleted: BUGSNAG_ORGANIZATION

### Notes

1) I have only been able to test the Bugsnag webhook locally with test data as the IBI firewall prevents direct access by Bugsnag.
2) We perhaps need to consider security implications of this webhook.


landonreed (changes requested):
This is looking good! I've provided some changes in https://github.com/ibi-group/otp-middleware/pull/149 that eliminate the need for BugsnagConfig, but let me know what you think of the approach there.

br648:
@landonreed I've merged your updates and made a few tweaks after reviewing. Although this approach is more complex, it does cover cases where events could be missed using the other approach.

README.md
```diff
@@ -90,15 +90,14 @@ OpenTripPlanner components that it is monitoring.
 
 #### Bugsnag Configuration Parameters
 
-These values can used as defined here (were applicable), or commented out so the default values are used. Parameters 
-that don't have default values (N/A) can be obtained my following the steps in the next section.
+These values can be used as defined here (were applicable), or commented out, so the default values are used. Parameters
```
landonreed:
```suggestion
These values can be used as defined here (where applicable), or commented out, so the default values are used. Parameters 
```

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
```diff
@@ -29,46 +37,106 @@
...
             0,
             BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_HOURS,
             TimeUnit.HOURS);
-
+        // Schedule projects already requested to offset by the job delay period.
+        Scheduler.scheduleJob(
+            new BugsnagEventRequestJob(projectsRequested),
+            BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_HOURS,
+            BUGSNAG_EVENT_REQUEST_JOB_DELAY_IN_HOURS,
+            TimeUnit.HOURS);
```
landonreed: Nit: On second pass, I suppose we could move this repeated call to scheduleJob into a private method that takes projectIds and an initial delay value.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -94,8 +102,9 @@ private void refreshEventRequest(BugsnagEventRequest request) {
                 Persistence.bugsnagEventRequests.removeById(request.id);
             }
             default: {
-                // Request not completed by Bugsnag yet. Do nothing and await the next cycle/refresh.
-                // TODO: Update the request data in the database? What if the status has changed since the last fetch?
+                // Request not completed by Bugsnag yet. Update the event request to record new status (this may not have
+                // changed) and await the next cycle/refresh.
+                Persistence.bugsnagEventRequests.replace(request.id, refreshedRequest);
```
landonreed: Nice catch! Although, are we voiding the projectId and other custom fields we add? There should probably be an updateEventRequest method that handles this.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -94,8 +102,9 @@ private void refreshEventRequest(BugsnagEventRequest request) {
                 Persistence.bugsnagEventRequests.removeById(request.id);
```
landonreed: I just noticed all these cases are missing break statements.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -78,8 +81,13 @@ private void refreshEventRequest(BugsnagEventRequest request) {
...
+                // First, remove "stale" requests (i.e., those that are older than this latest one) for this project.
+                Persistence.bugsnagEventRequests.removeFiltered(
+                    Filters.and(
+                        lte("dateCreated", request.dateCreated),
+                        eq("projectId", request.projectId)
+                    )
+                );
                 // Next, get and store the new events from the completed request and notify users.
                 List<BugsnagEvent> newEvents = getNewEvents(refreshedRequest);
                 if (newEvents.size() > 0) {
```
landonreed, line 87: What if these stale requests cover a different time period? Seems like there is a risk for that with the new approach.

Also, are we replacing the event request in Mongo once it's completed?

landonreed (changes requested):
Thanks for covering my tracks. There looks to be a couple of little things that we missed.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
br648: Nice catch! Updated to include the `projectId` and `daysInPast`.

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
br648: I've updated to remove the request from Mongo. All older requests will have to follow the same life cycle to be processed/remove etc.

br648:
Updated to hopefully cover missed items. Any thoughts on testing all those potential outcomes!

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
landonreed: Sorry I don't think my prior comment was super clear. I think removing the completed request will cause issues with our logic in `BugsnagJobs#getLatestComparisonRequest`. I have some recommended changes in #150.

landonreed (approved):
OK, approved if you think the changes I recommended in #150 make sense/will work.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagDispatcher.java
```diff
@@ -52,12 +52,9 @@
         "Content-Type", "application/json"
     );
 
-    /**
-     * Filter object defining the boundaries on which event requests will be based.
-     */
-    private static final String EVENT_REQUEST_FILTER = buildEventRequestFilter();
+    private static final Map<Integer, String> REQUEST_FILTERS = new HashMap<>();
```
evansiroky: Please add JavaDoc about what this variable is used for and what the keys represent.

evansiroky: After further examining the code, I don't think saving these request filters like this is needed since the computation and memory savings are minimal.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagDispatcher.java
```diff
@@ -103,44 +102,57 @@ private static String buildEventRequestFilter() {
...
         boolean create = eventDataRequestId == null;
         URI eventDataRequestUri = HttpUtils.buildUri(
             BUGSNAG_API_URL,
-            "organizations", BUGSNAG_ORGANIZATION, "event_data_requests", eventDataRequestId
+            "projects", projectId, "event_data_requests", eventDataRequestId
         );
         LOG.debug("Making Bugsnag request: {}", eventDataRequestUri);
-
+        String filter = create
+            ? REQUEST_FILTERS.computeIfAbsent(daysInPast, BugsnagDispatcher::buildEventRequestFilter)
```
evansiroky: I don't think saving these request filters like this is needed since the computation and memory savings are minimal.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
```diff
@@ -17,36 +28,118 @@
...
+
+    /**
+     * Retrieve the latest event request for a project.
+     */
+    private static BugsnagEventRequest getLatestComparisonRequest(String projectId) {
+        BugsnagEventRequest lastCompleted = getLatestCompletedRequestForProject(projectId);
+        BugsnagEventRequest lastIncomplete = getLatestIncompleteRequestForProject(projectId);
+        if (lastIncomplete == null) return lastCompleted;
+        boolean lastIncompleteIsExpired = lastIncomplete.status.equalsIgnoreCase("expired");
+        if (lastCompleted == null) return lastIncompleteIsExpired ? null : lastIncomplete;
```
evansiroky: Can you add some inline comments describing what each of these items is checking for?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -71,20 +78,36 @@ private void refreshEventRequest(BugsnagEventRequest request) {
...
+                    sendEmailForEvents(newEvents.size());
+                }
+                break;
+            case "expired":
+                // First, remove the expired request.
+                LOG.info("Event data request for project {} has expired. Removing from database.", request.projectId);
+                request.delete();
+                // Next, immediately trigger a new event data request to replace the expired one.
+                LOG.info("Triggering a new event data request for project {} to replace previously expired.", request.projectId);
+                BugsnagEventRequestJob.triggerEventDataRequestForProject(request.projectId, request.daysInPast);
```
evansiroky: Shouldn't this recalculate the `daysInPast` variable in case an expired request has occurred multiple days in the past?

src/main/java/org/opentripplanner/middleware/bugsnag/jobs/BugsnagEventHandlingJob.java
```diff
@@ -142,4 +164,29 @@ private void removeStaleEvents() {
...
+            if (webHookDelivery != null) {
+                LOG.info("New event delivered via the Bugsnag webhook. Storing and notifying subscribed admin users.");
+                Persistence.bugsnagEvents.create(new BugsnagEvent(webHookDelivery));
+                // Notify any subscribed users about new events.
+                sendEmailForEvents(1);
+            }
+        } catch (JsonProcessingException e) {
+            BugsnagReporter.reportErrorToBugsnag("Failed to parse webhook delivery!", request.body(), e);
+        }
+    }
```
evansiroky: This class is specifically for a Job, so I think putting this webhook handler in here is adding too many responsibilities to this class. See proposed refactor in https://github.com/ibi-group/otp-middleware/pull/151.

evansiroky (changes requested):
Looking pretty good, just a few comments and one proposed refactor PR.

src/main/java/org/opentripplanner/middleware/bugsnag/BugsnagJobs.java
```diff
@@ -17,36 +28,137 @@
...
+        return (compareRequest != null)
+            ? Math.min(hoursSinceLastRequest, BUGSNAG_REPORTING_WINDOW_IN_DAYS * 24)
+            : BUGSNAG_REPORTING_WINDOW_IN_DAYS * 24;
+    }
+
+    /**
+     * Return the number of hours (rounded to a whole day) that have passed since the date provided.
+     */
+    public static int getHoursRoundedToWholeDaySinceDate(long historicDateInMillis) {
+        long reportingGapMillis = new Date().getTime() - historicDateInMillis;
```
evansiroky: Please use DateTimeUtils in case we want to mock the time in a test in the future.

```suggestion
        long reportingGapMillis = DateTimeUtils.currentTimeMillis() - historicDateInMillis;
```

src/main/java/org/opentripplanner/middleware/models/BugsnagEventRequest.java
```diff
@@ -29,16 +36,62 @@
...
+    @Override
+    public boolean delete() {
+        return Persistence.bugsnagEventRequests.removeById(id);
+    }
+
+    @JsonIgnore
+    @BsonIgnore
+    public long getTimeWindowEndInMillis() {
+        long timeWindowMillis = TimeUnit.MILLISECONDS.convert(this.daysInPast, TimeUnit.DAYS);
+        return this.dateCreated.getTime() + timeWindowMillis;
```
evansiroky: Shouldn't this just be as follows?

```suggestion
        return this.dateCreated.getTime();
```

It would seem that the end of the applicable time window of a request for events would be whenever the event request was created.

src/test/java/org/opentripplanner/middleware/bugsnag/BugsnagTest.java
```diff
@@ -1,67 +1,299 @@
...
+        bugsnagEventRequest.delete();
+    }
+
+    /**
+     * Provide a 'pending' event request which when refreshed (using CompletedEventDataRequest.son) is updated to
+     * 'completed'. Using the url parameter from the refreshed event request, process the event data response
+     * (eventDataResponse.json) and save the Bugsnag event. Confirm that the refreshed event request status has been
+     * updated to 'completed' and that the Bugsnag event has been saved.
+     *
+     * Note: the CompletedEventDataRequest.son contains a hardcoded reference to the wiremock URL:
```
evansiroky: Typo?

```suggestion
     * Note: the CompletedEventDataRequest.json contains a hardcoded reference to the wiremock URL:
```

src/test/java/org/opentripplanner/middleware/bugsnag/BugsnagTest.java
```diff
@@ -1,67 +1,299 @@
...
+                )
+        );
+        BugsnagEventHandlingJob bugsnagEventHandlingJob = new BugsnagEventHandlingJob();
+        bugsnagEventHandlingJob.refreshEventRequest(bugsnagEventRequest);
+        BugsnagEventRequest updatedBugsnagEventRequest = Persistence.bugsnagEventRequests.getById(bugsnagEventRequest.id);
+        assertEquals(updatedBugsnagEventRequest.status, "COMPLETED");
+        bugsnagEventRequest.delete();
+        BugsnagEvent bugsnagEvent = Persistence.bugsnagEvents.getOneFiltered(Filters.eq("projectId", projectId));
+        assertNotNull(bugsnagEvent);
+        Persistence.bugsnagEvents.removeById(bugsnagEvent.id);
```
evansiroky: This seems like some kind of post-test cleanup. It might make sense to put this in a `@BeforeEach`-decorated method and always remove all bugsnagEvents/bugsnagEventRequests. This way if a single test fails before hitting the cleanup section, it can be ensured that the database will be good to go. This way tests can be certain to not have effects on other tests.

src/test/java/org/opentripplanner/middleware/bugsnag/BugsnagTest.java
```diff
@@ -1,67 +1,299 @@
...
+                Filters.and(
+                    Filters.eq("projectId", projectId),
+                    Filters.eq("eventDataRequestId", replacedEventDataRequestId)
+                )
+            );
+        assertNotNull(newBugsnagEventRequest);
+        // Rounding hour to nearest whole day bumps this by a day from the original 2 to 3 days.
+        assertEquals(3, newBugsnagEventRequest.daysInPast);
+        assertEquals("PREPARING", newBugsnagEventRequest.status);
+        newBugsnagEventRequest.delete();
```
evansiroky: This seems like some kind of post-test cleanup. It might make sense to put this in a `@BeforeEach`-decorated method and always remove all bugsnagEvents/bugsnagEventRequests. This way if a single test fails before hitting the cleanup section, it can be ensured that the database will be good to go. This way tests can be certain to not have effects on other tests.

Same comment at the end in each of the next 6 tests.

evansiroky (changes requested):
Just a few more comments.

evansiroky (approved):

evansiroky:
Assigning back to @br648 to coordinate merging and releasing.

PR #152, opened 2021-08-27
Connected Data Platform - Phase 1

br648:
### Checklist

### Description

This PR is the first phase of the CDP. It covers the following tasks:

1. Allows a user to remove trip history.
2. Anonimizes trip data.
3. Uploads zipped trip data to S3.
4. Manages which trip history days should be uploaded to S3.

### New Configuration Parameters

CONNECTED_DATA_PLATFORM_S3_BUCKET_NAME: otp-middleware-dev
CONNECTED_DATA_PLATFORM_S3_FOLDER_NAME: trip-data
CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_FREQUENCY_IN_MINUTES: 5

### Notes

This update is dependant on the UI providing a batch id. See https://github.com/opentripplanner/otp-react-redux/pull/478


br648:
@landonreed This is now ready for first review. The tests are failing because GitHub will need the new S3 config params and I'm not sure how to update them. The required trip summary params will need updated. For now, I have focused on getting the whole process in place first.

src/main/java/org/opentripplanner/middleware/cdp/AnonymizedTripRequest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.opentripplanner.middleware.models.OtpUser;
+
+/**
+ * Class to hold anonymized trip requests only.
+ */
+public class AnonymizedTripRequest {
+    /**
+     * User Id. {@link OtpUser#id} of user making trip request.
+     */
+    public String userId;
```
miles-grant-ibigroup: Is this removed before being written to a zip file? I'm guessing it's here to allow users to be forgotten, but it does feel strange to have a `userId` field on a class titled `AnonymizedTripRequest`

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,219 @@
...
+                tempZipFile
+            );
+            return true;
+        } catch (S3Exception | IOException e) {
+            LOG.error("Failed to process trip data for {}", startOfDay, e);
+            return false;
+        } finally {
+            // Delete the temporary zip file if it was created. This is done here in case the S3 upload fails.
+            try {
+                if (tempZipFile != null && !IS_TEST) {
```
miles-grant-ibigroup: It might be a good idea to leave a warning in the console that the file isn't actually being deleted when test mode is active?

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,219 @@
...
+            } catch (IOException e) {
+                LOG.error("Failed to delete temp zip file {}", tempZipFile.getAbsolutePath(), e);
+            }
+        }
+    }
+
+    /**
+     * Get all dates that have not been uploaded and place in a Set class.
+     */
+    private static Set<Date> getIncompleteUploadsAsSet() {
```
miles-grant-ibigroup: Since this is only used once, would this not be cleaner as an override of `getIncompleteUploads()`?

src/main/java/org/opentripplanner/middleware/models/TripHistoryUpload.java
```diff
@@ -0,0 +1,66 @@
...
+import org.bson.conversions.Bson;
+import org.opentripplanner.middleware.cdp.TripHistoryUploadStatus;
+import org.opentripplanner.middleware.persistence.Persistence;
+
+import java.util.Date;
+
+public class TripHistoryUpload extends Model {
+
+    public Date uploadDate;
+    public String status = TripHistoryUploadStatus.PENDING.getValue();;
```
miles-grant-ibigroup: Double semi-colon!

src/main/java/org/opentripplanner/middleware/cdp/TripHistoryUploadStatus.java
```diff
@@ -0,0 +1,26 @@
...
+     */
+    PENDING("PENDING");
+
+    private final String value;
+
+    TripHistoryUploadStatus(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
```
miles-grant-ibigroup: Is this needed?

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,239 @@
...
+        // Update 'storeTripHistory' value.
+        otpUser.storeTripHistory = false;
+        mockAuthenticatedRequest(
+            String.format("%s/%s", OTP_USER_PATH, otpUser.id),
+            JsonUtils.toJson(otpUser),
+            otpUser,
+            HttpMethod.PUT
+        );
+        assertEquals(
+            1,
```
miles-grant-ibigroup: We expect this because we've only created one TripHistory, right?

configurations/default/env.yml.tmp
```diff
@@ -9,6 +9,10 @@ BUGSNAG_PROJECT_NOTIFIER_API_KEY: project-notifier-api-key
 #BUGSNAG_REPORTING_WINDOW_IN_DAYS: 14
 BUGSNAG_WEBHOOK_PERMITTED_IPS: 104.196.245.109, 104.196.254.247
 
+CONNECTED_DATA_PLATFORM_S3_BUCKET_NAME: bucket-name
+CONNECTED_DATA_PLATFORM_S3_FOLDER_NAME: folder-name
+CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_DELAY_IN_MINUTES: 5
```
miles-grant-ibigroup: Is "delay" the best term here? Isn't it more of a frequency?

miles-grant-ibigroup (commented):
Looks really good! Just a few small questions here and there about details

src/main/java/org/opentripplanner/middleware/cdp/AnonymizedTripRequest.java
br648: This was included in the design doc by Landon. I think the plan was to hold the anonymized data separately and the user id was the key. I've implemented slightly differently (didn't see the need for a separate table) so have now removed from this class.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
br648: Since it is only used once I have removed.

src/main/java/org/opentripplanner/middleware/cdp/TripHistoryUploadStatus.java
br648: yeah, MongoDB can't save enums. This allows me to save the String value of it.

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
br648: yeah, I'll add a comment.

configurations/default/env.yml.tmp
br648: I've used delay in the past, but agree frequency is better, have updated!

miles-grant-ibigroup (approved):
Thanks for those changes! Looks great! Will pass this PR to Evan now.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -118,6 +119,15 @@ U preCreateHook(U user, Request req) {
...
         Auth0Users.validateExistingUser(user, preExistingUser, req, this.persistence);
+        if (user instanceof OtpUser) {
+            OtpUser otpUser = (OtpUser) user;
+            OtpUser existingOtpUser = (OtpUser) preExistingUser;
+            if(!otpUser.storeTripHistory && existingOtpUser.storeTripHistory) {
+                // If an OTP user no longer wants their trip history stored, remove all history from MongoDB and all
+                // files previously uploaded to S3.
+                ConnectedDataManager.removeUsersTripHistory(otpUser.id);
+            }
+        }
```
evansiroky, line 134: I was wondering if a user's data should also be removed if they delete their account, but I suppose this approach is best so that users would specifically note that they want to remove their trip history and then they can delete their account. I'm not even sure if it's possible to delete an account tho, lol.

src/main/java/org/opentripplanner/middleware/cdp/AnonymizedTripRequest.java
```diff
@@ -0,0 +1,33 @@
+package org.opentripplanner.middleware.cdp;
```
evansiroky: Please spell out this package name completely.

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -82,4 +92,33 @@ public boolean delete() {
...
     }
+
+    private AnonymizedTripRequest getAnonimized() {
+        return new AnonymizedTripRequest(batchId, fromPlace, toPlace);
+    }
+
+    /**
+     * Get all trip requests between two dates.
+     */
+    private static FindIterable<TripRequest> getTripRequests(Date start, Date end) {
```
evansiroky: It may just be a personal opinion, but since some of these static methods are only ever called once and from the ConnectedDataManager class that perhaps these methods belong in that class instead.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Remove a user's trip requests and trip summaries from the database. Record the user's trip dates so that all
+     * data previously uploaded to s3 can be recompiled and uploaded again.
+     */
+    public static void removeUsersTripHistory(String userId) {
+        Set<Date> userTripDates = new HashSet<>();
+        for (TripRequest request : TripRequest.requestsForUser(userId)) {
+            userTripDates.add(getStartOfDay(request.dateCreated));
+            // This will delete the trip summaries as well.
+            request.delete();
```
evansiroky: Is there a way to do a bulk-delete instead of a bunch of individual deletes?

src/main/java/org/opentripplanner/middleware/models/TripHistoryUpload.java
```diff
@@ -0,0 +1,66 @@
...
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Sorts;
+import org.bson.codecs.pojo.annotations.BsonIgnore;
+import org.bson.conversions.Bson;
+import org.opentripplanner.middleware.cdp.TripHistoryUploadStatus;
+import org.opentripplanner.middleware.persistence.Persistence;
+
+import java.util.Date;
+
+public class TripHistoryUpload extends Model {
```
evansiroky, line 17: Can you add some JavaDoc about what this model is used for?

src/main/java/org/opentripplanner/middleware/models/TripHistoryUpload.java
```diff
@@ -0,0 +1,66 @@
...
+    public TripHistoryUpload(Date uploadDate) {
+        this.uploadDate = uploadDate;
+        this.status = TripHistoryUploadStatus.PENDING.getValue();
+    }
+
+    /**
+     * Get all incomplete uploads.
+     */
+    @BsonIgnore
+    public static FindIterable<TripHistoryUpload> getIncompleteUploads() {
```
evansiroky: It'd also be my personal preference to move these static methods into the classes where they're called from.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,216 @@
...
+    public static void processTripHistory() {
+        List<TripHistoryUpload> incompleteUploads = getIncompleteUploads();
+        incompleteUploads.forEach(tripHistoryUpload -> {
+            if (compileAndUploadTripHistory(tripHistoryUpload.uploadDate)) {
+                // Update the status to 'completed' if successfully compiled and uploaded.
+                tripHistoryUpload.status = TripHistoryUploadStatus.COMPLETED.getValue();
+                Persistence.tripHistoryUploads.replace(tripHistoryUpload.id, tripHistoryUpload);
+            }
+        });
+    }
```
evansiroky: It seems like it'd make sense to put this in the `TripHistoryUploadJob`.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,216 @@
...
+                newDate.before(getStartOfDay(new Date())))
+            ) {
+                // If the new date is the same or after the first ever upload date, add it to the upload list. This acts
+                // as a back stop to prevent historic uploads being created indefinitely. Also, make sure the new date
+                // is before today because the trip data for it isn't uploaded until after midnight.
+                Persistence.tripHistoryUploads.create(new TripHistoryUpload(newDate));
+            }
+        });
+        // Kick-off trip data processing to recompile and upload trip data to S3 minus the user's trip history.
+        processTripHistory();
```
evansiroky: This could be a long-running process, so I think it'd be best to start a `TripHistoryUploadJob`.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -118,6 +119,15 @@ U preCreateHook(U user, Request req) {
     @Override
     U preUpdateHook(U user, U preExistingUser, Request req) {
         Auth0Users.validateExistingUser(user, preExistingUser, req, this.persistence);
+        if (user instanceof OtpUser) {
+            OtpUser otpUser = (OtpUser) user;
+            OtpUser existingOtpUser = (OtpUser) preExistingUser;
+            if(!otpUser.storeTripHistory && existingOtpUser.storeTripHistory) {
+                // If an OTP user no longer wants their trip history stored, remove all history from MongoDB and all
+                // files previously uploaded to S3.
+                ConnectedDataManager.removeUsersTripHistory(otpUser.id);
```
evansiroky, line 128: See other related comments about how this could involve starting a long-running process and therefore should trigger the creation of a job.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,216 @@
...
+        } else {
+            Set<LocalDate> betweenDays = DateTimeUtils.getDatesBetween(
+                getDatePlusNumberOfDays(latest.uploadDate,1),
+                now
+            );
+            betweenDays.forEach(day -> {
+                Persistence.tripHistoryUploads.create(new TripHistoryUpload(getStartOfDay(day)));
+            });
+        }
+    }
```
evansiroky: This would probably also make sense to include in the `TripHistoryUploadJob`.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
```diff
@@ -0,0 +1,216 @@
...
+    private static boolean compileAndUploadTripHistory(Date date) {
+        Date startOfDay = getStartOfDay(date);
+        String zipFileName = getFileName(startOfDay, ZIP_FILE_NAME_SUFFIX);
+        File tempZipFile = null;
+        try {
+            String tempFile = String.join("/", FileUtils.getTempDirectory().getAbsolutePath(), zipFileName);
+            FileUtils.writeFileToZip(
+                tempFile,
+                getFileName(startOfDay, DATA_FILE_NAME_SUFFIX),
+                getAnonymizedTripHistory(startOfDay, getEndOfDay(date))
```
evansiroky: I'm a little worried that this might create a massive JSON file that would result in an out-of-memory error. Is there a way you can write this data in a streaming fashion?

src/main/java/org/opentripplanner/middleware/utils/FileUtils.java
```diff
@@ -31,4 +42,64 @@ public static String getFileContents(String pathAndFileName) throws IOException
...
+     * Gets the operating system's temp directory.
+     */
+    public static File getTempDirectory() {
+        return new File(System.getProperty("java.io.tmpdir"));
+    }
+
+    /**
+     * Extracts the contents of a file contented within a zip file.
+     */
+    public static String getContentsOfFileInZip(String zipFileNameAndPath, String fileName) throws IOException {
```
evansiroky, line 101: It appears this is only used during testing which is good since it attempts to read the entire file into memory which could result in an out-of-memory error if used on a massive zip file. I'd add a warning to the JavaDoc of the class noting this just in case.

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,241 @@
...
+            "/",
+            FileUtils.getTempDirectory().getAbsolutePath(),
+            zipFileName
+        );
+        String fileContents = getContentsOfFileInZip(
+            tempFile,
+            getFileName(startOfYesterday, ConnectedDataManager.DATA_FILE_NAME_SUFFIX)
+        );
+        String contents = ConnectedDataManager.getAnonymizedTripHistory(startOfYesterday, getEndOfDay(startOfYesterday));
+        assertEquals(fileContents, contents);
```
evansiroky: This doesn't seem to be a great assertion here since a subcomponent of the `processTripHistory` method is being called to generate the expected data. So long as the implementation remains the same, it's always going to match and thus what you're really testing is anything up until the calling of `getAnonymizedTripHistory`. In this case, I'd recommend creating a snapshot of the data to assert against. See example of snapshotting in gtfs-lib: https://github.com/conveyal/gtfs-lib/blob/94ff5fda5785020e3a689e8f5daf33291b4b7625/src/test/java/com/conveyal/gtfs/graphql/GTFSGraphQLTest.java#L99

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,241 @@
...
+            "/",
+            FileUtils.getTempDirectory().getAbsolutePath(),
+            zipFileName
+        );
+        String fileContents = getContentsOfFileInZip(
+            tempFile,
+            getFileName(startOfYesterday, ConnectedDataManager.DATA_FILE_NAME_SUFFIX)
+        );
+        TripHistory tripHistory = JsonUtils.getPOJOFromJSON(fileContents, TripHistory.class);
+        assertTrue(tripHistory.tripRequests.stream().anyMatch(tripRequest -> tripRequest.batchId.equals(batchId)));
```
evansiroky: Please add a comment about what this assertion checks.

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,241 @@
...
+        // Trip request and summary are removed as part of the 'removeUsersTripHistory' method. Setting these to null
+        // prevents a delete error under the tidy-up process.
+        tripRequest = null;
+        tripSummary = null;
+        fileContents = getContentsOfFileInZip(
+            tempFile,
+            getFileName(startOfYesterday, ConnectedDataManager.DATA_FILE_NAME_SUFFIX)
+        );
+        tripHistory = JsonUtils.getPOJOFromJSON(fileContents, TripHistory.class);
+        assertTrue(tripHistory.tripRequests.stream().noneMatch(tripRequest -> tripRequest.batchId.equals(batchId)));
```
evansiroky: Please add a comment about what this assertion checks.

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,241 @@
...
+            tempFile,
+            getFileName(startOfYesterday, ConnectedDataManager.DATA_FILE_NAME_SUFFIX)
+        );
+        TripHistory tripHistory = JsonUtils.getPOJOFromJSON(fileContents, TripHistory.class);
+        assertTrue(tripHistory.tripRequests.stream().anyMatch(tripRequest -> tripRequest.batchId.equals(batchId)));
+        ConnectedDataManager.removeUsersTripHistory(userId);
+        // Trip request and summary are removed as part of the 'removeUsersTripHistory' method. Setting these to null
+        // prevents a delete error under the tidy-up process.
+        tripRequest = null;
+        tripSummary = null;
```
evansiroky: Test cases shouldn't have side effects on other test cases if at all possible. Is there a way these variables can be made to have a scope active only within this method?

src/test/java/org/opentripplanner/middleware/cdp/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+        // Set the zip file name and temp file name so they are both removed as part of the tidy-up process.
+        zipFileName = getFileName(tenDaysInThePast, ConnectedDataManager.ZIP_FILE_NAME_SUFFIX);
+        tempFile = String.join(
+            "/",
+            FileUtils.getTempDirectory().getAbsolutePath(),
+            zipFileName
+        );
```
evansiroky: Test cases shouldn't have side effects on other test cases if at all possible. Is there a way these variables can be made to have a scope active only within this method?

evansiroky (dismissed):
Looks pretty good. Most of my comments are about perhaps personal refactoring preferences. However, the two items that I believe are blocking are:

- try to make writing to the zip file a fully-streamed process
- make it so an API request to remove a user's trip history starts a job that runs independently of the API request.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
br648: I could go with `removeFiltered(eq("userId", userId))`, but I would still need to loop through each in order to get the trip request id so I could delete the related trip summaries.

src/main/java/org/opentripplanner/middleware/cdp/ConnectedDataManager.java
br648: Done.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Anonymous version of {@link org.opentripplanner.middleware.otp.response.Itinerary} containing only parameters
+ * flagged as anonymous.
+ */
+public class AnonymizedItinerary {
+
+    /**
+     * Duration of the trip on this itinerary, in seconds.
+     */
+    public Long duration = 0L;
```
leonardehrenfried, line 16: I see that there are both member variables of type `Long` and `long`. Is this one intentionally nullable?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Time that the trip arrives.
+     */
+    public Date endTime = null;
+
+    /**
+     * The number of transfers this trip has.
+     */
+    public Integer transfers = 0;
```
leonardehrenfried, line 31: Same here. This is nullable. Should it be this way?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Duration of the trip on this itinerary, in seconds.
+     */
+    public Long duration = 0L;
+
+    /**
+     * Time that the trip departs.
+     */
+    public Date startTime = null;
```
leonardehrenfried, line 21: You're introducing `java.util.Date` which is problematic as it has so many deprecated methods, check: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Date.html

It also has a terrible API and John Skeet says ["You should avoid it if you can."](https://codeblog.jonskeet.uk/2017/04/23/all-about-java-util-date/. Luckily it has been replaced with a much nicer one since Java 8: `java.time`

I seriously suggest you should switch to it.

I have quite a bit experience with `java.time` and would love to help if you're unfamiliar with it.

Edit: I see that you're using `java.time` elsewhere in this PR. Why not use `OffsetDateTime` or `Instant` instead of Date?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Batch Id. Id for trip requests planned together but representing different modes.
+     */
+    public String batchId;
+
+    /**
+     * From place Lat. Trip starting point.
+     */
+    public Double fromPlaceLat;
```
leonardehrenfried: What's the reason for the lat/long part to be put into separate variables? Would it not make sense to use some type of `Coordinates` object?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        this.fromPlaceLat = fromCoordinates.latitude;
+        this.fromPlaceLon = fromCoordinates.longitude;
+        this.toPlaceLat = toCoordinates.latitude;
+        this.toPlaceLon = toCoordinates.longitude;
+        if (tripRequest. requestParameters != null) {
+            this.date = tripRequest.requestParameters.get("date");
+            this.time = tripRequest.requestParameters.get("time");
+            this.tripType = tripRequest.requestParameters.get("arriveBy");
+            if (tripType != null && tripType.equalsIgnoreCase("true")) {
+                tripType = "Arrive By";
```
leonardehrenfried: Isn't this pretty "stringly typed"? Would you not want to use an enum?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,324 @@
...
+import static org.opentripplanner.middleware.utils.DateTimeUtils.getStringFromDate;
+
+/**
+ * Responsible for collating, anonymizing and uploading to AWS s3 trip requests and related trip summaries.
+ */
+public class ConnectedDataManager {
+
+    // If set to true, no files are upload to S3 or deleted from the local disk. This is expected to be carried out by
+    // the unit tests instead.
+    public static boolean IS_TEST;
```
leonardehrenfried: I'm not sure that this isn't a code smell. Could it not be replace with a parameter on the method to stop the deletion from happening?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
```diff
@@ -0,0 +1,85 @@
...
+     * passed midnight and any days missed due to downtime.
+     */
+    public static void stageUploadDays() {
+        Date now = new Date();
+        TripHistoryUpload latest = TripHistoryUpload.getLatest();
+        if (latest == null) {
+            // No data held, add the previous day as the first day to be uploaded.
+            Persistence.tripHistoryUploads.create(
+                new TripHistoryUpload(
+                    getStartOfDay(
```
leonardehrenfried: This is a good example where `java.time`'s API will make things a lot easier.

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
```diff
@@ -201,4 +207,72 @@ public static ZonedDateTime makeOtpZonedDateTime(String dateString, String timeS
     public static ZonedDateTime makeOtpZonedDateTime(Date date) {
         return ZonedDateTime.ofInstant(date.toInstant(), getOtpZoneId());
     }
+
+    /**
+     * Converts a {@link LocalDateTime} object into a {@link Date} object using the Otp zone id.
+     */
+    public static Date convertToDate(LocalDateTime dateToConvert) {
```
leonardehrenfried, line 213: I'm repeating myself but I would use a `OffsetDateTime` or `ZonedDateTime`.

leonardehrenfried: This will make lots of these utiity methods redundant as they are so much easier to use.

src/main/java/org/opentripplanner/middleware/utils/LatLongUtils.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Objects;
+import java.util.Random;
+
+/**
+ * Utility class to randomize lat/long coordinates. Information on this can be found here:
+ * https://gis.stackexchange.com/questions/25877/generating-random-locations-nearby. Original code (in Javascript) can
+ * be found and tested here: http://jsfiddle.net/hoolymama/56wzdtax/.
+ */
+public class LatLongUtils {
+    public static boolean IS_TEST = false;
```
leonardehrenfried: Again, I'm not too hot about this. I would create some form of randomiser interface or method that has two implementations: one for production and one for testing.

leonardehrenfried (changes requested):
Looks good, but I have a few questions.

The only thing that I really care about though is not using `java.util.Date`.

src/main/java/org/opentripplanner/middleware/utils/LatLongUtils.java
br648: Fair point. I have overloaded the method.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: Yes, have updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: It would yes. I have updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: yep! Have converted to an enum.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
br648: This was a copy from the original Itinerary class which also has Long and long. I have updated all to Long to be consistent.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
br648: This was a copy from the original Itinerary class which also has Integer. I have left this as is.

br648:
@leonardehrenfried I have addressed your PR feedback minus anything related to Date/Time. This will be covered in tonights (05/10) tech call. I will address according to feedback from that.

leonardehrenfried (approved):

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
br648: Migrated.

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
br648: Migrated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
br648: I have migrated to java.time where possible.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * How far the user has to walk, in meters.
+     */
+    public Double walkDistance = 0.0;
+
+    /**
+     * How much time is spent walking, in seconds.
+     */
+    public long walkTime = 0;
```
binh-dam-ibigroup: Make this one nullable too?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -0,0 +1,57 @@
+package org.opentripplanner.middleware.connecteddataplatform;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * Anonymous version of {@link org.opentripplanner.middleware.otp.response.Itinerary} containing only parameters
+ * flagged as anonymous.
```
binh-dam-ibigroup: Could you clarify "flagged"? Actually, could you clarify "anonymized"? Maybe a better wording could be "parameters that don't contain precise user or location data".  (Also update JavaDoc description of other anonymized classes that have a similar description.)

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedPlace.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Anonymous version of {@link org.opentripplanner.middleware.otp.response.Place} containing only parameters
+ * flagged as anonymous.
+ */
+public class AnonymizedPlace {
+
+    public Date arrival;
+    public Date departure;
+    public Coordinates coordinates;
```
binh-dam-ibigroup, line 17: Add a comment saying the coordinates will be scrambled upon writing to a file.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -0,0 +1,73 @@
...
+    public AnonymizedTripRequest() {
+    }
+
+    public AnonymizedTripRequest(
+        TripRequest tripRequest,
+        Coordinates fromCoordinates,
+        Coordinates toCoordinates
+    ) {
+        this.batchId = tripRequest.batchId;
+        this.fromPlace = fromCoordinates;
```
binh-dam-ibigroup, line 57: Remove duplicate statement.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
```diff
@@ -0,0 +1,74 @@
...
+        processTripHistory(false);
+    }
+
+    /**
+     * Add to the upload list any dates between now and the last upload date. This will cover a new day once
+     * passed midnight and any days missed due to downtime.
+     */
+    public static void stageUploadDays() {
+        LocalDate now = LocalDate.now();
+        TripHistoryUpload latest = TripHistoryUpload.getLatest();
```
binh-dam-ibigroup: This should probably be `getLatestCompleted` (in JavaDoc too). The current check does not look for complete status of upload.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final int HISTORIC_UPLOAD_DAYS_BACK_STOP = 20;
+
+    public void run() {
+        stageUploadDays();
+        processTripHistory(false);
+    }
+
+    /**
+     * Add to the upload list any dates between now and the last upload date. This will cover a new day once
```
binh-dam-ibigroup: Description is incorrect. Update description to reflect that the code triggers upload of trip requests from the last upload date up to the day before now.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+     * Stream trip requests to file. This approach is used to avoid having a large amount of data in memory which could
+     * cause an out-of-memory error if there are a lot of trip requests to process.
+     *
+     * Process:
+     *
+     * 1) Extract unique batch ids between two dates.
+     * 2) For each batch id get matching trip requests.
+     * 3) Workout which trip request uses the most modes.
+     * 4) Define lat/lon for 'from' and 'to' places.
+     * 5) Anonymize trip request with the most modes.
```
binh-dam-ibigroup, line 119: Add to this line: ", scrambling location coordinates for non-public locations"

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+     * 7) Write anonymized trip request and related summaries to file.
+     */
+    private static void streamAnonymousTripsToFile(
+        String pathAndFileName,
+        LocalDateTime start,
+        LocalDateTime end,
+        boolean isTest
+    ) throws IOException {
+        // Get distinct batchId values between two dates.
+        DistinctIterable<String> uniqueBatchIds = Persistence.tripRequests.getDistinctFieldValues(
```
binh-dam-ibigroup, line 130: The UI currently sends out all requests with batch IDs set to -1, so another string needs to be used for uniqueness.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+                    Filters.eq("batchId", uniqueBatchId)
+                ),
+                Sorts.descending("dateCreated", "batchId")
+            );
+
+            TripRequest tripRequest = getTripRequestWithMaxModes(tripRequests);
+
+            // Get all trip summaries matching the batch id.
+            FindIterable<TripSummary> tripSummaries = Persistence.tripSummaries.getFiltered(
+                eq("batchId", uniqueBatchId),
```
binh-dam-ibigroup, line 166: Trip summaries should be matched to one of the fetched trip requests above by `tripRequestId`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+        ) {
+            return (isFirstLeg)
+                ? itineraries.get(0).legs.get(0).transitLeg
+                : itineraries.get(0).legs.get(itineraries.get(0).legs.size() - 1).transitLeg;
+        }
+        return false;
+    }
+
+    /**
+     * Obtain anonymize trip data for the given date, write to zip file, upload the zip file to S3 and finally delete
```
binh-dam-ibigroup: Typo "anonymized"

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -118,6 +120,18 @@ U preCreateHook(U user, Request req) {
     @Override
     U preUpdateHook(U user, U preExistingUser, Request req) {
         Auth0Users.validateExistingUser(user, preExistingUser, req, this.persistence);
+        if (user instanceof OtpUser) {
+            OtpUser otpUser = (OtpUser) user;
+            OtpUser existingOtpUser = (OtpUser) preExistingUser;
+            if(!otpUser.storeTripHistory && existingOtpUser.storeTripHistory) {
+                // If an OTP user no longer wants their trip history stored, remove all history from MongoDB.
```
binh-dam-ibigroup, line 127: Do we need to show a confirmation message about that in the UI?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+        boolean placeIsPublic = true;
+        for (TripSummary tripSummary : tripSummaries) {
+            placeIsPublic = isLegTransit(tripSummary.itineraries, isFirstLeg);
+            if (!placeIsPublic) {
+                // if a single trip summary is not public the place lat/lon must be randomized.
+                break;
+            }
+        }
+
+        String coords = place.split("::")[1].trim();
```
binh-dam-ibigroup: Careful, the UI might send just the coordinates (if the geocoder does not return anything, which is unlikely).
If that happens, the format will just be `lat,lon` and `::` will not be present.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,334 @@
...
+        Coordinates coordinates = new Coordinates(
+            Double.parseDouble(coords.split(",")[0]),
+            Double.parseDouble(coords.split(",")[1])
+        );
+        return placeIsPublic ? coordinates : LatLongUtils.getRandomizedCoordinates(coordinates, isTest);
+    }
+
+    /**
+     * Extract the trip request that uses the most modes.
```
binh-dam-ibigroup: Expand on the reasoning behind this.

src/main/java/org/opentripplanner/middleware/models/TripRequest.java
```diff
@@ -36,24 +38,27 @@
      */
     public String toPlace;
 
-    /**
-     * Query params. Query parameters influencing trip.
-     */
-    //TODO: This could be the request parameters returned as part of the plan response. Would be POJO based instead of just text.
-    public String queryParams;
+    /** A dictionary of the parameters provided in the request that triggered this response. */
+    public HashMap<String, String> requestParameters;
```
binh-dam-ibigroup, line 42: Note for self: Ooooh this change needs to be reflected in https://github.com/opentripplanner/otp-react-redux/pull/329.

src/main/java/org/opentripplanner/middleware/utils/FileUtils.java
```diff
@@ -31,4 +41,82 @@ public static String getFileContents(String pathAndFileName) throws IOException
...
+
+    /**
+     * Gets the operating system's temp directory.
+     */
+    public static File getTempDirectory() {
+        return new File(System.getProperty("java.io.tmpdir"));
+    }
+
+    /**
+     * Extracts the contents of a file contented within a zip file. This method will read the entire contents of a file
```
binh-dam-ibigroup: Replace "contented" with "contained".

src/test/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataPlatformTest.java
```diff
@@ -0,0 +1,426 @@
...
+            tempFile,
+            getFileName(startOfYesterday, ConnectedDataManager.DATA_FILE_NAME_SUFFIX)
+        );
+        MatcherAssert.assertThat(fileContents, matchesSnapshot());
+
+        // Confirm that all non transit lat/lon's have been randomized (with test lat/lon).
+        List<AnonymizedTrip> anonymizedTrips = JsonUtils.getPOJOFromJSONAsList(fileContents, AnonymizedTrip.class);
+        assertNotNull(anonymizedTrips);
+        assertEquals(LatLongUtils.TEST_LAT, anonymizedTrips.get(0).tripRequest.fromPlace.lat);
+        assertEquals(LatLongUtils.TEST_LON, anonymizedTrips.get(0).tripRequest.fromPlace.lon);
```
binh-dam-ibigroup: You assert using `Coordinates.equals`.

binh-dam-ibigroup (changes requested):
It looks like some changes in the logic for compiling trips and for determining the last complete S3 update are necessary. See also other changes.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
br648: I'm looking for the last trip history upload regardless of status as a starting point for staging upload days see `TripHistoryUploadJob.stageUploadDays()`. I've updated the method to `getLastCreated` and updated the Javadoc accordingly.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: This was discussed at the initial design meeting with Landon. The UI would be updated to provide a unique id and not -1. I don't think there is another param that can be used?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: See previous comment. I think this needs to remain batch id.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
br648: Yes, probably prior to the user hitting send/update. This can be done with the batch id update.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
binh-dam-ibigroup: You are right on this one. A change in https://github.com/opentripplanner/otp-react-redux/pull/329 will cause only one trip request to be recorded, but trip summaries for trip requests under the same batch will still be recorded and have no parent `tripRequestId`... In that case could we remove the `tripRequestId` field from the `TripSummary` class?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -232,21 +233,22 @@ private static Coordinates getPlaceCoordinates(
     }
 
     /**
-     * Extract the trip request that uses the most modes.
+     * A single trip query results in many calls from the UI to OTP covering different combinations of modes. The trip
+     * request to be included in the anonymous trip data must be the one which uses the most modes.
```
binh-dam-ibigroup: This method could probably deleted in favor of retrieving the first request found for a batch. We don't use the number of modes involved for anything, do we?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
```diff
@@ -22,20 +22,21 @@ public void run() {
...
+     * completed) trip history upload. This will cover a new day once passed midnight and any days missed due to
+     * downtime.
      */
     public static void stageUploadDays() {
         LocalDate now = LocalDate.now();
-        TripHistoryUpload latest = TripHistoryUpload.getLatest();
-        if (latest == null) {
+        TripHistoryUpload lastCreated = TripHistoryUpload.getLastCreated();
+        if (lastCreated == null) {
             // No data held, add the previous day as the first day to be uploaded.
```
binh-dam-ibigroup, line 33: Should the same logic from the else block be applied on first run? (upload the last say 20 days?)

binh-dam-ibigroup (changes requested):
Looks mostly good, but remove method `getTripRequestWithMostModes` because we don't consider the number of modes anywhere else for this PR.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: @binh-dam-ibigroup this method only exists so that the trip request with the most modes is serialized and uploaded to S3. This way the client can see the maximum number of modes that were used to produce the trip summaries. It is possible that I haven't understood this correctly, if so, let me know.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
br648: No, this is to cover the case where the job has not run before e.g. initial deployment.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: Addressed after tech call and code update.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -234,21 +239,27 @@ private static Coordinates getPlaceCoordinates(
...
-            ) {
-                requestWithMostModes = tripRequest;
+            if (request == null) {
+                // Select the first request. All subsequent requests will be the same apart from the mode (which is
+                // overwritten at the end of this method).
+                request = tripRequest;
             }
+            Set<String> uniqueModesForRequest = new HashSet<>(
+                // Expecting the comma separating modes to be encoded as '%2C'.
+                Arrays.asList(tripRequest.requestParameters.get("mode").split("%2C"))
```
binh-dam-ibigroup: You could just add this to `allUniqueModes`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -0,0 +1,350 @@
...
+        Set<String> allUniqueModes = new HashSet<>();
+        for (TripRequest tripRequest : tripRequests) {
+            if (request == null) {
+                // Select the first request. All subsequent requests will be the same apart from the mode (which is
+                // overwritten at the end of this method).
+                request = tripRequest;
+            }
+            Set<String> uniqueModesForRequest = new HashSet<>(
+                // Expecting the comma separating modes to be encoded as '%2C'.
+                Arrays.asList(tripRequest.requestParameters.get("mode").split("%2C"))
```
binh-dam-ibigroup: This assumption is incorrect. It looks like `TripRequest` objects are created in `OtpRequestProcessor` using the OTP response (and not the URL query params), where the mode value is in the form below, not encoded: 
```js
requestParameters: {
  ...
  mode: "CAR_PARK,WALK,TRANSIT",
  ...
}
```

binh-dam-ibigroup: Also, the values appear to be stored that way in Mongo too, so there is no `%2C` in the modes extracted from Mongo.

binh-dam-ibigroup (changes requested):
Please address https://github.com/ibi-group/otp-middleware/pull/152/files#r738677278.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: Good spot! Updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
br648: Thanks. Updated.

src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
br648: Addressed here: https://github.com/opentripplanner/otp-react-redux/pull/478

binh-dam-ibigroup (approved):
Looks good and seems to work now! Thanks for the changes.

PR #155, opened 2021-10-18
Introduce endpoint for running OTP1 and 2 in parallel

leonardehrenfried:
### Checklist

### Description

This PR adds the ability to run OTP1 and OTP2 in parallel by parametrizing all calls in `OtpRequestProcessor` and related classes.

## Questions

- Do all instances have two versions?
- Should this be backwards-compatible?

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
@@ -45,10 +45,6 @@ public class OtpRequestProcessor implements Endpoint {
     private static final Logger LOG = LoggerFactory.getLogger(OtpRequestProcessor.class);
 
     /**
-     * Endpoint for the OTP Middleware's OTP proxy
-     */
-    public static final String OTP_PROXY_ENDPOINT = "/otp";
-    /**
      * URL to OTP's documentation.
      */
     private static final String OTP_DOC_URL = "http://otp-docs.ibi-transit.com/api/index.html";
@@ -77,13 +73,13 @@ public class OtpRequestProcessor implements Endpoint {
             .withDescription("If a third-party application is making a trip plan request on behalf of an end user (OtpUser), the user id must be specified.")
             .build();
         restApi.endpoint(
-            EndpointDescriptor.endpointPath(OTP_PROXY_ENDPOINT).withDescription("Proxy interface for OTP endpoints. " + OTP_DOC_LINK),
+            EndpointDescriptor.endpointPath(basePath).withDescription("Proxy interface for OTP endpoints. " + OTP_DOC_LINK),
             HttpUtils.NO_FILTER
-        ).get(path(basePath + "*")
+        ).get(path("/*")
                 .withDescription("Forwards any GET request to OTP. " + OTP_DOC_LINK)
                 .withQueryParam(USER_ID)
                 .withProduces(List.of(MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML)),
-                (request, response) -> OtpRequestProcessor.proxy(otpVersion, request, response)
+                (request, response) -> OtpRequestProcessor.proxy(otpVersion, basePath, request, response)
         );
     }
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -83,7 +89,7 @@ public void bind(final SparkSwagger restApi) {
      * trip history) the response is intercepted and processed. In all cases, the response from OTP (content and HTTP
      * status) is passed back to the requester.
      */
-    private static String proxy(Request request, spark.Response response) {
+    private static String proxy(OtpVersion version, String basePath, Request request, spark.Response response) {
```
binh-dam-ibigroup: Could this method become a class member (not static) so you can remove the `version` and `basePath` args?

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -34,6 +34,9 @@ tags:
 - name: "otp"
   description: "Proxy interface for OTP endpoints. Refer to <a href='http://otp-docs.ibi-transit.com/api/index.html'>OTP's\
     \ API documentation</a> for OTP's supported API resources."
+- name: "otp2"
+  description: "Proxy interface for OTP endpoints. Refer to <a href='http://otp-docs.ibi-transit.com/api/index.html'>OTP's\
```
binh-dam-ibigroup: I think OTP v2 should be mentioned in the description.

src/main/resources/latest-spark-swagger-output.yaml
```diff
@@ -920,6 +923,25 @@ paths:
       responses:
         "200":
           description: "successful operation"
+  /otp2/*:
+    get:
+      tags:
+      - "otp2"
+      description: "Forwards any GET request to OTP. Refer to <a href='http://otp-docs.ibi-transit.com/api/index.html'>OTP's\
```
binh-dam-ibigroup: Version 2 should be mentioned in this description.

binh-dam-ibigroup (commented):
Can the modified `proxy` method be made non-static so we can continue to use the `this::proxy` notation? Also mention version 2 in the description of the added OTPv2 endpoints.

binh-dam-ibigroup (changes requested):
Could you address https://github.com/ibi-group/otp-middleware/pull/155#pullrequestreview-783433784 please?

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
leonardehrenfried: Done.

src/main/resources/latest-spark-swagger-output.yaml
leonardehrenfried: Good point.

src/main/resources/latest-spark-swagger-output.yaml
leonardehrenfried: Good point.

leonardehrenfried:
Is it possible that the failing test assertion which has the following comment
```
// Rounding hour to nearest whole day bumps this by a day from the original 2 to 3 days.
```
fails because it's run at a time when the rounding doesn't work as expected?

leonardehrenfried:
I've cherry-picked a commit by @br648 from another branch since the test assertion is different depending on what time of day you run the tests.

binh-dam-ibigroup (approved):
LGTM

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -18,58 +18,74 @@
...
+        }
+        /**
+         * URI location of the OpenTripPlanner API (e.g., https://otp-server.com/otp). Requests sent to this URI should
+         * return OTP version info.
+         */
+        public String uri() {
+            return uri;
+        }
+        public String toString() { return name; }
+    }
```
evansiroky: It seems this enum could be in its own separate file, but I don't have a super strong opinion about this.

evansiroky (approved):
Just one non-blocking comment.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
leonardehrenfried: Good point.

leonardehrenfried:
I've fixed Evan's comment and tehrefore will merge.

PR #159, opened 2021-12-16
fix(deps): Use log4j 2.16.0 with spark-swagger.

binh-dam-ibigroup:
### Checklist

- [na] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented

### Description

This PR temporarily addresses the log4j CVE-2021-44228 vulnerability from use of spark-swagger, until spark-swagger gets that patch. With the change, the correct version of log4j will be inserted in the final jar.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -261,6 +261,12 @@
             <artifactId>spark-swagger</artifactId>
             <version>1.0.0.48</version>
         </dependency>
+        <!-- temporary override for spark-swagger log4j dependency -->
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.16.0</version>
+        </dependency>
 
         <!-- Auth0 Client library for using Management API  -->
         <dependency>


philip-cline (approved):

leonardehrenfried (approved):

br648:
@binh-dam-ibigroup This update still includes log4j version 1 in the final jar as it is still a dependency of spark-swagger (mvn dependency:tree). The class used to instantiate the logging class is different between the two versions:

Log4j: 

```
import org.apache.log4j.Logger;
private static Logger LOG = Logger.getLogger(Clazz.class);
```

Log4j2:

```
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;
private static Logger LOG = LogManager.getLogger(Clazz.class);
```

Without Log4j version 1, I suspect spark-swagger would fail. I believe the only thing we can do is wait for the third party update.

leonardehrenfried:
Log4j version 1 is not susceptible to the vulnerability, only 2.0.0 - 2.15.0: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228

If I'm reading the above report correctly then having log4j 1 on the classpath isn't a security risk.

Can you confirm this?

br648:
Log4j 1 is EOL and has other vulnerabilities which haven't been addressed. The recomendation is to migrate to version 2.

leonardehrenfried:
In such a case it would not have been necessary to even override it. Sorry, @binh-dam-ibigroup.

leonardehrenfried:
Good point. Lets hope the author will cut a release soon.

PR #170, opened 2022-06-24
CDP refactor

br648:
### Checklist

### Description

This CDP refactor changes trip request processing from daily to hourly. It also reduces the amount of data that is anonymized and uploaded to S3. The following parameter has been added to allow CDP to be disabled:

CONNECTED_DATA_PLATFORM_ENABLED

Prior to this update being deployed, the contents of the TripHistoryUpload table must be removed.


src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +70,122 @@ public AnonymizedTripRequest() {
...
         }
+
+        // Extract all trip summary itineraries, convert to anonymized itineraries and group.
+        int tripSummaryId = 1;
+        for (TripSummary tripSummary : tripSummaries) {
+            if (tripSummary.error != null) {
+                // If trip summary has an error, add it to the anonymized trip request and don't attempt to process the
+                // trip summary itineraries, because there won't be any.
+                this.error = tripSummary.error;
+                break;
```
miles-grant-ibigroup, line 96: Won't this cause `tripSummaryId` to get out of sync?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +70,122 @@ public AnonymizedTripRequest() {
...
+                }
+                itin.legs.add(anonymizedLeg);
+            }
+            anonymizedItineraries.add(itin);
+        }
+        return anonymizedItineraries;
     }
+
+    /**
+     * Create {@link Coordinates} for a place. Replace lat/lon values with the lat/lon values created for the trip
```
miles-grant-ibigroup: Does this remove the coordinate anonymization feature previously present?

miles-grant-ibigroup (commented):
Thanks for these changes! Looks good in my initial testing, just a few small detail questions!

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: My take on this, and it could be wrong, is that if one of the trip summaries has an error they all will. So, if the first trip summary has an error, record this and then don't process the itineraries or subsequent trip summaries. In this case`tripSummaryId` is redundant.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: No that is still present. The method `ConnectedDataManager.getPlaceCoordinates` works out if the first and last coords should be randomized. These coords are then passed in to `AnonymizedTripRequest` and ultimately into `AnonymizedTripRequest.getPlaceCoordinates` which I've renamed to `AnonymizedTripRequest.getLegCoordinates`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
miles-grant-ibigroup: Oh yeah this makes sense thanks!

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: @miles-grant-ibigroup the latest refactor does remove the randomizing of coords. We just don't provide them now.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -137,55 +134,98 @@ private List<AnonymizedItinerary> getItineraries(
...
+     * Define the position of the first transit leg. If all walking legs, return the number of legs plus one to
+     * represent all walk legs.
+     */
+    private int getFirstTransitLeg(List<Leg> legs) {
+        for (int i = 0; i <= legs.size(); i++) {
+            if (Boolean.TRUE.equals(legs.get(i).transitLeg)) {
+                return i;
+            }
+        }
+        return legs.size() + 1;
```
miles-grant-ibigroup: Why doesn't this overflow? Wouldn't this return `2` for a single walking leg?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -137,55 +134,98 @@ private List<AnonymizedItinerary> getItineraries(
...
+    }
+
+    /**
+     * Define the position of the last transit leg. If all walking legs, return zero to represent all walk legs.
```
miles-grant-ibigroup: The code will correctly also strip locations from a bike or escooter leg (since they are not transit legs) but the comment should reflect this as well

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -137,55 +134,98 @@ private List<AnonymizedItinerary> getItineraries(
...
-        Coordinates tripRequestCoordinates
-    ) {
-        return (!isTransitLeg && isFirstOrLastLegOfTrip) ? tripRequestCoordinates : new Coordinates(place.lat, place.lon);
+    private void processLegCoordinates(List<Leg> legs) {
+        int firstTransitLeg = getFirstTransitLeg(legs);
+        int lastTransitLeg = getLastTransitLeg(legs);
+        for (int i = 0; i <= legs.size() -1; i++) {
+            Leg leg = legs.get(i);
+            if (
+                leg.mode.equalsIgnoreCase("walk") &&
```
miles-grant-ibigroup: All non-transit legs should have their locations stripped! Is there any way to use the `isTransit` leg property?

miles-grant-ibigroup (changes requested):
This looks great! Just a few quick questions, although it may be me just misunderstanding things

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: I've updated this along with the comments, hopefully it is clearer now.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: Updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: Yep! Updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
miles-grant-ibigroup: Ah that makes sense now thank you!

miles-grant-ibigroup (approved):
Looks great, thanks for the changes!

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedLeg.java
```diff
@@ -30,9 +34,6 @@ public class AnonymizedLeg {
     public String tripId;
 
     // Parameters for non transit leg.
-    public Boolean hailedCar;
-    public Boolean rentedBike;
-    public Boolean rentedCar;
     public Boolean rentedVehicle;
```
binh-dam-ibigroup, line 37: Is `rentedVehicle` still needed?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
```diff
@@ -10,6 +10,16 @@
  */
 public class AnonymizedItinerary {
 
+    /**
+     * Reference to identify which trip summary an itinerary is associated with.
+     */
+    public Integer tripSummaryId;
+
+    /**
+     * Unique itinerary id.
```
binh-dam-ibigroup: Looking at `AnonymizedTripRequest`, this comment should state that the id is unique for the trip summary/request.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +69,162 @@ public AnonymizedTripRequest() {
...
+
+    /**
+     * Define the coordinates for all legs. If the leg is non-transit and is before the first transit leg or after
+     * the last transit leg the coordinates must be removed for privacy. Any transit legs or non-transit legs
+     * between the first and last transit legs, the coordinates can be provided.
+     */
+    private void processLegCoordinates(List<Leg> legs) {
+        int firstTransitLeg = getFirstTransitLeg(legs);
+        // No need to find the last transit leg if the first transit leg indicates that all legs are non-transit.
+        int lastTransitLeg = (firstTransitLeg == Integer.MAX_VALUE) ? Integer.MIN_VALUE : getLastTransitLeg(legs);
```
binh-dam-ibigroup: Suffix these two variable names with `Index`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +69,162 @@ public AnonymizedTripRequest() {
...
+        for (int i = 0; i <= legs.size() -1; i++) {
+            Leg leg = legs.get(i);
+            if (Boolean.TRUE.equals(!leg.transitLeg) && (i < firstTransitLeg || i > lastTransitLeg)) {
```
binh-dam-ibigroup: Can we just write `if (!leg.transitLeg...`?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -117,129 +122,127 @@ public static void removeUsersTripHistory(String userId) {
...
+        FindIterable<TripSummary> tripSummaries = Persistence.tripSummaries.getFiltered(
+            eq(batchIdFieldName, uniqueBatchId),
+            Sorts.descending(dateCreatedFieldName)
+        );
+        // Anonymize trip request.
+        return new AnonymizedTripRequest(
+            tripRequest,
+            tripSummaries,
+            getPlaceCoordinates(tripSummaries, true, tripRequest.fromPlace),
+            getPlaceCoordinates(tripSummaries, false, tripRequest.toPlace)
```
binh-dam-ibigroup: Because this is the only call to this constructor, consider moving the last two arguments into the constructor.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -285,34 +290,35 @@ private static boolean isLegTransit(List<Itinerary> itineraries, boolean isFirst
     }
 
     /**
-     * Obtain anonymized trip data for the given date, write to zip file, upload the zip file to S3 and finally delete
+     * Obtain anonymize trip data for the given hour, write to zip file, upload the zip file to S3 and finally delete
```
binh-dam-ibigroup: Revert to "Obtain anonymized trip data", or just say "~Obtain~Anonymize trip data".

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -359,4 +366,10 @@ public static boolean shouldProcessTripHistory(String configBucketName) {
         return !Strings.isBlank(configBucketName);
     }
 
+    /**
+     * Enable connected data platform if configured to do so.
+     */
+    private static boolean enableConnectedDataPlatform() {
```
binh-dam-ibigroup: Rename to `isConnectedDataPlatformEnabled` and update JavaDoc accordingly.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
```diff
@@ -3,58 +3,67 @@
...
-                    Persistence.tripHistoryUploads.create(new TripHistoryUpload(day.atStartOfDay().toLocalDate()));
-                }
-            });
+            // Stage first ever upload hour.
+            Persistence.tripHistoryUploads.create(new TripHistoryUpload(previousWholeHourFromNow));
+            LOG.debug("Staging first ever upload hour: {}.", previousWholeHourFromNow);
+            return;
+        }
+        // Stage all hours between the last hour uploaded and an hour ago.
+        List<LocalDateTime> betweenHours = DateTimeUtils.getHoursBetween(lastCreated.uploadHour, previousWholeHourFromNow);
```
binh-dam-ibigroup, line 41: Previously created `TripHistoryUpload` will not have field `uploadHour`, so the job will fail.
What do you recommend to handle existing entries in the `TripHistoryUpload` collection?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +69,162 @@ public AnonymizedTripRequest() {
...
         this.batchId = tripRequest.batchId;
         this.fromPlace = fromCoordinates;
         this.toPlace = toCoordinates;
-        if (tripRequest. requestParameters != null) {
+        if (tripRequest.requestParameters != null) {
             this.date = tripRequest.requestParameters.get("date");
             this.time = tripRequest.requestParameters.get("time");
-            String tripType = tripRequest.requestParameters.get("arriveBy");
-            if (tripType != null && tripType.equalsIgnoreCase("true")) {
+            String tripRequestType = tripRequest.requestParameters.get("arriveBy");
```
binh-dam-ibigroup: Rename variable to `isArriveBy`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -50,24 +69,162 @@ public AnonymizedTripRequest() {
...
+            }
+        }
+    }
+
+    /**
+     * Define the position of the first transit leg. If all legs are non-transit return {@link Integer#MAX_VALUE} to
+     * represent this. This will then force the calling method to remove coordinates from all legs.
+     */
+    private int getFirstTransitLeg(List<Leg> legs) {
+        for (int i = 0; i <= legs.size(); i++) {
```
binh-dam-ibigroup: `i <= legs.size() - 1`

src/main/java/org/opentripplanner/middleware/utils/Coordinates.java
```diff
@@ -6,14 +6,14 @@
...
-    public double lon;
+    public Double lon;
+    public Double lat;
 
     /** Required for JSON serialization */
     public Coordinates() {
     }
 
-    public Coordinates(double lat, double lon) {
+    public Coordinates(Double lat, Double lon) {
```
binh-dam-ibigroup, line 16: Instead of allowing nulls, would it be better to not provide/persist an instance of `Coordinates` in both `AnonymizedTripRequest` and `AnonymizedLeg` the trip if both `lat` and `lon` are null?

binh-dam-ibigroup (changes requested):
Looks pretty solid, but I have two blocking comments, one regarding handling of existing upload entries in Mongo, one regarding an index out of bounds both leading to crashes.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedLeg.java
br648: I think so yes. Point 5 from Caleb here: https://docs.google.com/document/d/1gELtMoJNxSnhAYilCrCfEIBzlNj9h2pkAEycqX1mh8A/edit#heading=h.44jfqj3j3g8t

Although, if the mode value has to be checked each time to workout the rented vehicle perhaps not. Thoughts?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
br648: The value that is unique for the trip summary/request is the `tripSummaryId` value above this one. I have updated the comment on that parameter instead.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: I could but SonarLint suggests doing it this way because the Boolean value is a class and not a primitive. Happy to change this and others if you think it is cleaner?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
br648: My understanding is that this never made it passed soak testing so the existing values can be deleted. If that is not the case quite a lot of work will be required to handle it. Let me find out about the first point before I implement the second.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: Good spot, thanks.

src/main/java/org/opentripplanner/middleware/utils/Coordinates.java
br648: That would be cleaner. Updated.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedLeg.java
binh-dam-ibigroup: Not sure about the requirement to check the vehicle type, so this should be fine for now.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedItinerary.java
binh-dam-ibigroup: Sure. But then maybe it's best to rename `itineraryId` to `itineraryIndex` because it is more like a sequence number? (and update that field's JavaDoc accordingly)

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -67,22 +68,17 @@ public class AnonymizedTripRequest {
     public AnonymizedTripRequest() {
     }
 
-    public AnonymizedTripRequest(
-        TripRequest tripRequest,
-        FindIterable<TripSummary> tripSummaries,
-        Coordinates fromCoordinates,
-        Coordinates toCoordinates
-    ) {
+    public AnonymizedTripRequest(TripRequest tripRequest,FindIterable<TripSummary> tripSummaries) {
```
binh-dam-ibigroup: Insert space after comma.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -103,6 +99,49 @@ public AnonymizedTripRequest(
         }
     }
 
+    /**
+     * Workout if the first or last leg is a transit leg. If the leg is a transit leg the coordinates provided by OTP
+     * can be used. If not they are removed. The place value is assumed to be in the format 'location :: lat,lon'.
```
binh-dam-ibigroup: Remove spaces surrounding `::`.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -103,6 +99,49 @@ public AnonymizedTripRequest(
...
+        );
+    }
+
+    /**
+     * Using the legs from the first itinerary, define whether the first or last leg is a transit leg. It is assumed
+     * that the first and last legs are the same for all itineraries. If the leg is transit, return true else false.
+     * E.g. If the first leg is non transit, the related 'fromPlace' lat/lon is removed because it is not a public
+     * location.
+     */
+    private static boolean isLegTransit(List<Itinerary> itineraries, boolean isFirstLeg) {
```
binh-dam-ibigroup: This function name is not clear. Maybe rename to `isAnyFirstOrLastLegTransit`, and also have the function scan for all itineraries instead of just the first one?

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -103,6 +99,49 @@ public AnonymizedTripRequest(
...
+        if (itineraries != null &&
+            !itineraries.isEmpty() &&
+            itineraries.get(0).legs != null &&
```
binh-dam-ibigroup: Extract variable for `itineraries.get(0)`

src/test/resources/snapshots/org/opentripplanner/middleware/ConnectedDataPlatform/ConnectedDataPlatformTest/canCreateZipFileForTripSummaryWithError-0.json
```diff
@@ -1 +1 @@
-"[{\"tripRequest\":{\"batchId\":\"783726\",\"fromPlace\":{\"lat\":33.64070037704429,\"lon\":-84.44622866991179},\"toPlace\":{\"lat\":33.64070037704429,\"lon\":-84.44622866991179},\"date\":\" 2021-09-22\",\"time\":\"15:54\",\"tripType\":\"DEPART_AT\",\"mode\":\"WALK,BUS,RAIL\",\"maxWalkDistance\":\"1027\",\"optimize\":\"QUICK\"},\"tripSummaries\":[{\"batchId\":\"783726\",\"tripPlan\":{\"date\":null,\"itineraries\":[],\"errorMessage\":\"Trip is not possible.  You might be trying to plan a trip outside the map data boundary.\"}}]}]"
\ No newline at end of file
+"[{\"batchId\":\"783726\",\"fromPlace\":null,\"toPlace\":null,\"date\":\" 2021-09-22\",\"time\":\"15:54\",\"tripType\":\"DEPART_AT\",\"mode\":[\"WALK\",\"BUS\",\"RAIL\"],\"maxWalkDistance\":\"1027\",\"optimize\":\"QUICK\",\"itineraries\":[],\"error\":{\"id\":400,\"msg\":\"Trip is not possible.  You might be trying to plan a trip outside the map data boundary.\",\"missing\":[\"from\"],\"noPath\":true}}]"
```
binh-dam-ibigroup, line 1: I don't see a need to serialize `fromPlace` or `toPlace` if the value is null, what do you think? (same applies to those fields for legs).

binh-dam-ibigroup (commented):
See the rest of my comments for now, I think we are getting close. Let me know what you decide regarding the old trip upload entries.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
br648: @binh-dam-ibigroup It's that a thumbs up to change or leave?!

src/test/resources/snapshots/org/opentripplanner/middleware/ConnectedDataPlatform/ConnectedDataPlatformTest/canCreateZipFileForTripSummaryWithError-0.json
br648: I think for a consistent payload it should remain.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
binh-dam-ibigroup: It's fine to follow SonarLint.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/TripHistoryUploadJob.java
binh-dam-ibigroup: Any updates on this?

br648: No. I'm waiting on Jon and/or Miles. Both are off at the moment.

binh-dam-ibigroup: Cool. I'll reassign this issue to you until we hear back.

br648: Heard back from Miles. This is only deployed th ALT middleware which isn't used in production. I have therefore added to this PR description to delete the older data before deploying.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -34,40 +44,240 @@ public class AnonymizedTripRequest {
...
+            if (!areAllFirstOrLastLegsTransit(tripSummary.itineraries, isFirstLeg)) {
+                // If any trip summary itinerary first or last leg is not public, do not provide coordinates.
+                return null;
+            }
+        }
+        // The UI might send just the coordinates (if the geocoder does not return anything, which is unlikely).
+        // If that happens, the format will just be lat,lon and :: will not be present.
+        String coords = (place.contains("::")) ? place.split("::")[1].trim() : place;
+        return new Coordinates(
+            Double.parseDouble(coords.split(",")[0]),
```
binh-dam-ibigroup: Extract variable for `coords.split`

src/main/java/org/opentripplanner/middleware/connecteddataplatform/AnonymizedTripRequest.java
```diff
@@ -34,40 +44,240 @@ public class AnonymizedTripRequest {
...
+    /**
+     * Define the coordinates for all legs. If the leg is non-transit and is before the first transit leg or after
+     * the last transit leg the coordinates must be removed for privacy. Any transit legs or non-transit legs
+     * between the first and last transit legs, the coordinates can be provided.
+     */
+    private void processLegCoordinates(List<Leg> legs) {
+        int firstTransitLegIndex = getFirstTransitLeg(legs);
+        // No need to find the last transit leg if the first transit leg indicates that all legs are non-transit.
+        int lastTransitLegIndex = (firstTransitLegIndex == Integer.MAX_VALUE) ? Integer.MIN_VALUE : getLastTransitLeg(legs);
+        for (int i = 0; i <= legs.size() -1; i++) {
```
binh-dam-ibigroup: Insert space before 'minus'.

binh-dam-ibigroup (approved):
Good to go, let's coordinate when you are ready to merge (merging will autodeploy, so we'll have to add the environment variable and update Mongo before hand).

miles-grant-ibigroup:
Environment variables are updated -- I tried to update Mongo but couldn't get access to the relevant databases.

br648:
Tests are failing, I think because OTP isn't available: `OtpDispatcherResponse{statusCode=***00, responseBody='<html><body><h2>***00 Internal Server Error</h2></body></html>'PARSE_EXCEPTION}`

@binh-dam-ibigroup Any idea how to fix?

binh-dam-ibigroup:
> Tests are failing, I think because OTP isn't available: `OtpDispatcherResponse{statusCode=***00, responseBody='<html><body><h2>***00 Internal Server Error</h2></body></html>'PARSE_EXCEPTION}`
> 
> @binh-dam-ibigroup Any idea how to fix?

@br648 Not sure, but it seems that the first error in the e2e tests is about an unmocked OTP request, see https://github.com/ibi-group/otp-middleware/runs/7285669028?check_suite_focus=true#step:9:301. Could you check how this is happening?

miles-grant-ibigroup:
The OTP mock server was having some trouble. Now that this has been resolved the tests pass. I will hold off on merging until we've wiped the appropriate databases.

PR #178, opened 2022-11-16
CDP Support for GraphQL Plan Requests

miles-grant-ibigroup:
### Checklist

- [x] The description lists all relevant PRs included in this release _(remove this if not merging to master)_

### Description

The current CDP only supports queries which are made via the `/plan` REST endpoint. To read graphQL queries we need to read those. This PR adds support for doing so. 

Introduces a new `OTP_GRAPHQL_ENDPOINT` config parameter which can probably be left at the default.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
@@ -227,8 +227,10 @@ public class OtpRequestProcessor implements Endpoint {
      * @return Returns false if there was an error.
      */
     private static boolean handlePlanTripResponse(Request request, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
+        return handlePlanTripResponse(request.queryParams("batchId"), request.queryParams("fromPlace"), request.queryParams("toPlace"), otpDispatcherResponse, otpUser);
+    }
+    private static boolean handlePlanTripResponse(String batchId, String fromPlace, String toPlace, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
         boolean result = true;
-        String batchId = request.queryParams("batchId");
         if (batchId == null) {
             batchId = BATCH_ID_NOT_PROVIDED;
         }
@@ -249,8 +251,8 @@ public class OtpRequestProcessor implements Endpoint {
                 TripRequest tripRequest = new TripRequest(
                     otpUser.id,
                     batchId,
-                    request.queryParams("fromPlace"),
-                    request.queryParams("toPlace"),
+                    fromPlace,
+                    toPlace,
                     otpResponse.requestParameters
                 );
                 // only save trip summary if the trip request was saved
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
@@ -7,6 +7,7 @@ import io.github.manusant.ss.rest.Endpoint;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -26,6 +27,7 @@ import org.opentripplanner.middleware.utils.HttpUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import spark.Request;
+import spark.Response;
 
 import javax.ws.rs.core.MediaType;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -164,14 +167,52 @@ private String proxyPost(Request request, spark.Response response) {
...
+                /*
+                * Inserting the from and to place directly into the graphQL request is *not* supported.
+                * Since these requests are only consumed by the CDP if a user has consented, we know
+                * that all CDP-applicable requests will originate from OTP-RR or other projects
+                * which are aware of the middleware's existence. We can therefore assume that
+                * the requests will use graphQL variables, the way that OTP-RR does.
+                *
+                * Other requests will still be proxied, just not stored.
+                */
+                Map graphQlVariables = (Map) getPOJOFromJSON(requestBody, Map.class).get("variables");
```
br648: Repalce `Map` with `HashMap`.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -164,14 +167,52 @@ private String proxyPost(Request request, spark.Response response) {
...
+
+                String fromPlace = (String) graphQlVariables.get("fromPlace");
+                String toPlace = (String) graphQlVariables.get("toPlace");
+
+                // Follows the method used in otp-ui core-utils storage.js
+                String randomBatchId = Integer.toString((int) (Math.random() * 1_000_000_000), 36);
```
br648, line 198: Replace `Math.random()` with `Random.nextInt()`

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -164,14 +167,52 @@ private String proxyPost(Request request, spark.Response response) {
...
+                if(!handlePlanTripResponse(randomBatchId, fromPlace, toPlace, otpDispatcherResponse, otpUser)) {
+                    logMessageAndHalt(
+                            request,
+                            HttpStatus.INTERNAL_SERVER_ERROR_500,
+                            "Failed to save trip history."
+                    );
+                    return null;
+                }
+            } catch(JsonProcessingException e) {
+                LOG.warn("Invalid GraphQL Request received. Still passing to OTP2: " + e.getMessage());
```
br648: Don't use String concat, use `LOG.warn("message {}", e.getMessage());`

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -164,14 +167,52 @@ private String proxyPost(Request request, spark.Response response) {
...
+            } catch(NullPointerException e) {
+                LOG.warn("Failed to read variables from GraphQL Plan request. Still passing to OTP2: " + e.getMessage());
```
br648: Don't use String concat, use LOG.warn("message {}", e.getMessage());

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -227,8 +268,10 @@ private OtpUser checkUserPermissions(Request request) {
      * @return Returns false if there was an error.
      */
     private static boolean handlePlanTripResponse(Request request, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
+        return handlePlanTripResponse(request.queryParams("batchId"), request.queryParams("fromPlace"), request.queryParams("toPlace"), otpDispatcherResponse, otpUser);
```
br648: Put each param on it's own line to aid readability.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -227,8 +268,10 @@ private OtpUser checkUserPermissions(Request request) {
...
+    }
+    private static boolean handlePlanTripResponse(String batchId, String fromPlace, String toPlace, OtpDispatcherResponse otpDispatcherResponse, OtpUser otpUser) {
```
br648: Add new line between methods. Put each param on it's own line to aid readability.

br648 (changes requested):
SonarLint (https://plugins.jetbrains.com/plugin/7973-sonarlint) has picked up most of these comments. It is worth installing. 

Apart from that, I was going to suggest added a test or two to cover this refacor, but I can't see any original tests in this area, so not so sure! Add a test if you think it would be beneficial.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
miles-grant-ibigroup: I tried this but it didn't behave correctly!

br648 (approved):
LGTM

daniel-heppner-ibigroup (approved):
This all looks good, will be interesting to see how it works with the new GraphQL queries! One  thing that we might consider is that OTP supports a `from` and `to` field in addition to `fromPlace` and `toPlace`. These use a different format and contain just the lat/lng. I don't think we will ever use this in OTP-RR, though.

miles-grant-ibigroup:
Good to know thanks! If in the future we ever use this we can always add more parsing and make it configurable what's enabled and what's not

PR #182, opened 2023-03-16
feat(MonitoredTripController): add support for otp2 to checkItinerary

daniel-heppner-ibigroup:
### Checklist

### Description

This PR adds OTP2 support to the checkItinerary endpoint. It does this with a new query parameter added to the URL to specify the OTP version. The default version is 1 to ensure backwards compatibility.


daniel-heppner-ibigroup:
I am open to better ways to do this, I am not very familiar with the code base as a whole. We could also add otp version to the config since I don't believe we'll ever have a middleware instance being used by two different versions at the same time. 

Also, we will eventually need to upgrade this to use GraphQL and improve the accuracy of this endpoint. This is just a stopgap so there's no regression when we move to OTP2.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -3,10 +3,12 @@
 import io.github.manusant.ss.ApiEndpoint;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.mongodb.client.model.Filters;
+import io.github.manusant.ss.descriptor.ParameterDescriptor;
```
br648: Remove unused import.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -1,6 +1,11 @@
 package org.opentripplanner.middleware.otp;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
```
br648: Remove unused import.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +30,23 @@ public String uri() {
         return uri;
     }
     public String toString() { return name; }
+
+    public static OtpVersion intToVersion(int versionAsInt) {
+        switch(versionAsInt) {
+            case 1:
+            default:
```
br648, line 44: `default` clause should be at the end of the `switch` statement for readability. `case 1:` will obvs have to move as well.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +30,23 @@ public String uri() {
...
+                return OTP1;
+            case 2:
+                return OTP2;
+        }
+    }
+
+    public static OtpVersion getOtpVersionFromRequest(Request request) {
```
br648, line 54: Add method comment.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -171,8 +171,7 @@ public static boolean hasConfigProperty(String name) {
         return hasConfigProperty(envConfig, name);
     }
 
-    /**
-     * Returns true if the given config has the requested property.
+    /** * Returns true if the given config has the requested property.
```
br648: Revert this change.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +30,23 @@ public String uri() {
         return uri;
     }
     public String toString() { return name; }
+
+    public static OtpVersion intToVersion(int versionAsInt) {
```
br648: Change the method to `private`. Consider changing this method name to `getOtpVersion` and the parameter to `int version`.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +30,23 @@ public String uri() {
...
+            case 2:
+                return OTP2;
+        }
+    }
+
+    public static OtpVersion getOtpVersionFromRequest(Request request) {
+        try {
+            return intToVersion(Integer.parseInt(request.queryParamOrDefault("otpVersion", "1")));
+        } catch (NumberFormatException e) {
+            logMessageAndHalt(request, HttpStatus.BAD_REQUEST_400, "Error parsing otpVersion parameter", e);
```
br648: Nit. Full stop.

br648 (changes requested):
A few minor changes. Overall I think this is the best approach over using a config value.

br648 (approved):
Thanks for the changes. All good.

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +29,33 @@ public String uri() {
...
     public String toString() { return name; }
+
+    /**
+     * Converts integer to OTP Version object.
+     * @param version Version number as an integer
+     * @return OtpVersion object
+     */
+    private static OtpVersion getOtpVersion(int version) {
+        switch(version) {
+            case 2:
```
miles-grant-ibigroup, line 41: Don't we have an enum we can use here instead?

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -46,6 +47,7 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
             .post(path("/checkitinerary")
                     .withDescription("Returns the itinerary existence check results for a monitored trip.")
                     .withRequestType(MonitoredTrip.class)
+                    .withQueryParam().withName("otpVersion").withDefaultValue("1").and()
```
miles-grant-ibigroup: Can we use the enum here too?

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
```diff
@@ -25,4 +29,33 @@ public String uri() {
...
+    }
+
+    /**
+     * Gets an OTP version from the Spark request header
+     * @param request Spark request object
+     * @return OtpVersion specified in request
+     */
+    public static OtpVersion getOtpVersionFromRequest(Request request) {
+        try {
+            return getOtpVersion(Integer.parseInt(request.queryParamOrDefault("otpVersion", "1")));
```
miles-grant-ibigroup, line 56: Can we use the enum here as well?

miles-grant-ibigroup (changes requested):
I'm confused -- is this adding OTP2 support? I see that OTP2 as a version is being passed through correctly, but I'm not seeing what we're doing with that new field?

I'm also not entirely sold on using a url parameter to decide whether to use otp1 or otp2. Wouldn't this maybe make more sense as a config parameter? Do we expect one middleware to handle both at once?

src/main/java/org/opentripplanner/middleware/otp/OtpVersion.java
daniel-heppner-ibigroup: I am not sure what you mean. The problem is that the value coming from the API is an int.

daniel-heppner-ibigroup:
> I'm also not entirely sold on using a url parameter to decide whether to use otp1 or otp2. Wouldn't this maybe make more sense as a config parameter? Do we expect one middleware to handle both at once?

@miles-grant-ibigroup That's what I was referring to with this. I'll look into doing it in the config instead of with qp.

> I am open to better ways to do this, I am not very familiar with the code base as a whole. We could also add otp version to the config since I don't believe we'll ever have a middleware instance being used by two different versions at the same time.

miles-grant-ibigroup (changes requested):
Still unclear as to what the OTP2 parameter is actually doing -- what change does it make if you set it to OTP2? Also I wouldn't name the configuration key OTP_VERSION, since both are actively supported at the same time.

Finally, it's a nitpick but would it be possible to replace all `"1"` and `"2"` with `OtpVersion.OTP1`/`OtpVersion.OTP2`?

daniel-heppner-ibigroup:
> since both are actively supported at the same time.

I made this change because you said:

> Wouldn't this maybe make more sense as a config parameter? Do we expect one middleware to handle both at once?

I think at this point I'm as confused as you are.

PR #188, opened 2023-10-10
fix(HttpUtils): Use UTF-8 when forwarding request bodies to OTP

binh-dam-ibigroup:
### Checklist

### Description

Looking for advice on unit tests for these changes if practical.

This PR fixes the OTP request maker so that the body content uses UTF-8 encoding to support all international content.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
+++ b/src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
@@ -21,6 +21,7 @@ import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
 import java.net.URI;
 import java.net.http.HttpTimeoutException;
+import java.nio.charset.StandardCharsets;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.format.DateTimeParseException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
```diff
@@ -102,26 +101,16 @@ public static HttpResponseValues httpRequestRawResponse(URI uri, int timeoutInSe
...
-                    HttpPut putRequest = new HttpPut(uri);
-                    putRequest.setEntity(new StringEntity(bodyContent));
-                    putRequest.setConfig(timeoutConfig);
-                    httpUriRequest = putRequest;
-                } catch (UnsupportedEncodingException e) {
-                    LOG.error("Unsupported encoding type", e);
-                    return null;
-                }
+                HttpPut putRequest = new HttpPut(uri);
+                putRequest.setEntity(new StringEntity(bodyContent, "UTF-8"));
```
br648: Could use `StandardCharsets.UTF_8` instead.

br648 (approved):
The change is good. Not sure how to test this to be honest!

src/main/java/org/opentripplanner/middleware/utils/HttpUtils.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/188/commits/7fdc4ae631ddacd1bea52f2eac6429f5b982a847, thanks!

JymDyerIBI:
Agree, the change looks good. We don't have an `HttpUtilsTest` class at all, but I'll poke around and see how best to test charsets like this.

It looks like this can be exercised by adding something similar to `ApiTestUtils#makePostRequest()` and `#makePutRequest()` methods and exercising those from the unit tests.

JymDyerIBI:
It looks like this can be exercised by adding `ApiTestUtils#makePostRequest()` and `#makePutRequest()` methods and exercising those from the unit tests.

JymDyerIBI:
Digging further, I see that POST and PUT are exercised in `ApiUserFlowTest` and `ConnectedDataPlatformTest` unit tests, so I think that should suffice to show that these changes don't break anything.

JymDyerIBI (approved):
Looks good to me.

PR #189, opened 2023-10-30
For proxied responses, we have been combining headers which we need to dedupe

JymDyerIBI:
Of the duplicates, "Access-Control-Allow-Origin" is the most problematic because browsers stop on that as an error.

### Checklist

### Description

When we proxy a POST request through the middleware we have request headers from the middleware's Spark Framework and we update them with the request headers from the GraphQL endpoint being proxied to.  Unfortunately the `com.sparkjava.spark.Response#header()` method doesn't let you update a header, it creates duplicate headers. So I added a filter to prevent that.

Apparently the RFC for response headers is unclear (and thus undefined) on the legality of duplicate headers, but browsers specifically fail on duplicate `Access-Control-Allow-Origin` headers.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
@@ -216,7 +216,6 @@ public class OtpRequestProcessor implements Endpoint {
         // spark.Response#header() will duplicate rather than update, so filter out duplicates.
         Arrays.stream(otpDispatcherResponse.headers)
             .filter(h -> !response.raw().containsHeader(h.getName()))
-            //.peek(h -> LOG.info("NEW {}={}", h.getName(), h.getValue()))
             .forEach(h -> response.header(h.getName(), h.getValue()));
 
         response.status(otpDispatcherResponse.statusCode);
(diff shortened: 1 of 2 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
```diff
@@ -210,11 +210,15 @@ private String proxyPost(Request request, Response response) {
...
-
         }
 
-        // provide response to requester as received from OTP server
-        Arrays.stream(otpDispatcherResponse.headers).forEach(header -> response.header(header.getName(), header.getValue()));
+        // Add response headers to requester as it was received from OTP server, but beware that
+        // spark.Response#header() will duplicate rather than update, so filter out duplicates.
+        Arrays.stream(otpDispatcherResponse.headers)
+            .filter(h -> !response.raw().containsHeader(h.getName()))
+            //.peek(h -> LOG.info("NEW {}={}", h.getName(), h.getValue()))
```
binh-dam-ibigroup: Remove this line before merging.

binh-dam-ibigroup (approved):
Good finding!

leonardehrenfried (approved):
This code looks fine but I can also make the CORS headers optional in OTP itself (but will increase the risk of a misconfiguration breaking a deployment.)

src/main/java/org/opentripplanner/middleware/controllers/api/OtpRequestProcessor.java
JymDyerIBI: Done.

JymDyerIBI: > Remove this line before merging.

I did and suddenly there are CI failures in the end-2-end testing. This shouldn't triggers those. Rerunning.

PR #203, opened 2023-12-20
Live Tracking Data

br648:
### Checklist

### Description

Jira ticket: https://ibisolutions.atlassian.net/browse/OTP-953

Intial work to update the OTP middleware API to receive, store and provide response for live tracking data.

New config param:

TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS: 5


src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -0,0 +1,10 @@
+package org.opentripplanner.middleware.triptracker;
+
+public enum TripInstruction {
+
```
binh-dam-ibigroup, line 7: Add a comment saying that more instruction types can be added.

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private static boolean isTripAssociatedWithUser(Request request, String tripId) {
+        var user = Auth0Connection.getUserFromRequest(request);
+
+        var monitoredTrip = Persistence.monitoredTrips.getById(tripId);
+        if (
+            monitoredTrip == null ||
+            (user.adminUser != null && !monitoredTrip.userId.equals(user.adminUser.id)) ||
+            (user.otpUser != null && !monitoredTrip.userId.equals(user.otpUser.id)) ||
+            (user.apiUser != null && !monitoredTrip.userId.equals(user.apiUser.id))
```
binh-dam-ibigroup: I think only `OtpUsers` can have monitored trips. An `ApiUser` or an `AdminUser` could create and links trips on behalf of other `OtpUser`s but I don't see a case where they alone would monitor trips, unless they are also `OtpUser`s.

src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
```diff
@@ -0,0 +1,226 @@
...
+        HttpResponseValues response = makeRequest(
+            START_TRACKING_TRIP_PATH,
+            JsonUtils.toJson(createInitialTrackingPayload("unassociated-trip-id")),
+            getMockHeaders(soloOtpUser),
+            HttpMethod.POST
+        );
+
+        var startTrackingResponse = JsonUtils.getPOJOFromJSON(response.responseBody, StartTrackingResponse.class);
+        assertEquals("Monitored trip is not associated with this user!", startTrackingResponse.message);
+        assertEquals(HttpStatus.BAD_REQUEST_400, response.status);
```
binh-dam-ibigroup: Should the response be "forbidden" instead?

binh-dam-ibigroup (commented):
Just a few comments after a first quick glance at this PR. Nothing screaming for far, but I'll have to play more with it.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/ec0152118d75a99b1cdf92e774c3e088927e6273

src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/ec0152118d75a99b1cdf92e774c3e088927e6273

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
br648: That does look the case so have addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/ec0152118d75a99b1cdf92e774c3e088927e6273. I can revert later if need be.

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
```diff
@@ -0,0 +1,137 @@
...
+        this.path = apiPrefix + SECURE + "monitoredtrip";
+    }
+
+    @Override
+    public void bind(final SparkSwagger restApi) {
+        restApi.endpoint(
+                endpointPath(path).withDescription("Interface for tracking monitored trips."),
+                HttpUtils.NO_FILTER
+            )
+            .post(path("/starttracking")
```
binh-dam-ibigroup, line 46: Could you add the request body (`TrackingPayload`) and response template for all three endpoints, so that it appears in the Swagger docs?

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Get the expected tracking payload for the request. The populated contents of the payload will be determined by
+     * the trip stage.
+     */
+    private static TrackingPayload getTrackingPayloadFromRequest(Request request) {
+        TrackingPayload payload;
+        try {
+            payload = getPOJOFromRequestBody(request, TrackingPayload.class);
```
binh-dam-ibigroup: `return` the result here and inline `payload`.

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
```diff
@@ -0,0 +1,137 @@
...
+        } catch (JsonProcessingException e) {
+            logMessageAndHalt(request, HttpStatus.BAD_REQUEST_400, "Error parsing JSON tracking payload.", e);
+            return null;
+        }
+        return payload;
```
binh-dam-ibigroup: I would say `return null` as fallback.

src/main/java/org/opentripplanner/middleware/models/TrackedJourney.java
```diff
@@ -0,0 +1,51 @@
...
+import org.opentripplanner.middleware.triptracker.TrackingPayload;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TrackedJourney extends Model {
+
+    public String journeyId;
```
binh-dam-ibigroup: We should just use the `id` of the record (`_id` in Mongo) instead of the extra `journeyId` field... unless there a reason to use a separate id? We can still assign the `id` field in Mongo when writing for the first time.

src/main/java/org/opentripplanner/middleware/triptracker/TrackingLocation.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public int bearing;
+
+    public Double lat;
+
+    public Double lon;
+
+    public int speed;
+
+    public long timestamp;
```
binh-dam-ibigroup: Any opinion for storing the timestamp as a date in Mongo, but still passing it around as a number?

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public static final int TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS
+        = getConfigPropertyAsInt("TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS", 5);
+
+    /**
+     * Start tracking by providing a unique journey id and tracking update frequency to the caller.
+     */
+    public static StartTrackingResponse startTrackingTrip(TrackingPayload payload) {
+
```
binh-dam-ibigroup: I think if there is a tracked journey that is ongoing (i.e. not ended) for this trip, then starting a new tracking should not be permitted. (Otherwise, someone could generate an infinite number of journey ids and that would be a problem).

src/main/java/org/opentripplanner/middleware/triptracker/TrackingPayload.java
```diff
@@ -0,0 +1,25 @@
+package org.opentripplanner.middleware.triptracker;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Trip tracking payload that covers the expect parameters for starting, updating and ending tracking.
+ */
+public class TrackingPayload {
```
binh-dam-ibigroup: For documentation (Swagger) purposes, I think we should separate the payload types.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Update the tracking location information provided by the caller.
+     */
+    public static UpdateTrackingResponse updateTracking(TrackingPayload payLoad, TrackedJourney trackedJourney) {
+
+        // Update tracked journey.
+        trackedJourney.update(payLoad);
+        Persistence.trackedJourneys.replace(trackedJourney.id, trackedJourney);
```
binh-dam-ibigroup: It is just the `locations` field that is being explicitly updated, correct? Try using `Persistence.update` instead and pass the list with all the locations, and this will also update the `lastUpdated` field (right now that field is not getting updated).

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -0,0 +1,92 @@
...
+            getTripStatus(TripStage.UPDATE)
+        );
+    }
+
+    /**
+     * End tracking by saving the end condition and date.
+     */
+    public static void endTracking(TrackedJourney trackedJourney) {
+        trackedJourney.end("Tracking terminated by user.");
+        Persistence.trackedJourneys.replace(trackedJourney.id, trackedJourney);
```
binh-dam-ibigroup: Same here, try to use `Persistence.update`, passing the updated fields.

binh-dam-ibigroup (changes requested):
I think the payload types need to be split. After that, add the Swagger documentation for the payloads and responses for the new endpoints. Only one active `TrackedJourney` should be allowed for a trip, and consider removing `journeyId` in favor of the `id` field in Mongo. Nice-to-have is to use `Persistence.update` to update `TrackedJourney` records.

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
br648: Updated for all apart from the end tracking response, bc there isn't one.

src/main/java/org/opentripplanner/middleware/triptracker/TrackingLocation.java
br648: I wasn't sure how the JSON to POJO marshalling would work, but it seems to handle the conversion of a long into a Date object so have updated this to a Date.

br648:
@binh-dam-ibigroup have you seen this issue before? This is related to defining the response: `.withResponses(SwaggerUtils.createStandardResponses(StartTrackingResponse.class))`

`Resolver error at paths./api/secure/monitoredtrip/starttracking.post.responses.200.responseSchema.$ref
Could not resolve reference: #/definitions/StartTrackingResponse`

I'm getting the same with with `UpdateTrackingResponse`.

This is based on the following in `MonitoredTripController` which works `.withResponses(SwaggerUtils.createStandardResponses(ItineraryExistence.class))`

I've try a host of things with no luck. I thought I'd get your thoughts on this before I use up anymore time.

binh-dam-ibigroup:
> @binh-dam-ibigroup have you seen this issue before? This is related to defining the response: `.withResponses(SwaggerUtils.createStandardResponses(StartTrackingResponse.class))`
> 
> `Resolver error at paths./api/secure/monitoredtrip/starttracking.post.responses.200.responseSchema.$ref Could not resolve reference: #/definitions/StartTrackingResponse`
> 

This seems to be working for me:
image

binh-dam-ibigroup:
> > @binh-dam-ibigroup have you seen this issue before? This is related to defining the response: `.withResponses(SwaggerUtils.createStandardResponses(StartTrackingResponse.class))`
> > `Resolver error at paths./api/secure/monitoredtrip/starttracking.post.responses.200.responseSchema.$ref Could not resolve reference: #/definitions/StartTrackingResponse`
> 
> This seems to be working for me

I see. It's been a while, I'm looking into why the response type is not correctly added.

br648:
@binh-dam-ibigroup My bad, I was referring to when I run otp-middleware locally and browse to http://localhost@4567. Do you see the same issue?

binh-dam-ibigroup:
> @binh-dam-ibigroup My bad, I was referring to when I run otp-middleware locally and browse to http://localhost@4567. Do you see the same issue?

Yes, same indeed

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
```diff
@@ -40,98 +37,56 @@ public TrackedTripController(String apiPrefix) {
...
                 endpointPath(path).withDescription("Interface for tracking monitored trips."),
                 HttpUtils.NO_FILTER
             )
             .post(path("/starttracking")
                     .withDescription("Initiates the tracking of a monitored trip.")
                     .withProduces(JSON_ONLY)
-                    .withResponses(SwaggerUtils.createStandardResponses()),
+                    .withRequestType(StartTrackingPayload.class)
+                    .withResponses(SwaggerUtils.createStandardResponses(StartTrackingResponse.class)),
                 (request, response) -> trackTrip(request, TripStage.START), JsonUtils::toJson)
```
binh-dam-ibigroup: Do we need to funnel individual stages to `trackTrip` if it is just to split them later?

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -89,4 +140,74 @@ private static String getTripStatus(TripStage tripStage) {
                 return TripStatus.NO_STATUS.name();
         }
     }
+
+    /**
+     * Confirm that the monitored trip (that the user is on) belongs to them.
```
binh-dam-ibigroup: Remove parentheses, I think.

src/main/java/org/opentripplanner/middleware/triptracker/payload/EndTrackingPayload.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.opentripplanner.middleware.utils.JsonUtils.logMessageAndHalt;
+
+public class EndTrackingPayload {
+
+    public String journeyId;
+
+    /**
+     * Get the expected tracking payload for the request.
+     */
+    public static EndTrackingPayload getPayloadFromRequest(Request request) {
```
binh-dam-ibigroup: This method is duplicated with the other states. Could we use generics here?

src/main/java/org/opentripplanner/middleware/models/TrackedJourney.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public TrackedJourney(StartTrackingPayload trackingPayload) {
+        startTime = new Date();
+        locations.add(trackingPayload.location);
+        tripId = trackingPayload.tripId;
+    }
+
+    public void update(UpdatedTrackingPayload trackingPayload) {
+        this.locations.addAll(trackingPayload.locations);
```
binh-dam-ibigroup: Probably for another PR: if two update requests with identical bodies are sent one after the other, should we add the same locations from the two requests twice?

binh-dam-ibigroup:
> @binh-dam-ibigroup have you seen this issue before? This is related to defining the response: `.withResponses(SwaggerUtils.createStandardResponses(StartTrackingResponse.class))`
> `Resolver error at paths./api/secure/monitoredtrip/starttracking.post.responses.200.responseSchema.$ref Could not resolve reference: #/definitions/StartTrackingResponse`

I figured out what is going on. Basically, spark-swagger only seems to generate the response type if it is part of a GET method.  You can see that for e.g. `VerificationResult` (for SMS verification) or `ItineraryExistence` (because it is GETted as part of the `monitoredtrip` endpoint). Adding `withResponseType` to a POST method does not work, unfortunately.

One way to address that is to add a dummy GET method to the endpoint, for instance:
```
            .get(path("/starttracking-unused")
                .withResponseType(StartTrackingResponse.class),
                (request, response) -> null, JsonUtils::toJson
            )
```
but then that GET endpoint will be added to the documentation, and you'll have to add the code to remove it.

The better/easier approach would be temporily add the dummy GET, then add the desired return types to the `definitions` section of file `src/main/resources/swagger-supplements.yaml`, then remove the dummy GET.

binh-dam-ibigroup:
> The better/easier approach would be temporily add the dummy GET, then add the desired return types to the `definitions` section of file `src/main/resources/swagger-supplements.yaml` ...

You should probably add comments in that file explaining why those types were added.

binh-dam-ibigroup (changes requested):
The code is generally working well.
- For the swagger documentation: see https://github.com/ibi-group/otp-middleware/pull/203#issuecomment-1877681951
- See some refactoring suggestions also.
- Keep the issue about persisting the same sequence of locations multiple times in mind but that comment is not blocking for this PR.

src/main/java/org/opentripplanner/middleware/models/TrackedJourney.java
br648: I've addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/bf853cbd1f8120460c6c026ad633b9ad8940b604. Let me know what you think!

src/main/java/org/opentripplanner/middleware/triptracker/payload/EndTrackingPayload.java
br648: Good point. Addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/bf853cbd1f8120460c6c026ad633b9ad8940b604

src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/203/commits/bf853cbd1f8120460c6c026ad633b9ad8940b604

br648:
@binh-dam-ibigroup Using just `.withResponseType` seems to be a good work-around. See: src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java

src/main/java/org/opentripplanner/middleware/models/TrackedJourney.java
```diff
@@ -45,8 +46,14 @@ public TrackedJourney(StartTrackingPayload trackingPayload) {
         tripId = trackingPayload.tripId;
     }
 
+    /**
+     * Remove duplicates before adding to tracked locations.
+     */
     public void update(UpdatedTrackingPayload trackingPayload) {
-        this.locations.addAll(trackingPayload.locations);
+        LinkedHashSet<TrackingLocation> trackedLocations = new LinkedHashSet<>(locations);
```
binh-dam-ibigroup: Let's just revert that for now. I think there is more to that. My previous suggestion was more about handling if the same update request is sent twice but that's more a nice-to-have.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -210,4 +214,17 @@ private static TrackedJourney getOngoingTrackedJourney(String tripId) {
             )
         );
     }
+
+    /**
+     * Get the expected tracking payload for the request.
+     */
+    private static <T> T getPayloadFromRequest(Request request, Class<T> payloadClass) {
```
binh-dam-ibigroup, line 221: Nice

src/main/java/org/opentripplanner/middleware/triptracker/TrackingLocation.java
```diff
@@ -27,4 +28,21 @@ public TrackingLocation(int bearing, Double lat, Double lon, int speed, Date tim
         this.speed = speed;
         this.timestamp = timestamp;
     }
+
+    /**
+     * The time stamp has been omitted from the equals and hashCode methods. This is to prevent duplicate location
```
binh-dam-ibigroup: Let's revert that and think through it.

binh-dam-ibigroup (commented):
Sorry, could you revert the changes regarding duplicate location tracking?

binh-dam-ibigroup:
> @binh-dam-ibigroup Using just `.withResponseType` seems to be a good work-around. See: src/main/java/org/opentripplanner/middleware/controllers/api/TrackedTripController.java

Good finding!

binh-dam-ibigroup (approved):
I think this is a good start!

JymDyerIBI (approved):
👍  Looks good to me.

PR #212, opened 2024-01-12
 OTP-983 Test Short Time Format

JymDyerIBI:
### Checklist

### Description

Java’s “short” time format has changed to conform with the Unicode "CLDR 42.0" spec. The space character in a string like “5:30 PM” is no longer a simple ASCII space but a “narrow non-breaking space.” ASCII space is U+20 and nnbsp is U+202F under Unicode. Here's where the change was introduced:

https://bugs.openjdk.org/browse/JDK-8304925

Unit tests comparing our time formats failed under JDK 20.

### Solution

Using a regular expression pattern so we can detect either the nnbsp or the ASCII space, unit tests can  work for all JDKs. Here's an example regular expression to match against:
```
"[\\u202f ]PM" 
```
Another “short” format change in JDK 20 is that Chinese is now using 24hr format (making it consistent with the Javascript Format.JS used in the UI). So a regular expression for 5:44 PM would look like this:
```
 "(17|下午5):44"
```

### DIFF
--- a/src/test/java/org/opentripplanner/middleware/models/TripMonitorNotificationTest.java
+++ b/src/test/java/org/opentripplanner/middleware/models/TripMonitorNotificationTest.java
@@ -9,7 +9,8 @@ import java.time.ZonedDateTime;
 import java.util.Date;
 import java.util.Locale;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.text.MatchesPattern.matchesPattern;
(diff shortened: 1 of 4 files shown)


JymDyerIBI:
Commits have passed the Maven CI, which uses an older JDK and thereby demonstrates that the code is backward-compatible. Code also satisfies Maven builds under JDK 20, which I have done by hand since the CI is not set up to test multiple JDKs.

binh-dam-ibigroup (approved):
Good finding!

br648 (approved):
Good spot. Works for JDK 11.

PR #214, opened 2024-02-02
Expected Traveller Positioning

br648:
### Checklist

### Description

This PR attempts to estimate where a traveller is based on cross referencing the current time with the start/end of a leg and points along it. The expected position is then compared to the travellers current position to provide a trip status e.g. on track, deviated etc.

There are certainly shortcomings with regard to speed offsetting the true expected position which needs to be discussed.

There is also a lot of repetitive processing which could be avoided by doing once and saving to the DB.


src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,186 @@
...
+            }
+            segmentStartTime = segmentEndTime;
+        }
+        return null;
+    }
+
+    /**
+     * Convert the time in seconds to milliseconds.
+     */
+    public static long getTimeInMilliseconds(double timeInSeconds) {
```
binh-dam-ibigroup: Rename to `secondsToMilliseconds` or `secondsToMillis`

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,186 @@
...
+            Point.at(Coordinate.fromDegrees(end.lat), Coordinate.fromDegrees(end.lon))
+        );
+    }
+
+    /**
+     * Calculate the distance between each position and from this the number of seconds spent in each segment. If the
+     * time spent in a segment is less than five seconds, group these segments. This assumes that the leg is being
+     * traversed at a constant speed for simplicity.
+     *
+     * TODO: Need to include speed!
```
binh-dam-ibigroup: Is `durationPerMeter` the speed?

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,186 @@
...
+     * Remove all duplicate points.
+     */
+    public static List<Position> removeDuplicatePoints(List<Position> positions) {
+        return new ArrayList<>(new HashSet<>(positions));
+    }
+
+    /**
+     * Order points to match leg traversal.
+     */
+    public static SortedSet<Position> orderPoints(List<Position> positions) {
```
binh-dam-ibigroup: What was the need for ordering points on a leg geometry?

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -237,4 +251,91 @@ private static <T> T getPayloadFromRequest(Request request, Class<T> payloadClas
...
+
+    /**
+     * Get the expected leg by comparing the current time against the start and end time of each leg. If the current time
+     * is between the start and end time of a leg, this is the leg we expect the traveler to be on.
+     */
+    private static Leg getExpectedLeg(ZonedDateTime timeNow, MonitoredTrip monitoredTrip) {
+        if (canUseTripLegs(monitoredTrip)) {
+            for (Leg leg : monitoredTrip.itinerary.legs) {
+                if (isTimeInRange(
+                    leg.getScheduledStartTime(),
```
binh-dam-ibigroup: For transit legs, we will have to use the realtime start/end times if available, as they correspond to actual conditions.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -237,4 +251,91 @@ private static <T> T getPayloadFromRequest(Request request, Class<T> payloadClas
...
+    }
+
+
+    /**
+     * Get the acceptable 'on track' boundary in meters for mode.
+     */
+    private static double getModeBoundary(String mode) {
+        switch (mode.toUpperCase()) {
+            case "WALK":
+                return 5;
```
binh-dam-ibigroup: There is probably a speed assumption that we should mention (and refine later).

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,211 @@
...
+    @AfterEach
+    public void tearDownAfterTest() {
+        if (trackedJourney != null) {
+            trackedJourney.delete();
+            trackedJourney = null;
+        }
+    }
+
+    // This is WIP. These real points and time are nowhere near the expected. This is probably due to the speed of the
+    // subway.
```
binh-dam-ibigroup: We should expect disruption in GPS signal from travel through certain tunnels and I think that is what is happening in the sample.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -237,4 +251,91 @@ private static <T> T getPayloadFromRequest(Request request, Class<T> payloadClas
...
+     * position.
+     */
+    public static TripStatus getTripStatus(TrackedJourney trackedJourney, MonitoredTrip monitoredTrip) {
+        TrackingLocation lastLocation = trackedJourney.locations.get(trackedJourney.locations.size() - 1);
+        Coordinates currentPosition = new Coordinates(lastLocation.lat, lastLocation.lon);
+        ZonedDateTime currentTime = ZonedDateTime.ofInstant(lastLocation.timestamp.toInstant(), DateTimeUtils.getOtpZoneId());
+        var expectedLeg = getExpectedLeg(currentTime, monitoredTrip);
+        if (canUseLeg(expectedLeg)) {
+            List<ManageLegTraversal.Segment> segments = interpolatePoints(expectedLeg);
+            Coordinates expectedPosition = getSegmentPosition(expectedLeg.getScheduledStartTime(), currentTime, segments);
```
binh-dam-ibigroup: Right here is the crux of the logic. Extract these functions that return the expected position:
- `getExpectedPosition(Instant instant, Itinerary itinerary)`
- `getExpectedPosition(Instant instant, Leg leg)`
 
Then you can write tests around what these function should return.

binh-dam-ibigroup: My gut feeling is that you don't need to use a timezone because the underlying start/end time of each leg is not zoned (although `getScheduledStart/EndTime` returns zoned time info).

binh-dam-ibigroup (commented):
I think you are on the right track. Here are my comments after first reading. We should discuss more.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
br648: @leonardehrenfried What order do positions come out of `PolylineUtils.decode`? Do they match the order in the encoded polyline? I'm ordering them because once decoded there doesn't seem to be an order applied. This maybe down to this method or more likely the original order in the encoded polyline.

leonardehrenfried: Ah, you're using my re-packaged polyline decoder lib! :heart_decoration: 

They should be in the same order as they were encoded. If they aren't then there is a bug somewhere.

What is the source of the polyline here? Is it OTP?

br648: @leonardehrenfried The source is indeed OTP.
This is the polyline I'm decoding: [polyline.txt](https://github.com/ibi-group/otp-middleware/files/14167534/polyline.txt)

leonardehrenfried: If I use https://geojson.io to decode the polyline, I get one going from South Atlanta towards the city centre and then a bit further north. Is this what you would expect?

leonardehrenfried: Screenshot from 2024-02-05 16-19-13

It's a bit hard to see but it's the black line starting from the airport.

br648: Yes, that is the route I'm expecting.

leonardehrenfried: And that's also what is being returned from the decode method? 

The coordinates don't follow any kind of order, they just follow the shape which can create loops, turn back on itself or just bad zigzaggy data.

br648: I might have been going off on a tangent here, the order from `Polyline.decode` looks ok. Let me get back to you...

br648: Points before ordering: 
[points_before_ordering.txt](https://github.com/ibi-group/otp-middleware/files/14168835/points_before_ordering.txt)

Points after ordering:
[points_after_ordering.txt](https://github.com/ibi-group/otp-middleware/files/14168841/points_after_ordering.txt)

There are a lot of differences. I need the points to be ordered so I can traverse the leg.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
br648: Yes, have also renamed it.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: image
image

br648: I've plotted the first 100 lat/lon points. The first is the trip, the second the provided traveller positions. The data isn't accurate enough to associate the two.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
br648: I've updated to use start/end times regardless of leg type.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
br648: Updated and working with just the start/end time for now i.e. no zoned.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
leonardehrenfried: I worked on a project before that tried to matched phone gps to route shapes and the data was always very, very messy and hard to match. So, this doesn't surprise me at all.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private static TrackedJourney trackedJourney;
+
+    private static final int SUBWAY_LEG_ID = 1;
+
+    private static final int FINAL_LEG_ID = 2;
+
+    @BeforeAll
+    public static void setUp() throws IOException {
+        assumeTrue(IS_END_TO_END);
```
binh-dam-ibigroup: Maybe this is the file that shouldn't require the whole e2e setup. Mongo access is probably not needed. That makes it easier to run everything locally and quickly.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,306 @@
...
+        TrackedJourney trackedJourney = new TrackedJourney();
+        trackedJourney.tripId = busStopJusticeCenterTrip.id;
+        HashMap<String, Integer> outcome = new HashMap<>();
+        for (String trace : traces) {
+            String[] traceValues = trace.split(",");
+            TrackingLocation trackingLocation = new TrackingLocation();
+            ZonedDateTime zdt = ZonedDateTime.parse(traceValues[0]);
+            trackingLocation.timestamp = new Date(zdt.toInstant().toEpochMilli());
+            trackingLocation.lat = Double.parseDouble(traceValues[1]);
+            trackingLocation.lon = Double.parseDouble(traceValues[2]);
```
binh-dam-ibigroup: This is an important piece of logic that converts converts coordinates into a `TrackingLocation`, extract it into a  constructor for `TrackingLocation` or into a helper method in the `main` code.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,306 @@
...
+                    new TrackingLocation(
+                        segment.coordinates.lat,
+                        segment.coordinates.lon,
+                        new Date((currentTime.toInstant().toEpochMilli()))
+                    )
+                );
+                assertEquals(
+                    ManageTripTracking.getTripStatus(trackedJourney, monitoredTrip).name(),
+                    TripStatus.ON_TRACK.name()
+                );
```
binh-dam-ibigroup: This piece of code should be the subject of a parameterized test for a given leg and leg geometry. Basically, for test params you want to pass a timestamp, a position, and an expected status.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,306 @@
...
+        List<ManageLegTraversal.Segment> segments = createSegmentsForLeg(SUBWAY_LEG_ID);
+        double cumulative = 0;
+        for (ManageLegTraversal.Segment segment : segments) {
+            cumulative += segment.timeInSegment;
+        }
+        assertEquals(monitoredTrip.itinerary.legs.get(SUBWAY_LEG_ID).duration, Math.round(cumulative));
+    }
+
+    @Test
+    void cumulativeSegmentTimeMatchesWalkLegDuration() {
```
binh-dam-ibigroup, line 162: Hmm, I didn't think we would need to use cumulative time, but I'll give it another thought.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,273 @@
...
+     * positions within all legs.
+     *
+     * @param currentCoordinates Traveler's current position.
+     * @param itinerary Trip itinerary containing all trip legs.
+     * @return The shortest distance to any trip position.
+     */
+    public static double getDistanceFromNearestTripPosition(Coordinates currentCoordinates, Itinerary itinerary) {
+        double shortestDistance = Double.MAX_VALUE;
+        if (canUseTripLegs(itinerary)) {
+            for (Leg leg : itinerary.legs) {
```
binh-dam-ibigroup: This is interesting. Maybe we need this method, but just for a leg and for the transition points between legs, instead of the entire itinerary.
My reasoning is that once someone engages in one leg or a transition point between two legs, they are not (should not be) going back to a previous leg.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: No, this is from me initially copying the `TrackedTripControllerTest` test. Have addressed.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: These tests may not be needed. I created them to make sure the cumulative segment times equal the leg duration.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -23,29 +23,23 @@ private ManageLegTraversal() {
...
-     * Define the shortest distance to any position within a trip by checking the traveler's position against all
-     * positions within all legs.
-     *
-     * @param currentCoordinates Traveler's current position.
-     * @param itinerary Trip itinerary containing all trip legs.
-     * @return The shortest distance to any trip position.
+     * Define the segment within a leg that is the closest to the traveler's current position.
      */
-    public static double getDistanceFromNearestTripPosition(Coordinates currentCoordinates, Itinerary itinerary) {
+    public static ManageLegTraversal.Segment getLegSegmentNearestToCurrentPosition(Leg leg, Coordinates currentCoordinates) {
```
binh-dam-ibigroup: Do we need to qualify `Segment`?

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+        TripStatus tripStatus = ManageTripTracking.getTripStatus(trackedJourney, busStopJusticeCenterTrip);
+        assertEquals(expected, tripStatus);
+    }
+
+    private static Stream<Arguments> createTrace() {
+        return Stream.of(
+            Arguments.of("2024-01-26T19:07:05Z", 33.95022, -83.9906, TripStatus.BEHIND_SCHEDULE),
+            Arguments.of("2024-01-26T19:07:08Z", 33.95022, -83.9906, TripStatus.ON_SCHEDULE),
+            Arguments.of("2024-01-26T19:07:50Z", 33.95022, -83.9906, TripStatus.AHEAD_OF_SCHEDULE),
+            Arguments.of("2024-01-26T19:07:43Z", 33.6408211227185, -84.4465224660685, TripStatus.DEVIATED),
```
binh-dam-ibigroup: Maybe insert a comment, especially this case, for where the coordinate is with respect to the leg geometry.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+    /**
+     * Make sure that all the required leg parameters are present.
+     */
+    private static boolean canUseLeg(Leg leg) {
+        return
+            leg != null &&
+                leg.duration > 0 &&
+                leg.legGeometry != null &&
+                leg.legGeometry.points != null &&
+                !leg.legGeometry.points.isEmpty();
```
binh-dam-ibigroup: Shift indent left?

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+     * Calculate the distance between each position and from this the number of seconds spent in each segment. If the
+     * time spent in a segment is less than five seconds, group these segments. This assumes that the leg is being
+     * traversed at a constant speed for simplicity.
+     *
+     * @param orderedPositions Points along a leg.
+     * @param timePerMeter  The average time to cover a meter within a leg.
+     * @return A list of segments, around five seconds in size with an associated lat/lon.
+     */
+    public static List<Segment> getTimeInSegments(SortedSet<Position> orderedPositions, double timePerMeter, String mode) {
+        int minimumSegmentTime = 5;
```
binh-dam-ibigroup: Make this a class constant.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+        List<Segment> segments = new ArrayList<>();
+        List<Position> positions = new ArrayList<>(orderedPositions);
+        Coordinates groupCoordinates = null;
+        double groupSegmentTime = 0;
+        double cumulativeTime = 0;
+        for (int i = 0; i < positions.size() - 1; i++) {
+            Coordinates c1 = new Coordinates(positions.get(i).getLatitude(), positions.get(i).getLongitude());
```
binh-dam-ibigroup: Nice-to-have: a constructor for `Coordinates` that takes a `Position`.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+            Arguments.of("2024-01-26T19:12:25Z", 33.9517224, -83.9929082, TripStatus.NO_STATUS)
+        );
+    }
+
+    @Test
+    void canTrackLegWithoutDeviating() {
+        for (int legId = 0; legId < monitoredTrip.itinerary.legs.size(); legId++) {
```
binh-dam-ibigroup: All instances of `monitoredTrip.itinerary.legs` should be `monitoredTrip.matchingItinerary.legs`. Indeed, real-time tracking is against the real-time itinerary, especially when transit is involved.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+    private static MonitoredTrip monitoredTrip;
+    private static MonitoredTrip busStopJusticeCenterTrip;
+
+    private static final int SUBWAY_LEG_ID = 1;
+
+    private static final int FINAL_LEG_ID = 2;
+
+    @BeforeAll
+    public static void setUp() throws IOException {
+        monitoredTrip = JsonUtils.getPOJOFromJSON(
```
binh-dam-ibigroup: Do we need a JSON file for `monitoredTrip`? Can we construct a monitored trip from, in this instance `busStopJusticeCenterTrip`?

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.opentripplanner.middleware.triptracker.ManageLegTraversal.getSecondsToMilliseconds;
+import static org.opentripplanner.middleware.triptracker.ManageLegTraversal.interpolatePoints;
+
+public class ManageLegTraversalTest {
+
+    private static MonitoredTrip monitoredTrip;
+    private static MonitoredTrip busStopJusticeCenterTrip;
```
binh-dam-ibigroup: This should be an `Itinerary` instead, and it should be renamed to `busStopToJusticeCenterItinerary` (Remove the wrapper from the JSON content too.)

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+            CommonTestUtils.getTestResourceAsString("controllers/api/bus-stop-justice-center-trip.json"),
+            MonitoredTrip.class
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("createTrace")
+    void canTrackTrip(String time, double lat, double lon, TripStatus expected) {
+        TrackedJourney trackedJourney = new TrackedJourney();
+        trackedJourney.tripId = busStopJusticeCenterTrip.id;
```
binh-dam-ibigroup: I think you might have intended to use `monitoredTrip` here and for `getTripStatus`.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
binh-dam-ibigroup: Rename `legId` => `legIndex`.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    void cumulativeSegmentTimeMatchesWalkLegDuration() {
+        List<ManageLegTraversal.Segment> segments = createSegmentsForLeg(FINAL_LEG_ID);
+        double cumulative = 0;
+        for (ManageLegTraversal.Segment segment : segments) {
+            cumulative += segment.timeInSegment;
+        }
+        assertEquals(monitoredTrip.itinerary.legs.get(FINAL_LEG_ID).duration, cumulative);
```
binh-dam-ibigroup: Specify a delta for floating point comparison.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    void cumulativeSegmentTimeMatchesSubwayLegDuration() {
+        List<ManageLegTraversal.Segment> segments = createSegmentsForLeg(SUBWAY_LEG_ID);
+        double cumulative = 0;
+        for (ManageLegTraversal.Segment segment : segments) {
+            cumulative += segment.timeInSegment;
+        }
+        assertEquals(monitoredTrip.itinerary.legs.get(SUBWAY_LEG_ID).duration, Math.round(cumulative));
```
binh-dam-ibigroup: Specify a delta for floating point comparison.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
binh-dam-ibigroup: In a least one trip walking west - southwest, the segments are returned in the inverse order from the steps (i.e. the departing location in the itinerary was related to the last segment). So you will have to set the comparator direction based on the general direction of travel?

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param timePerMeter  The average time to cover a meter within a leg.
+     * @return A list of segments, around five seconds in size with an associated lat/lon.
+     */
+    public static List<Segment> getTimeInSegments(SortedSet<Position> orderedPositions, double timePerMeter, String mode) {
+        int minimumSegmentTime = 5;
+        List<Segment> segments = new ArrayList<>();
+        List<Position> positions = new ArrayList<>(orderedPositions);
+        Coordinates groupCoordinates = null;
+        double groupSegmentTime = 0;
+        double cumulativeTime = 0;
```
binh-dam-ibigroup, line 181: This is assuming that the segments are in the order they will be traveled, but as written, for the case where the segment order is inverse of the travel order, `cumulativeTime` will not mean anything.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+            Comparator.comparing(Position::getLatitude).thenComparing(Position::getLongitude)
+        );
+        sortedPoints.addAll(positions);
+        return sortedPoints;
+    }
+
+    public static class Segment {
+
+        /** The coordinates associated with this segment. */
+        public Coordinates coordinates;
```
binh-dam-ibigroup: Why is there only one set of coordinates for a segment as opposed to the coordinates of the two extremities?
Having the two extremities also makes easier the calculation of the distance to a segment.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Get the expected leg by comparing the current time against the start and end time of each leg. If the current time
+     * is between the start and end time of a leg, this is the leg we expect the traveler to be on.
+     */
+    public static Leg getExpectedLeg(Instant timeNow, Itinerary itinerary) {
+        if (canUseTripLegs(itinerary)) {
+            for (Leg leg : itinerary.legs) {
+                if (leg.startTime != null && leg.endTime != null && (isTimeInRange(
```
binh-dam-ibigroup: Can we return the first leg of the itinerary if the current time is before the start of the itinerary? The context for that is when the first leg is a walk leg and people decide to start travel early.

binh-dam-ibigroup (changes requested):
We are getting there, however two flaws need to be addressed among others - happy to talk more anytime:
- Ordering of segments in a leg: if solely based on coordinates, then cumulative time does not hold a meaning in all cases/travel directions.
- Segments could have two extremities not one, for better distance calculations.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
```diff
@@ -83,4 +83,13 @@ public static double getModeBoundary(String mode) {
...
         }
     }
+
+    /**
+     * Get the distance to the nearest segment point.
+     */
+    private static double getDistanceFromSegment(Coordinates currentPosition, ManageLegTraversal.Segment segment) {
+        double distanceFromStart = getDistance(currentPosition, segment.start);
+        double distanceFromEnd = getDistance(currentPosition, segment.end);
+        return Math.min(distanceFromStart, distanceFromEnd);
```
binh-dam-ibigroup: I had in mind something more like this: https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Line_defined_by_two_points

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
binh-dam-ibigroup: Disregard this comment.

configurations/default/env.yml.tmp
```diff
@@ -75,4 +75,13 @@ SPARKPOST_KEY: your-api-key
 VALIDATE_ENVIRONMENT_CONFIG: true
 
 # The expected frequency to receive live journey location data.
-TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS: 5
\ No newline at end of file
+TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS: 5
```
binh-dam-ibigroup, line 78: Can you update README and env.schema.json too?

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
binh-dam-ibigroup: Can we just say `Segment` instead of `ManageLegTraversal.Segment`?

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -18,66 +18,100 @@
...
-            Arguments.of("2024-01-26T19:12:25Z", 33.9517224, -83.9929082, TripStatus.NO_STATUS)
+            Arguments.of(
+                getDateTimeAsString(startTime, before.cumulativeTime),
+                current.start.lat,
+                current.start.lon,
+                TripStatus.BEHIND_SCHEDULE
+            ),
+            Arguments.of(
+                getDateTimeAsString(startTime, current.cumulativeTime),
+                current.start.lat,
```
binh-dam-ibigroup, line 73: Can you add two cases please:
1. One with a tiny deviation on lat and lon that is still considered on schedule
2. Two with a tiny deviation regarding the time (i.e. `current.cumulativeTime +/- a few seconds) that are still considered "on-schedule".

configurations/default/env.yml.tmp
```diff
@@ -75,4 +75,13 @@ SPARKPOST_KEY: your-api-key
...
+# The minimum segment size in seconds for interpolated points.
+TRIP_TRACKING_MINIMUM_SEGMENT_TIME: 5
+# Trip tracking mode boundaries in meters
+TRIP_TRACKING_WALK_BOUNDARY: 5
+TRIP_TRACKING_BICYCLE_BOUNDARY: 10
+TRIP_TRACKING_BUS_BOUNDARY: 20
+TRIP_TRACKING_SUBWAY_BOUNDARY: 100
+TRIP_TRACKING_TRAM_BOUNDARY: 100
+TRIP_TRACKING_RAIL_BOUNDARY: 200
```
binh-dam-ibigroup, line 87: Maybe we need an on-schedule threshold too.

binh-dam-ibigroup (commented):
Getting there, can you add the test cases with the slight lat/lon deviation and time deviation, and update the schema and README.

configurations/default/env.yml.tmp
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/214/commits/d4cc0bb523319ae7ee4eaa9d70446a8aa7850ee8

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/214/commits/d4cc0bb523319ae7ee4eaa9d70446a8aa7850ee8

configurations/default/env.yml.tmp
br648: I don't follow, can you expand on this please?

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
br648: Dropped ordering... for now.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
br648: Addressed in https://github.com/ibi-group/otp-middleware/pull/214/commits/d4cc0bb523319ae7ee4eaa9d70446a8aa7850ee8

configurations/default/env.yml.tmp
binh-dam-ibigroup: I was thinking we might a time threshold, but I guess because we are tracking the distance to an expected location, the time threshold is probably not needed.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,229 @@
...
+
+    /** The smallest permitted time in seconds for a segment. */
+    public static final int TRIP_TRACKING_MINIMUM_SEGMENT_TIME
+        = getConfigPropertyAsInt("TRIP_TRACKING_MINIMUM_SEGMENT_TIME", 5);
+
+    private ManageLegTraversal() {
+    }
+
+    /**
+     * Define the segment that is the closest to the traveler's current position.
```
binh-dam-ibigroup: Mention the assumption that each segment is a straight line.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -33,60 +39,79 @@ private ManageTripTracking() {
...
-        // Provide response.
-        return new UpdateTrackingResponse(
-            getInstructions(TripStage.UPDATE),
-            getTripStatus(TripStage.UPDATE)
-        );
+                    try {
+                        TripStatus status = getTripStatus(trackedJourney, monitoredTrip.itinerary);
+                        // Provide response.
+                        return new UpdateTrackingResponse(
+                            getInstructions(TripStage.UPDATE),
```
binh-dam-ibigroup: This returns `STAY_ON_BUS` all the time... maybe switch it to `NO_INSTRUCTION` for now?

binh-dam-ibigroup (approved):
I'm going ahead and approve this as it hits the basic hallmarks from the trip tracking endpoint.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    private static boolean canUseTripLegs(Itinerary itinerary) {
+        return
+            itinerary != null &&
+            itinerary.legs != null &&
+            !itinerary.legs.isEmpty();
+    }
+
+    /**
+     * Using the duration of a leg and it's points, produce a list of segments each containing a representative
```
leonardehrenfried:
```suggestion
     * Using the duration of a leg and its points, produce a list of segments each containing a representative
```

pom.xml
```diff
@@ -347,5 +347,23 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- google-polyline-codec. https://gitlab.com/leonard.ehrenfried/google-polyline-codec -->
+        <dependency>
+            <groupId>io.leonard</groupId>
```
leonardehrenfried, line 352: I feel honoured.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,230 @@
...
+    /**
+     * Get the segment which contains the current time.
+     */
+    public static Segment getSegmentFromTime(
+        Instant segmentStartTime,
+        Instant currentTime,
+        List<Segment> segments
+    ) {
+        for (Segment segment : segments) {
+            // Offset the end time by a faction to avoid exact times being attributed to the wrong previous segment.
```
leonardehrenfried:
```suggestion
            // Offset the end time by a fraction to avoid exact times being attributed to the wrong previous segment.
```

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -0,0 +1,230 @@
...
+        double totalDistance = getDistanceTraversedForLeg(positions);
+        return (totalDistance > 0)
+            ? getTimeInSegments(positions, expectedLeg.duration / totalDistance, expectedLeg.mode)
+            : new ArrayList<>();
+    }
+
+    /**
+     * Get the segment which contains the current time.
+     */
+    public static Segment getSegmentFromTime(
```
leonardehrenfried, line 134: Not sure if its customary in this project, but in OTP we add `@Nullable` to methods that can return null.

src/main/java/org/opentripplanner/middleware/triptracker/Segment.java
```diff
@@ -0,0 +1,29 @@
...
+    public Coordinates end;
+
+    /** The time spent in this segment in seconds. */
+    public double timeInSegment;
+
+    /** The leg mode associated with this segment. */
+    public String mode;
+
+    /** The cumulative time since the start of the leg. This includes time in segment. */
+    public double cumulativeTime;
```
leonardehrenfried: If these can be final?

leonardehrenfried (changes requested):
I'm not too involved in the business logic and can't comment too much about this, but I've found a few smaller things.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
br648: I'm not sure if it is, but I'm going to update to include.

leonardehrenfried (approved):

PR #219, opened 2024-02-26
Adapt OTP2 API Calls to GraphQL

JymDyerIBI:
### Checklist

### Description

Adapt OTP2 API calls to GraphQL.


src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -95,6 +97,41 @@ public ItineraryExistenceResult getResultForDayOfWeek(DayOfWeek dayOfWeek) {
...
+                case "date":
+                case "time":
+                case "fromPlace":
+                case "toPlace":
+                    builder.append("\"" + k + "\":\"" + v + "\",");
+                    break;
+
+                case "mode":
+                    builder.append("\"modes\":[");
+                    String[] modes = v.split(",");
```
binh-dam-ibigroup: Hmm you need to use the modes within the string builder here, otherwise no modes will be passed.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -95,6 +97,41 @@ public ItineraryExistenceResult getResultForDayOfWeek(DayOfWeek dayOfWeek) {
...
     }
 
+    /**
+     * Helper function to adapt OTP request parameters to GraphQL's JSON {@code variables} object, limited to
+     * variables specified in the GraphQL plan.  Parameter values are all simply Strings, so we can't use
+     * {@code ObjectMapper} to get the types right, must hardcode them specifically for ItineraryExistence so
+     * that the values that are actually strings have {@code \"} around them.
+     */
+    private static String paramsToVariables(Map<String, String> params) {
+        StringBuilder builder = new StringBuilder("{");
```
binh-dam-ibigroup: Can you use `JsonUtils.toJson` that does the same thing as this method?

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -22,16 +23,15 @@ public class OtpDispatcher {
...
 
-    /**
-     * Location of the OTP plan endpoint (e.g. /routers/default/plan).
-     */
-    public static final String OTP_PLAN_ENDPOINT = getConfigPropertyAsText("OTP_PLAN_ENDPOINT", "/routers/default/plan");
-
     /**
      * Location of the OTP GraphQL endpoint (e.g. /routers/default/index/graphql).
      */
     public static final String OTP_GRAPHQL_ENDPOINT = getConfigPropertyAsText("OTP_GRAPHQL_ENDPOINT", "/routers/default/index/graphql");
```
binh-dam-ibigroup: Use the new paths for OTP2 graphql: `/gtfs/v1`.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -57,6 +57,22 @@ public static OtpDispatcherResponse sendOtpPostRequest(
...
+    /**
+     * Send GraphQL POST request. Would ideally take a JSON Object, but the {@code "query"} object is
+     * followed by a string of JSON-ish GraphQL and the {@code "variables"} object is a proper JSON
+     * object of key/value string pairs so we put these together by hand.
+     * @version OTP version passed along to post request
+     * @variables a string of a JSON object of key/value string pairs
+     */
+    public static OtpDispatcherResponse sendGraphQLPostRequest(OtpVersion version, String variables) {
+        var body = new StringBuilder("{\"query\":\"");
+        body.append(GraphQLUtils.getSchema());
```
binh-dam-ibigroup: Rename `getSchema` to `getPlanQueryTemplate` or something similar. There is a separate actual GraphQL schema file used for parsing requests and mocks in OTP-react-redux.

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+
+public class GraphQLUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(GraphQLUtils.class);
+
+    // Lazily-initialized in getSchema()
+    private static String schema = null;
```
binh-dam-ibigroup: Rename to `planQueryTemplate`

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    static String schemaAsString(String resource) {
+        StringBuilder builder = new StringBuilder();
+        try (var reader = new BufferedReader(new InputStreamReader(
+            GraphQLUtils.class.getClassLoader().getResourceAsStream(resource)
+        ))) {
+            int value;
+            // All this low-level stuff is just to put a \ in front of " in the string.
+            while ((value = reader.read()) != -1) {
+                if ((char)value == '\n') builder.append("\\n");
```
binh-dam-ibigroup: Extract variable for `(char)value`.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -22,16 +23,15 @@ public class OtpDispatcher {
...
-
     /**
      * Location of the OTP GraphQL endpoint (e.g. /routers/default/index/graphql).
      */
     public static final String OTP_GRAPHQL_ENDPOINT = getConfigPropertyAsText("OTP_GRAPHQL_ENDPOINT", "/routers/default/index/graphql");
 
+    /**
+     * Location of the OTP plan endpoint (e.g. /routers/default/plan).
+     */
+    public static final String OTP_PLAN_ENDPOINT = getConfigPropertyAsText("OTP_PLAN_ENDPOINT", "/routers/default/plan");
```
binh-dam-ibigroup: (cosmetic) Revert this change.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -95,6 +97,41 @@ public ItineraryExistenceResult getResultForDayOfWeek(DayOfWeek dayOfWeek) {
...
+                    builder.append("\"modes\":[");
+                    String[] modes = v.split(",");
+                    builder.append("],");
+                    break;
+
+                default:
+                    break;
+            }
+        });
+        return builder.toString();
```
binh-dam-ibigroup: You need to close the block that was open at line 107.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -57,6 +57,22 @@ public static OtpDispatcherResponse sendOtpPostRequest(
...
+     * Send GraphQL POST request. Would ideally take a JSON Object, but the {@code "query"} object is
+     * followed by a string of JSON-ish GraphQL and the {@code "variables"} object is a proper JSON
+     * object of key/value string pairs so we put these together by hand.
+     * @version OTP version passed along to post request
+     * @variables a string of a JSON object of key/value string pairs
+     */
+    public static OtpDispatcherResponse sendGraphQLPostRequest(OtpVersion version, String variables) {
+        var body = new StringBuilder("{\"query\":\"");
+        body.append(GraphQLUtils.getSchema());
+        body.append("\",\\n\"variables\":");
```
binh-dam-ibigroup: The result is `\n"variables": {` ... is this a correct syntax?

image

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
```diff
@@ -57,6 +57,22 @@ public static OtpDispatcherResponse sendOtpPostRequest(
...
 
+    /**
+     * Send GraphQL POST request. Would ideally take a JSON Object, but the {@code "query"} object is
+     * followed by a string of JSON-ish GraphQL and the {@code "variables"} object is a proper JSON
+     * object of key/value string pairs so we put these together by hand.
+     * @version OTP version passed along to post request
+     * @variables a string of a JSON object of key/value string pairs
+     */
+    public static OtpDispatcherResponse sendGraphQLPostRequest(OtpVersion version, String variables) {
+        var body = new StringBuilder("{\"query\":\"");
```
binh-dam-ibigroup: There is a SonarLint refactor suggestion you could use here.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -95,6 +97,41 @@ public ItineraryExistenceResult getResultForDayOfWeek(DayOfWeek dayOfWeek) {
...
+     * {@code ObjectMapper} to get the types right, must hardcode them specifically for ItineraryExistence so
+     * that the values that are actually strings have {@code \"} around them.
+     */
+    private static String paramsToVariables(Map<String, String> params) {
+        StringBuilder builder = new StringBuilder("{");
+        params.forEach((k, v) -> {
+            switch (k) {
+                case "arriveBy":
+                case "numItineraries":
+                    builder.append("\"" + k + "\":" + v + ",");
```
binh-dam-ibigroup: Trailing commas will result in request failure.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
```diff
@@ -186,7 +223,8 @@ public void checkExistence() {
                 setResultForDayOfWeek(result, dayOfWeek);
             }
             // Send off each plan query to OTP.
-            OtpDispatcherResponse response = OtpDispatcher.sendOtpPlanRequest(OtpVersion.OTP1, otpRequest);
+            String variables = ItineraryExistence.paramsToVariables(otpRequest.requestParameters);
+            OtpDispatcherResponse response = OtpDispatcher.sendGraphQLPostRequest(OtpVersion.OTP2, variables);
             TripPlan plan = null;
             try {
                 plan = response.getResponse().plan;
```
binh-dam-ibigroup: for OTP2, it is `response.getResponse().data.plan`. I think we need to create a new class `Otp2Response` to reflect that.

binh-dam-ibigroup (changes requested):
I have not been able to obtain a response from the requests made to the graphql endpoint but was able to identify what changes are needed before we get to that.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: The keys and values we're reading are stored in a `Map<String, String>` so `JsonUtils.toJson`, `ObjectMapper`, and other packages would just put quotes around all values.  The values for things like `"arriveBy"` and `"numItineraries"` are not `String`s, so they need the special case to write them without the quotes.  For example it would create `"arriveBy":"false"` rather than `"arriveBy":false`.

I tried a few different approaches to this but they were all more cumbersome than the `StringBuilder` approach.

binh-dam-ibigroup: Okay, thanks for the explanation, can you add that in the method's JavaDoc?

binh-dam-ibigroup: @JymDyerIBI Would `new Gson().toJson(...)` work for converting these cases to JSON?

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Fixed in recent commit.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: I'm unclear on what to revert. This line and Javadoc are unchanged.
```
f38adca85 (Landon Reed        2020-07-24  25)     /**
b6c3f5ba0 (Rob Beer           2020-09-18  26)      * Location of the OTP plan endpoint (e.g. /routers/default/plan).
f38adca85 (Landon Reed        2020-07-24  27)      */
372fd820e (Landon Reed        2020-07-24  28)     public static final String OTP_PLAN_ENDPOINT = getConfigPropertyAsText("OTP_PLAN_ENDPOINT", "/routers/default/plan");
```

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Good catch. Fixed.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: How odd that this line was deleted. I'll restore it.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: I've edited the JavaDoc to hopefully be clearer.

I did try `Gson` to create the `"variables"` object. This involves creating a dummy class kept in sync with the GraphQL file, and some of what this helper method did was apply a "different names serializer" for things like changing `mode` into `"modes"`. However it got quite complicated for `"modes"` because it involved `Gson` type adapter features that I couldn't get working and the `Gson` documentation says they discourage the use of.

I also first tried a few GraphQL-specific packages that I hoped would produce a dummy class from `"variables"` but it seems that they all fall short with either variables or mutations.

binh-dam-ibigroup: Okay, don't worry about `Gson` now.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Oh, is it because I moved the declaration? I can put it back to where it was for diff purposes.

binh-dam-ibigroup: > Oh, is it because I moved the declaration? I can put it back to where it was for diff purposes.

Yes, please.

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
JymDyerIBI: I'm not sure what this means?

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Okay. The `resources/queries/planQuery.graphql` file is the same as OTP-react-redux's , though I see it needs an update.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: My refactor suggestion was to just make a String, not a StringBuilder in this case.

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
binh-dam-ibigroup: `(char)value` is repeated multiple times, so you can put that in a variable.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: This is restored in last commit.

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
JymDyerIBI: Renamed in last commit.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: Fixed in last commit.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: In last commit.

src/main/java/org/opentripplanner/middleware/models/ItineraryExistence.java
JymDyerIBI: Fixed in last commit.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: In last commit.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Have updated `.graphql` file and did renames as requested in last commit.

src/main/java/org/opentripplanner/middleware/utils/GraphQLUtils.java
JymDyerIBI: In last commit.

src/main/java/org/opentripplanner/middleware/otp/OtpDispatcher.java
JymDyerIBI: Fixed in last commit.

JymDyerIBI:
We want to deploy this code to `sound-transit-qa` but I'm running into an issue with the `transitMode`/`overrideMode` settings. If the query specifies `"LINK_LIGHT_RAIL"` and I pass that along in the GraphQL variables, I get this error from the backend:
```
Variable 'modes' has an invalid value: Invalid input for enum 'Mode'. No value found for name 'LINK_LIGHT_RAIL'
```
In the OTP code, but it looks like it considers a `TraverseMode` to be one of a set of hardcoded enums that doesn't recognize our configured `overrideMode`s at all.  I've got a temporary workaround, but I think we need to handle the `overrideMode`-to-`mode` mappings in otp-react-redux (for which they're configured) so that this code only passes a proper `mode` along.

binh-dam-ibigroup:
> We want to deploy this code to `sound-transit-qa` 

@JymDyerIBI  I have set that instance to track this branch, so as soon as you have a commit with passing tests, it will pick it up.
I'm happy to look into the issues you are encountering regarding the modes, let me know.

binh-dam-ibigroup:
@JymDyerIBI #239 should help by removing mocks in the tests for checking monitored trips and related test failures you saw.

binh-dam-ibigroup:
@JymDyerIBI See also #240 that removes another use of the mock OTP server.

src/main/resources/queries/planQuery.graphql
```diff
@@ -0,0 +1,254 @@
...
+  $wheelchair: Boolean
+) {
+  plan(
+    arriveBy: $arriveBy
+    banned: $banned
+    bikeReluctance: $bikeReluctance
+    carReluctance: $carReluctance
+    date: $date
+    fromPlace: $fromPlace
+    # Currently only supporting EN locale, used for times and text
```
binh-dam-ibigroup, line 25: Remove this comment (produces a syntax error at runtime)

binh-dam-ibigroup: Disregard, sorry^^.

PR #223, opened 2024-03-22
Turn by turn processing

br648:
### Checklist

### Description

This PR aims to provide turn by turn instructions to a traveler. It will inform the traveler of upcoming and immediate instructions based on the information provided in a step.

A list of points for the leg are created from the leg's geo points, start point, end point and step points. Starting with the geo points, the step points are injected based on being between two points. Any step points not added, will be injected based on proximity to a geo point. The final list of points (If b = beginning, p = point on leg, S = step and e = end) would be something like:

b-p-S-p-p-p-p-p-p-S-p-p-S-p-p-p-p-p-S-e

The travellers poisition is first snapped to one of these points to define a position on the leg. Using this as a starting point, the following points are traversed until a "step point" is found. The instruction then provided will be dependent on how far the traveller is from this step. Less than two meters an immediate instruction is given, less than ten meters an upcoming instruction is given, else no instruction is given.


src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -29,38 +29,38 @@ private ManageLegTraversal() {
      * Define the segment that is the closest to the traveler's current position. The assumption being that each segment
      * is a straight line.
      */
-    public static Segment getSegmentFromPosition(Leg leg, Coordinates currentCoordinates) {
+    public static TripSegment getSegmentFromPosition(Leg leg, Coordinates currentCoordinates) {
```
binh-dam-ibigroup: Maybe rename to `LegSegment`?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,217 @@
...
+            if (travelerPosition.tripSegmentFromPosition != null) {
+                return createInstruction(travelerPosition.expectedLeg, travelerPosition.tripSegmentFromPosition);
+            }
+            return NO_INSTRUCTION;
+        }
+        return NO_INSTRUCTION;
+    }
+
+    /**
+     * The trip instruction can only be provided if the trip is on or close to schedule.
```
binh-dam-ibigroup: I think you meant to write "if the traveler is close to the indicated route" (i.e. not deviated).

src/main/java/org/opentripplanner/middleware/utils/Coordinates.java
```diff
@@ -43,4 +55,9 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(lon, lat);
     }
+
+    @Override
+    public String toString() {
+        return lat + "," + lon + ",";
```
binh-dam-ibigroup: What is the trailing comma for?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,217 @@
...
+        TripSegment tripSegment,
+        Coordinates destination
+    ) {
+        double distanceFromLastStepToDestination = getDistance(stepSegment.start, destination);
+        double distanceFromActiveSegmentToDestination = getDistance(tripSegment.start, destination);
+        return distanceFromActiveSegmentToDestination < distanceFromLastStepToDestination;
+    }
+
+    /**
+     * Get instruction based distance and related step instruction.
```
binh-dam-ibigroup: Can you add an example of how that should look, please. I would assume this is a formatted code as opposed to human-readable content?

binh-dam-ibigroup (commented):
Haven't got into the weeds yet...

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
```diff
@@ -65,23 +64,22 @@ public enum TripStatus {
...
+            if (travelerPosition.tripSegmentFromPosition != null &&
+                isWithinModeBoundary(travelerPosition.currentPosition, travelerPosition.tripSegmentFromPosition)) {
+                Instant segmentCenterTime = travelerPosition
+                    .expectedLeg
                     .startTime
                     .toInstant()
-                    .plusSeconds((long) getSegmentTimeInterval(segmentFromPosition));
-                return currentTime.isBefore(segmentCenterTime) ? TripStatus.BEHIND_SCHEDULE : TripStatus.AHEAD_OF_SCHEDULE;
+                    .plusSeconds((long) getSegmentTimeInterval(travelerPosition.tripSegmentFromPosition));
+                return travelerPosition.currentTime.isBefore(segmentCenterTime)
```
binh-dam-ibigroup: Something is odd here, when following an itinerary very slowly starting before the start time, this will report me from being "Behind schedule" to being "ahead of schedule". This should probably be the other way around.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -118,20 +130,104 @@ private static Stream<Arguments> createTrace() {
...
+                getInstruction(UPCOMING_PREFIX, walkSteps.get(4)
+            ))
+        );
+    }
+
+    private static String getInstruction(String prefix, Step step) {
+        return String.format("%s%s", prefix, step.relativeDirection);
+    }
+
+    private static TripSegment createTripSegment(Coordinates point, double distance, double bearing, boolean oppositeDirection) {
```
binh-dam-ibigroup: What unit is `distance` in?

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -51,16 +61,18 @@ void canTrackTrip(String time, double lat, double lon, TripStatus expected) {
         TrackedJourney trackedJourney = new TrackedJourney();
         TrackingLocation trackingLocation = new TrackingLocation(time, lat, lon);
         trackedJourney.locations = List.of(trackingLocation);
-        TripStatus tripStatus = ManageTripTracking.getTripStatus(trackedJourney, busStopToJusticeCenterItinerary);
+        TravelerPosition travelerPosition = new TravelerPosition(trackedJourney, busStopToJusticeCenterItinerary);
+        TripStatus tripStatus = TripStatus.getTripStatus(travelerPosition);
+        String instruction = TripInstruction.getInstruction(tripStatus, travelerPosition);
```
binh-dam-ibigroup: This variable appears unused.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,217 @@
...
+                    // Assuming traveler is approaching the last step.
+                    return approachingStep(activeTripSegment, nearestStep, expectedLeg);
+                }
+            } else {
+                if (isTripSegmentNearerToDestination(nearestStep, activeTripSegment, destination)) {
+                    // Assuming traveler is passed the nearest step.
+                    Step nextStep = expectedLeg.steps.get(nearestStep.stepIndex + 1);
+                    return getInstruction(
+                        getDistance(activeTripSegment.end, new Coordinates(nextStep)),
+                        nextStep.relativeDirection
```
binh-dam-ibigroup: If this is a turn instruction, it should include the name of the path/street.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -118,20 +130,104 @@ private static Stream<Arguments> createTrace() {
...
+    private static String getInstruction(String prefix, Step step) {
+        return String.format("%s%s", prefix, step.relativeDirection);
+    }
+
+    private static TripSegment createTripSegment(Coordinates point, double distance, double bearing, boolean oppositeDirection) {
+        if (oppositeDirection) {
+            bearing = bearing - 180;
+        }
+        Coordinates end = createDestinationPoint(point,distance, bearing);
+        Coordinates start = createDestinationPoint(point,distance + 0.005, bearing);
```
binh-dam-ibigroup: Insert space before `distance`.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,217 @@
...
+    /**
+     * Get instruction based distance and related step instruction.
+     */
+    private static String getInstruction(double distance, String stepInstruction) {
+        if (stepInstruction != null) {
+            if (distance <= IMMEDIATE_INSTRUCTION_RADIUS_DISTANCE) {
+                return IMMEDIATE_PREFIX + stepInstruction;
+            } else if (distance <= UPCOMING_INSTRUCTION_RADIUS_DISTANCE) {
+                return UPCOMING_PREFIX + stepInstruction;
+            }
```
binh-dam-ibigroup: What do you think of setting the instruction to `DEVIATED` (instead of `NO_INSTRUCTION`) if the distance is too far?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,217 @@
...
+    private static boolean isLastStep(int stepIndex, Leg expectedLeg) {
+        return stepIndex == (expectedLeg.steps.size() - 1);
+    }
+
+    /**
+     * Provide instruction for approaching a step.
+     */
+    private static String approachingStep(TripSegment activeTripSegment, StepSegment nearestStep, Leg expectedLeg) {
+        return getInstruction(
+            getDistance(activeTripSegment.end, nearestStep.start),
```
binh-dam-ibigroup: I am not sure about what `activeTripSegment` represents when calling the `/starttracking` endpoint. With a walk-to-transit itinerary, when calling the `/starttracking` endpoint with the start location of the first walk step, the distance computed seems to use the ending point of the `activeTripSegment` which can be away from the starting location, whereas intuitively, the distance should be with respect to the starting location of the first step. Could you clarify?

src/main/java/org/opentripplanner/middleware/triptracker/StepSegment.java
```diff
@@ -0,0 +1,37 @@
...
+    public final int stepIndex;
+
+    public StepSegment(Coordinates start, Coordinates end, int stepIndex) {
+        super(start, end);
+        this.distance = getDistance(start, end);
+        this.stepIndex = stepIndex;
+    }
+
+    public StepSegment(Step startStep, int stepIndex) {
+        super(new Coordinates(startStep), null);
```
binh-dam-ibigroup: Any reason why the ending coordinate is left null?

binh-dam-ibigroup (commented):
More comments. Could you clarify what should be happening for the first step of a walt-to-transit itinerary and the last step of a transit-to-walk itinerary?

binh-dam-ibigroup (changes requested):
The main ask for now, per https://github.com/ibi-group/otp-middleware/pull/223#discussion_r1539451705, is to add the street name to the turn instruction.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: I associate deviated with trip tracking and not trip instruction. If you feel strongly about this, I'll change it. It would also be useful to get feedback from live testing.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: If I have understood correctly, the preceeding method `isTripSegmentNearerToDestination` calculates distance to destination based on start coordinates, the distance calculated here should also use the (activeTripSegment) start coordinates.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
binh-dam-ibigroup: I think we can defer that for now.

src/main/java/org/opentripplanner/middleware/triptracker/StepSegment.java
```diff
@@ -13,14 +13,28 @@ public class StepSegment extends Segment {
...
     public StepSegment(Coordinates start, Coordinates end, int stepIndex) {
         super(start, end);
         this.distance = getDistance(start, end);
         this.stepIndex = stepIndex;
     }
 
-    public StepSegment(Step startStep, int stepIndex) {
-        super(new Coordinates(startStep), null);
+    /**
+     * Used to hold the nearest step to the traveler's position, and it's index. The end coordinate and distance are
```
binh-dam-ibigroup: Typo "**its** index"

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -110,36 +114,41 @@ private static boolean isLastStep(int stepIndex, Leg expectedLeg) {
...
         return distanceFromActiveSegmentToDestination < distanceFromLastStepToDestination;
     }
 
     /**
-     * Get instruction based distance and related step instruction.
+     * Get instruction based on distance and step instructions. e.g.
+     * <p>
+     * "CONTINUE on Langley Drive"
+     * "RIGHT on service road"
+     * "Head WEST on sidewalk approx 6 meters"
```
binh-dam-ibigroup: Add TODO comment for internationalization and refinements to these generated instructions. Alternately, I can walk you through that and perhaps do that in this PR if the changes aren't too many.

src/main/java/org/opentripplanner/middleware/utils/Coordinates.java
```diff
@@ -58,6 +58,9 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return lat + "," + lon + ",";
+        return "Coordinates{" +
```
binh-dam-ibigroup: Use `String.format`

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -157,77 +156,73 @@ private static Stream<Arguments> createTurnByTurnTrace() {
...
+        boolean oppositeDirection
+    ) {
         if (oppositeDirection) {
             bearing = bearing - 180;
         }
-        Coordinates end = createDestinationPoint(point,distance, bearing);
-        Coordinates start = createDestinationPoint(point,distance + 0.005, bearing);
-        return new TripSegment(start, end);
+        Coordinates end = createDestinationPoint(point, distanceInMeters, bearing);
+        Coordinates start = createDestinationPoint(point,distanceInMeters + 5, bearing);
```
binh-dam-ibigroup: Insert space before `distanceInMeters`

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,226 @@
...
+     */
+    public static String getInstruction(double distance, Step step) {
+        if (step != null) {
+            String relativeDirection = (step.relativeDirection.equals("DEPART"))
+                ? "Head " + step.absoluteDirection
+                : step.relativeDirection;
+            return (Math.round(distance) == 0)
```
binh-dam-ibigroup: We need a threshold here. There is no way someone's phone will report the exact spot from the itinerary.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -110,36 +114,41 @@ private static boolean isLastStep(int stepIndex, Leg expectedLeg) {
...
+     * "RIGHT on service road"
+     * "Head WEST on sidewalk approx 6 meters"
      */
-    private static String getInstruction(double distance, String stepInstruction) {
-        if (stepInstruction != null) {
-            if (distance <= IMMEDIATE_INSTRUCTION_RADIUS_DISTANCE) {
-                return IMMEDIATE_PREFIX + stepInstruction;
-            } else if (distance <= UPCOMING_INSTRUCTION_RADIUS_DISTANCE) {
-                return UPCOMING_PREFIX + stepInstruction;
-            }
```
binh-dam-ibigroup: This was previously working. I am not able to get an instruction except for the first move after this last commit.

binh-dam-ibigroup (commented):
Can't get the instructions to appear when nearing a turn, except for the first instruction... Can you advise?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
binh-dam-ibigroup: Actually, just add a TODO comment. We'll need to coordinate with the mobile app folks to see what they will need in the end.

src/main/java/org/opentripplanner/middleware/utils/Coordinates.java
br648: I've made this update, but this is an interesting case. The toString was generated by Intellij!

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: I've reinstated `IMMEDIATE_INSTRUCTION_RADIUS_DISTANCE` with a value of 2 meters.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: Hopefully the latest updates should address this.

src/main/resources/env.schema.json
```diff
@@ -257,6 +257,31 @@
...
     },
+    "TRIP_INSTRUCTION_ARRIVED_PREFIX": {
+      "type": "string",
+      "examples": ["ARRIVED: "],
+      "description": "The prefix to use when at destination."
+    },
+    "TRIP_INSTRUCTION_IMMEDIATE_DISTANCE": {
+      "type": "integer",
+      "examples": ["2"],
+      "description": "The distance under which an immediate instruction is given."
```
binh-dam-ibigroup: In which unit is the distance (2 instances)?

src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
```diff
@@ -18,7 +18,7 @@
 import org.opentripplanner.middleware.testutils.PersistenceTestUtils;
 import org.opentripplanner.middleware.triptracker.ManageTripTracking;
 import org.opentripplanner.middleware.triptracker.TrackingLocation;
-import org.opentripplanner.middleware.triptracker.TripInstruction;
+import org.opentripplanner.middleware.triptracker.TravelerLocator;
```
binh-dam-ibigroup: This import seems unused.

src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
```diff
@@ -42,6 +42,7 @@
 import static org.opentripplanner.middleware.testutils.ApiTestUtils.TEMP_AUTH0_USER_PASSWORD;
 import static org.opentripplanner.middleware.testutils.ApiTestUtils.getMockHeaders;
 import static org.opentripplanner.middleware.testutils.ApiTestUtils.makeRequest;
+import static org.opentripplanner.middleware.triptracker.TripInstruction.NO_INSTRUCTION;
```
binh-dam-ibigroup: This import seems unused.

binh-dam-ibigroup:
~Not sure what you changed. When I call `starttracking` with coordinates very close (to the 6th decimal point) to the itinerary starting location, I get `DEVIATED` which is incorrect, and `NO_INSTRUCTION`, which will not be useful to users:~
```
{
  "frequencySeconds": 5,
  "instruction": "NO_INSTRUCTION",
  "journeyId": "xxxxx",
  "tripStatus": "DEVIATED",
  "message": null
}
```

README.md
```diff
@@ -269,6 +269,11 @@ The special E2E client settings should be defined in `env.yml`:
 | TRIP_TRACKING_SUBWAY_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by subway is considered on track. |
 | TRIP_TRACKING_TRAM_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by tram is considered on track. |
 | TRIP_TRACKING_RAIL_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by rail is considered on track. |
+| TRIP_INSTRUCTION_ARRIVED_PREFIX | string | Optional | ARRIVED:  | The prefix to use when at destination. |
+| TRIP_INSTRUCTION_IMMEDIATE_DISTANCE | integer | Optional | 2 | The distance under which an immediate instruction is given. |
+| TRIP_INSTRUCTION_IMMEDIATE_PREFIX | string | Optional | IMMEDIATE:  | The prefix to use when on an instruction. |
+| TRIP_INSTRUCTION_UPCOMING_DISTANCE | integer | Optional | 10 | The distance under which an upcoming instruction is given. |
+| TRIP_INSTRUCTION_UPCOMING_PREFIX | string | Optional | UPCOMING:  | The prefix to use when nearing an instruction. |
```
binh-dam-ibigroup: Hmm, except for distance thresholds, don't make these parameters configurable. I sense this stuff will have to eventually be in the language bundle instead.

README.md
```diff
@@ -269,6 +269,11 @@ The special E2E client settings should be defined in `env.yml`:
 | TRIP_TRACKING_SUBWAY_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by subway is considered on track. |
 | TRIP_TRACKING_TRAM_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by tram is considered on track. |
 | TRIP_TRACKING_RAIL_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by rail is considered on track. |
+| TRIP_INSTRUCTION_ARRIVED_PREFIX | string | Optional | ARRIVED:  | The prefix to use when at destination. |
+| TRIP_INSTRUCTION_IMMEDIATE_DISTANCE | integer | Optional | 2 | The distance under which an immediate instruction is given. |
```
binh-dam-ibigroup: Specify the unit for the distance thresholds. Are distances a radius or diameter from a turn point?

configurations/default/env.yml.tmp
```diff
@@ -76,12 +76,23 @@ VALIDATE_ENVIRONMENT_CONFIG: true
 
 # The expected frequency to receive live journey location data.
 TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS: 5
-# The minimum segment size in seconds for interpolated points.
+# The minimum trip segment size in seconds for interpolated points.
```
binh-dam-ibigroup: size => duration

configurations/default/env.yml.tmp
```diff
@@ -76,12 +76,23 @@ VALIDATE_ENVIRONMENT_CONFIG: true
...
 TRIP_TRACKING_BUS_BOUNDARY: 20
 TRIP_TRACKING_SUBWAY_BOUNDARY: 100
 TRIP_TRACKING_TRAM_BOUNDARY: 100
-TRIP_TRACKING_RAIL_BOUNDARY: 200
\ No newline at end of file
+TRIP_TRACKING_RAIL_BOUNDARY: 200
+
+# The prefix to use when at destination.
+TRIP_INSTRUCTION_ARRIVED_PREFIX
+# The distance under which an immediate instruction is given.
```
binh-dam-ibigroup: See my comment in README about these parameters.

src/main/java/org/opentripplanner/middleware/triptracker/LegSegment.java
```diff
@@ -0,0 +1,40 @@
...
+        super(start, end);
+        this.timeInSegment = timeInSegment;
+        this.mode = mode;
+        this.cumulativeTime = cumulativeTime;
+        this.distance = getDistance(start, end);
+    }
+
+    @Override
+    public String toString() {
+        return "LegSegment{" +
```
binh-dam-ibigroup: Maybe use String.format

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -0,0 +1,160 @@
...
+    public static String getInstruction(
+        TripStatus tripStatus,
+        TravelerPosition travelerPosition,
+        boolean isStartOfTrip
+    ) {
+        if (hasRequiredTripStatus(tripStatus) &&
+            travelerPosition.expectedLeg != null &&
+            travelerPosition.expectedLeg.mode.equalsIgnoreCase("walk")
+        ) {
+            TripInstruction tripInstruction = alignTravelerToTrip(travelerPosition, isStartOfTrip);
```
binh-dam-ibigroup, line 37: There is one case where a turn instruction is skipped if a leg segment is very short: (see that `prefix` is null but there is a turn instruction in the leg step.
image

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -0,0 +1,160 @@
...
+     * Does the step index reference the last step in the leg.
+     */
+    private static boolean isLastStep(int stepIndex, Leg expectedLeg) {
+        return stepIndex == (expectedLeg.steps.size() - 1);
+    }
+
+    /**
+     * Determine if the step is near to the destination than the traveler.
+     */
+    private static boolean isStepNearerToDestination(
```
binh-dam-ibigroup: This might be tricky. When you have to go around an impassable place to reach your destination, this method will think you are walking in the wrong direction. Would it be easier to just reference the step order within the leg?

binh-dam-ibigroup:
> When I call `starttracking` with coordinates very close (to the 6th decimal point) to the itinerary starting location, I get `DEVIATED` which is incorrect...

Disregard. It can be tricky to find the starting location.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
binh-dam-ibigroup: This is the itinerary where there is a left turn immediately folllowed by the right turn picture above: https://atlrides.com/#/?ui_activeSearch=lljkknis4&ui_activeItinerary=0&fromPlace=805%20Adair%20Avenue%20NE%2C%20Atlanta%2C%20GA%2C%20USA%3A%3A33.77990266501841%2C-84.36094356729174&toPlace=Monroe%20Dr%20NE%20at%20Yorkshire%20Rd%20NE%3A%3A33.79570180624357%2C-84.36787918489891&date=2024-04-05&time=14%3A34&arriveBy=false&mode=WALK%2CTRANSIT

src/main/java/org/opentripplanner/middleware/triptracker/TravelerPosition.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /** Traveler current time. */
+    public Instant currentTime;
+
+    public TravelerPosition(TrackedJourney trackedJourney, Itinerary itinerary) {
+        TrackingLocation lastLocation = trackedJourney.locations.get(trackedJourney.locations.size() - 1);
+        currentTime = lastLocation.timestamp.toInstant();
+        currentPosition = new Coordinates(lastLocation);
+        expectedLeg = getExpectedLeg(lastLocation.timestamp.toInstant(), itinerary);
+        legSegmentFromTime = getSegmentFromTime(lastLocation.timestamp.toInstant(), itinerary);
```
binh-dam-ibigroup: Reuse `currentTime`.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,92 @@
...
-    GET_OFF_BUS,
-    NO_INSTRUCTION;
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsInt;
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
 
-    // More instruction types will be added.
+
+public class TripInstruction {
+
+    /** The distance under which an immediate instruction is given. */
```
binh-dam-ibigroup: See README comment regarding the unit for the distance parameters.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -118,20 +126,125 @@ private static Stream<Arguments> createTrace() {
...
+        Step stepOne = walkSteps.get(0);
+        Coordinates stepOneCoords = new Coordinates(stepOne);
+        Step stepTwo = walkSteps.get(1);
+        Coordinates stepTwoCoords = new Coordinates(stepTwo);
+        Step stepThree = walkSteps.get(2);
+        Coordinates stepThreeCoords = new Coordinates(stepThree);
+        Step stepFour = walkSteps.get(3);
+        Coordinates stepFourCoords = new Coordinates(stepFour);
+        Step stepFive = walkSteps.get(4);
+        Coordinates stepFiveCoords = new Coordinates(stepFive);
```
binh-dam-ibigroup, line 165: Can these things be an array or a list instead? It will be easier to build your arguments stream from that.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -118,20 +126,125 @@ private static Stream<Arguments> createTrace() {
...
     void canAccumulateCorrectStartAndEndCoordinates() {
-        List<Segment> segments = createSegmentsForLeg();
-        for (int i=0; i < segments.size()-1; i++) {
-            Segment segmentOne = segments.get(i);
-            Segment segmentTwo = segments.get(i+1);
-            assertEquals(segmentOne.end.lat, segmentTwo.start.lat);
+        List<LegSegment> legSegments = createSegmentsForLeg();
+        for (int i = 0; i < legSegments.size()-1; i++) {
+            LegSegment legSegmentOne = legSegments.get(i);
+            LegSegment legSegmentTwo = legSegments.get(i+1);
```
binh-dam-ibigroup: Insert spaces between pluses and minuses.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -140,74 +253,99 @@ void canTrackLegWithoutDeviating() {
...
+                cumulativeTravelTime += legSegment.timeInSegment;
                 currentTime = startOfTrip.plus(getSecondsToMilliseconds(cumulativeTravelTime), ChronoUnit.MILLIS);
             }
         }
     }
 
+    @Test
+    void canTrackTripTurnByTurn() {
+        ZonedDateTime startOfTrip = ZonedDateTime.ofInstant(
+            busStopToJusticeCenterItinerary.legs.get(0).startTime.toInstant(),
```
binh-dam-ibigroup: Extract var for `busStopToJusticeCenterItinerary.legs.get(0)` and its first step.

binh-dam-ibigroup (changes requested):
The logic is much better this time, nice work! I think we are now in the minor tweaking phase. I'm going to ask you take the `PREFIX` params out of the configuration and make them constants for now, with the intent to move them to the i18n bundle in subsequent work, and specify in the docs (or in the parameter names) the units in which distances are expressed. Fixing the missed turn case is probably a nice-to-have (unless the second reviewer finds that blocking). See other refactoring suggestions.

README.md
br648: Removed.

src/main/java/org/opentripplanner/middleware/triptracker/LegSegment.java
br648: Done. Interesting: https://stackoverflow.com/questions/35933114/intellij-idea-template-for-tostring-method-using-string-format-instead-of-concat

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: @binh-dam-ibigroup The method name is a bit misleading so I have updated it to `isFirstPositionNearestToTarget`. I have also moved the method into the `GeometryUtils` class. The trip destination is only used if working if the last step, in all other cases the next step in the leg is used i.e. the step order within the leg.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: Sorry, I was meaning to remove this unless unit test.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
br648: I thought this was easier to follow by naming the steps and coords. The arguments don't always use the step with it's coords, e.g. "Approaching the destination." which is a combination of the last step and destination coords. Lemme know what you think, I may have misunderstood.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: This might be down to configuration. The distance to the turn from the traveler is 11.3 meters which is outside of the `UPCOMING` radius.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: @binh-dam-ibigroup is this now working again?

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: To be tested again by Binh.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -48,13 +42,14 @@ public static StartTrackingResponse startTracking(Request request) {
                 Persistence.trackedJourneys.create(trackedJourney);
 
                 try {
-                    TripStatus status = getTripStatus(trackedJourney, monitoredTrip.itinerary);
+                    TravelerPosition travelerPosition = new TravelerPosition(trackedJourney, monitoredTrip.itinerary);
```
binh-dam-ibigroup: This needs to be `monitoredTrip.journeyState.matchingItinerary`.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -83,12 +78,12 @@ public static UpdateTrackingResponse updateTracking(Request request) {
                     );
 
                     try {
-                        TripStatus status = getTripStatus(trackedJourney, monitoredTrip.itinerary);
+                        TravelerPosition travelerPosition = new TravelerPosition(trackedJourney, monitoredTrip.itinerary);
```
binh-dam-ibigroup: This needs to be `monitoredTrip.journeyState.matchingItinerary`.

binh-dam-ibigroup: The reason for that is that the itinerary matches for the recurring day of the monitored trip.

binh-dam-ibigroup:
@br648 Your last changes got instruction to head to the desired street if deviated, however I can't get upcoming turn directions when updating a location to near a turn.

binh-dam-ibigroup (changes requested):
I can't get turn instructions to show anymore. Any reason for that?

br648:
> I can't get turn instructions to show anymore. Any reason for that?

@binh-dam-ibigroup Sorry! I hadn't correctly set the "tripOnTrack" value. See: https://github.com/ibi-group/otp-middleware/pull/223/commits/34fd4fcd984f316c3eddc2dd5d217f220b567767

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
binh-dam-ibigroup: This is still happening, I cannot find a spot that triggers the turn instruction.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -0,0 +1,178 @@
...
+            if (firstStep == null) {
+                return null;
+            }
+            double distance = getDistance(travelerPosition.currentPosition, new Coordinates(firstStep));
+            return (distance <= TRIP_INSTRUCTION_UPCOMING_RADIUS)
+                ? new TripInstruction(distance, firstStep)
+                : null;
+        }
+
+        Coordinates destination = new Coordinates(travelerPosition.expectedLeg.to);
```
binh-dam-ibigroup: Rename to `legDestination`?

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
binh-dam-ibigroup: Okay, I got the turn instruction for this one, that was a tricky one to find!

binh-dam-ibigroup: I think regardless that this will be an issue when there is winding routing such as in the screenshot below. Just using the distance on line 129 in this file will cause the logic to skip turn steps such as the one circled in the screenshot, and that will be an issue later.

image

binh-dam-ibigroup (commented):
More comments regarding J-shaped itineraries.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -73,12 +75,10 @@ private static boolean hasRequiredTripStatus(TripStatus tripStatus) {
      */
     @Nullable
     private static TripInstruction getBackOnTrack(TravelerPosition travelerPosition) {
-        int nearestStepIndex = getNearestStep(travelerPosition.expectedLeg, travelerPosition.currentPosition);
-        Step nearestStep = getStep(travelerPosition.expectedLeg.steps, nearestStepIndex);
-        if (nearestStep == null) {
-            return null;
-        }
-        return new TripInstruction(nearestStep.streetName);
+        Step nearestStep = snapToLeg(travelerPosition);
```
binh-dam-ibigroup: Maybe rename to `snapToStep`.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -99,80 +99,113 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
...
+
+        List<Coordinates> finalPositions = new ArrayList<>();
+        for (int i = 0; i < allPositions.size() - 1; i++) {
+            Coordinates p1 = allPositions.get(i);
+            finalPositions.add(p1);
+            Coordinates p2 = allPositions.get(i + 1);
+            for (Step step : leg.steps) {
+                if (isPointBetween(p1, p2, new Coordinates(step)) && !injectedSteps.contains(step)) {
+                    finalPositions.add(new Coordinates(step));
+                    injectedSteps.add(step);
```
binh-dam-ibigroup, line 177: That seems clever

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -99,80 +99,113 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
...
+                .collect(Collectors.toList());
+            for (Step missedStep : missedSteps) {
+                int pointIndex = getNearestPointIndex(finalPositions, new Coordinates(missedStep));
+                if (pointIndex != -1) {
+                    finalPositions.add(pointIndex, new Coordinates(missedStep));
+                }
+            }
+
+        }
+        return finalPositions;
```
binh-dam-ibigroup, line 237: Some consecutive coordinates are duplicated, but it's probably fine to keep them.

image

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -99,80 +99,113 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
...
+                .stream()
+                .filter(step -> !injectedSteps.contains(step))
+                .collect(Collectors.toList());
+            for (Step missedStep : missedSteps) {
+                int pointIndex = getNearestPointIndex(finalPositions, new Coordinates(missedStep));
+                if (pointIndex != -1) {
+                    finalPositions.add(pointIndex, new Coordinates(missedStep));
+                }
+            }
+
```
binh-dam-ibigroup, line 212: Remove blank line.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
```diff
@@ -65,23 +64,24 @@ public enum TripStatus {
...
+    public static TripStatus getTripStatus(TravelerPosition travelerPosition) {
+        if (travelerPosition.expectedLeg != null) {
+            if (travelerPosition.legSegmentFromTime != null &&
+                isWithinModeBoundary(travelerPosition.currentPosition, travelerPosition.legSegmentFromTime)
+            ) {
                 return TripStatus.ON_SCHEDULE;
             }
-            if (segmentFromPosition != null && isWithinModeBoundary(currentPosition, segmentFromPosition)) {
-                Instant segmentCenterTime = expectedLeg
+            if (travelerPosition.legSegmentFromPosition != null &&
```
binh-dam-ibigroup, line 69: I would be in favor of putting the position first, then time after. If someone is ahead or behind schedule on a multi-leg itinerary, we wouldn't be able to get the correct leg or step they are at using this code.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,118 @@
...
-    GET_OFF_BUS,
-    NO_INSTRUCTION;
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsInt;
 
-    // More instruction types will be added.
+public class TripInstruction {
+
+    /** The radius in meters under which an immediate instruction is given. */
+    public static final int TRIP_INSTRUCTION_IMMEDIATE_RADIUS
+        = getConfigPropertyAsInt("TRIP_INSTRUCTION_IMMEDIATE_RADIUS", 2);
```
binh-dam-ibigroup, line 11: I feel this will have to be increased, but we will tweak that separately.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,118 @@
...
+
+    /** The radius in meters under which an upcoming instruction is given. */
+    public static final int TRIP_INSTRUCTION_UPCOMING_RADIUS
+        = getConfigPropertyAsInt("TRIP_INSTRUCTION_UPCOMING_RADIUS", 10);
+
+    /** The prefix to use when on an instruction. */
```
binh-dam-ibigroup:
```suggestion
    /** The prefix to use when at a street location with an instruction. */
```

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -1,30 +1,118 @@
...
+    public static final String TRIP_INSTRUCTION_IMMEDIATE_PREFIX = "IMMEDIATE: ";
+
+    /** The prefix to use when nearing an instruction. */
```
binh-dam-ibigroup:
```suggestion
    /** The prefix to use when nearing a street location with an instruction. */
```

binh-dam-ibigroup (changes requested):
Nice job handling the J-shaped routing. One small change regarding doing prioritizing obtaining the leg from the location over obtaining the leg using time.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: It shouldn't impact functionality, but I have removed them and made this part a bit cleaner.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
br648: @binh-dam-ibigroup This makes sense and I've made the change. It does mean that `legSegmentFromTime` will be redundant because a lat/lon will always be attributed to a leg segment regardless of how far it is from it. Any thoughts on this before I remove the related code?

binh-dam-ibigroup: I think it will be fine to remove.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
```diff
@@ -66,33 +66,34 @@ public enum TripStatus {
...
-                Instant segmentCenterTime = travelerPosition
+                Instant segmentStartTime = travelerPosition
                     .expectedLeg
                     .startTime
                     .toInstant()
-                    .plusSeconds((long) getSegmentTimeInterval(travelerPosition.legSegmentFromPosition));
-                return travelerPosition.currentTime.isBefore(segmentCenterTime)
-                    ? TripStatus.AHEAD_OF_SCHEDULE
-                    : TripStatus.BEHIND_SCHEDULE;
+                    .plusSeconds((long) getSegmentStartTime(travelerPosition.legSegmentFromPosition));
```
binh-dam-ibigroup, line 76: What was the reason to change from `getSegmentTimeInterval` to `getSegmentStartTime`? This change seems to report me "ahead of schedule" even if I am behind schedule.

binh-dam-ibigroup: Oh I see, never mind. If I am way behind schedule, the trip monitoring might have switched to the next available trip the next day already...

br648: Just to add this. The method now has to cover on schedule so it needs to consider the start and end segment times.

configurations/test/env.yml
```diff
@@ -51,4 +51,17 @@ MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
...
+TRIP_TRACKING_UPDATE_FREQUENCY_SECONDS: 5
+TRIP_TRACKING_MINIMUM_SEGMENT_TIME: 5
+TRIP_TRACKING_WALK_BOUNDARY: 5
+TRIP_TRACKING_BICYCLE_BOUNDARY: 10
+TRIP_TRACKING_BUS_BOUNDARY: 20
+TRIP_TRACKING_SUBWAY_BOUNDARY: 100
+TRIP_TRACKING_TRAM_BOUNDARY: 100
+TRIP_TRACKING_RAIL_BOUNDARY: 200
+TRIP_INSTRUCTION_IMMEDIATE_PREFIX_DISTANCE: 2
+TRIP_INSTRUCTION_IMMEDIATE_PREFIX: "IMMEDIATE: "
```
binh-dam-ibigroup: Delete this setting (it is not configurable)

configurations/test/env.yml
```diff
@@ -51,4 +51,17 @@ MAXIMUM_PERMITTED_MONITORED_TRIPS: 5
...
+TRIP_INSTRUCTION_UPCOMING_PREFIX_DISTANCE: 10
+TRIP_INSTRUCTION_UPCOMING_PREFIX: "UPCOMING: "
```
binh-dam-ibigroup: Delete this setting (it is not configurable)

binh-dam-ibigroup (approved):
Nice work, and the latest changes are so much better!
A couple of notes for subsequent work:
- To determine whether the user is on a walk leg that follows a transit/non-walk leg, the user's position should be used over the current time.
- If the user is within the "upcoming" radius of a turn, the turn instruction should probably be given even if the user is "Deviated" from the itinerary.

JymDyerIBI (approved):
This looks good, makes sense. I didn't replicate the `TravelerLocator` edge condition.

PR #227, opened 2024-04-23
Turn-by-turn tweaks

br648:
### Checklist

### Description

Jira: https://ibisolutions.atlassian.net/browse/OTP-1179

The following updates have been applied in this PR:

1. Configurable deviated thresholds for each mode
2. Current leg is obtained from position and not time
3. A turn instruction will be issued if deviated, but within a turn location upcoming radius.


src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -86,6 +86,22 @@ private static Stream<Arguments> createTrace() {
         LegSegment before = legSegments.get(8);
         LegSegment current = legSegments.get(10);
         LegSegment after = legSegments.get(12);
+        Coordinates deviatedCoordinates = createPoint(
+            new Coordinates(current.start.lat, current.start.lon),
```
binh-dam-ibigroup: Say `current.start` instead of `new Coordinates(current.start.lat, current.start.lon)`? (multiple instances)

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -86,6 +86,22 @@ private static Stream<Arguments> createTrace() {
...
+            90,
+            calculateBearing(
+                new Coordinates(current.start.lat, current.start.lon),
+                new Coordinates(after.start.lat, after.start.lon)
```
binh-dam-ibigroup: Say `after.start` instead of `new Coordinates(after.start.lat, after.start.lon)`? (multiple instances)

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -32,17 +32,15 @@ public static String getInstruction(
         TravelerPosition travelerPosition,
         boolean isStartOfTrip
     ) {
-        if (hasRequiredTripStatus(tripStatus) && hasRequiredWalkLeg(travelerPosition)
-        ) {
+        if (hasRequiredTripStatus(tripStatus) && hasRequiredWalkLeg(travelerPosition)) {
```
binh-dam-ibigroup: Combine condition `hasRequiredWalkLeg(travelerPosition)` for these two `if` blocks.

README.md
```diff
@@ -263,12 +263,16 @@ The special E2E client settings should be defined in `env.yml`:
...
-| TRIP_TRACKING_WALK_BOUNDARY | integer | Optional | 5 | The threshold in meters below which walking is considered on track. |
-| TRIP_TRACKING_BICYCLE_BOUNDARY | integer | Optional | 10 | The threshold in meters below which cycling is considered on track. |
-| TRIP_TRACKING_BUS_BOUNDARY | integer | Optional | 20 | The threshold in meters below which travelling by bus is considered on track. |
-| TRIP_TRACKING_SUBWAY_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by subway is considered on track. |
-| TRIP_TRACKING_TRAM_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by tram is considered on track. |
-| TRIP_TRACKING_RAIL_BOUNDARY | integer | Optional | 100 | The threshold in meters below which travelling by rail is considered on track. |
+| TRIP_TRACKING_WALK_ON_TRACK_RADIUS | integer | Optional | 5 | The threshold in meters below which walking is considered on track. |
+| TRIP_TRACKING_WALK_DEVIATED_RADIUS | integer | Optional | 10 | The threshold in meters below which walking is considered deviated. |
```
binh-dam-ibigroup: I don't think we need the `_DEVIATED_RADIUS` parameters. As written, this is making live-tracking instructions harder to use because trip status will say "no status" if someone misses a turn. Same probably applies to the other modes.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
```diff
@@ -96,33 +113,47 @@ public static double getSegmentStartTime(LegSegment legSegmentFromPosition) {
         return legSegmentFromPosition.cumulativeTime - legSegmentFromPosition.timeInSegment;
     }
 
+    /**
+     * Checks if the traveler's position is with an acceptable deviated distance of the mode type.
+     */
+    private static boolean isWithinModeDeviatedRadius(TravelerPosition travelerPosition) {
```
binh-dam-ibigroup: See the other comment. I don't think we need that.

binh-dam-ibigroup (changes requested):
I think the deviated radius params would suppress instructions, and that would make the tracking feature harder to use.

src/main/java/org/opentripplanner/middleware/triptracker/TripStatus.java
br648: Removed.

README.md
br648: Removed.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -58,29 +58,19 @@ private static boolean canUseLeg(Leg leg) {
...
-                    isTimeInRange(
-                        leg.startTime.toInstant(),
-                        // Offset the end time by a faction to avoid exact times being attributed to the wrong leg.
-                        leg.endTime.toInstant().minus(1, ChronoUnit.MILLIS),
-                        timeNow
-                    )) {
-                    return leg;
+                List<Coordinates> allPositions = getAllLegPositions(leg);
+                for (int j = 0; j < allPositions.size() - 1; j++) {
+                    if (isPointBetween(allPositions.get(j), allPositions.get(j + 1), position)) {
```
binh-dam-ibigroup: I think this condition is too restrictive. For the initial step, this will never work. For all intermediate steps, this takes over the on-track radii defined as parameters. I thought you had code to snap to a nearest leg segment, could we use that instead?

br648: I agree. The snap to nearest leg segment is only possible once the leg has been defined so that can't be used. However, I have implemented a similar approach for `getExpectedLeg` which should result in a lot more matches.

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -58,33 +58,26 @@ private static boolean canUseLeg(Leg leg) {
...
-                if (leg.startTime != null &&
-                    leg.endTime != null &&
-                    isTimeInRange(
-                        leg.startTime.toInstant(),
-                        // Offset the end time by a faction to avoid exact times being attributed to the wrong leg.
-                        leg.endTime.toInstant().minus(1, ChronoUnit.MILLIS),
-                        timeNow
-                    )) {
-                    return leg;
+                for (Coordinates positionOnLeg : getAllLegPositions(leg)) {
```
binh-dam-ibigroup: Sorry for going back and forth, but the snippet you had previously is still useful when someone engages the next leg after taking a turn as instructed. In that case we want to not show the turn instruction.
So how about using `isPointBetween` first, then fallback on `getDistance`?

```java
                List<Coordinates> allPositions = getAllLegPositions(leg);
                for (int j = 0; j < allPositions.size() - 1; j++) {
                    if (isPointBetween(allPositions.get(j), allPositions.get(j + 1), position)) {
                        return leg;
```

br648: No worries. Updated in: https://github.com/ibi-group/otp-middleware/pull/227/commits/e0e990c067a831c06664bd93009fc4582463d00c

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -148,202 +148,150 @@ private static Stream<Arguments> createTrace() {
...
-            ),
-            Arguments.of(
-                new TurnTrace(
-                    edmundParkDriveToRockSpringsItinerary,
-                    createPoint(sixthGeoPoint, 52, calculateBearing(sixthGeoPoint, new Coordinates(rockSpringsStepTwo))),
-                    new TripInstruction(0, rockSpringsStepTwo).build(),
-                    false,
-                    "Immediate second step."
+                    originCoords,
+                    new TripInstruction(10, stepOne).build(),
```
binh-dam-ibigroup: Could you please reorder these tests, so that stuff about the first step is first, and so on.

br648: Done also did a bit of tidying: https://github.com/ibi-group/otp-middleware/pull/227/commits/1ef9e08ae26c204c2533f7de5cd25847a53487eb

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
```diff
@@ -148,202 +148,150 @@ private static Stream<Arguments> createTrace() {
...
                 )
             ),
             Arguments.of(
                 new TurnTrace(
-                    adairAvenueToMonroeDriveItinerary,
                     createPoint(pointBeforeTurn, 16, calculateBearing(pointBeforeTurn, virginiaAvenuePoint)),
                     new TripInstruction(2, virginiaAvenue).build(),
                     false,
-                    "Turn left on to Virginia Avenue."
+                    "Turn left on to Virginia Avenue (Test to make sure turn is not missed)."
```
binh-dam-ibigroup: We are missing a case where the position is immediately after a turn, but still within the "upcoming" radius. In that case, the instruction should be on-track. On the field, the upcoming turn instruction still appears after having taking the turn as instructed.

binh-dam-ibigroup (changes requested):
Add one test case for a location immediately following a turn, but still within the "upcoming" radius.

src/test/java/org/opentripplanner/middleware/controllers/api/ManageLegTraversalTest.java
binh-dam-ibigroup: Add this case:
```java
            Arguments.of(
                new TurnTrace(
                    createPoint(pointAfterTurn, 0, calculateBearing(pointAfterTurn, virginiaAvenuePoint)),
                    NO_INSTRUCTION,
                    false,
                    "After turn left on to Virginia Avenue should not produce turn instruction."
                )
            ),
```
where 
```java
        Coordinates pointAfterTurn = new Coordinates(33.78165, -84.36484);
```

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -98,11 +100,8 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
...
-        Coordinates legDestination = new Coordinates(travelerPosition.expectedLeg.to);
-        double distanceToDestination = getDistance(travelerPosition.currentPosition, legDestination);
-        if (distanceToDestination <= TRIP_INSTRUCTION_UPCOMING_RADIUS) {
-            // Assuming traveler is approaching the destination.
-            return new TripInstruction(distanceToDestination, travelerPosition.expectedLeg.to.name);
+        if (isApproachingDestination(travelerPosition)) {
+            return new TripInstruction(getDistanceToDestination(travelerPosition), travelerPosition.expectedLeg.to.name);
         }
 
         Step nextStep = snapToStep(travelerPosition);
```
binh-dam-ibigroup, line 107: At line 108: Add a check that the current position is actually not in a segment of the next step.

br648: @binh-dam-ibigroup Changes in https://github.com/ibi-group/otp-middleware/pull/227/commits/1ef640b13dbe36dad3f071e96864874753e19713. A change in location of mere centimeters can make a difference between getting an instruction or not! I'm not sure how well that is going to perform in the field.

binh-dam-ibigroup (approved):
Much better with the new changes!

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
```diff
@@ -58,28 +58,46 @@ private static boolean canUseLeg(Leg leg) {
...
+    private static Leg getNearestLeg(Coordinates position, Itinerary itinerary) {
+        double shortestDistance = Double.MAX_VALUE;
+        Leg nearestLeg = null;
+        for (int i = 0; i < itinerary.legs.size(); i++) {
+            Leg leg = itinerary.legs.get(i);
+            for (Coordinates positionOnLeg : getAllLegPositions(leg)) {
+                double distance = getDistance(positionOnLeg, position);
+                if (distance < shortestDistance) {
+                    nearestLeg = leg;
+                    shortestDistance = distance;
```
miles-grant-ibigroup, line 88: I forsee problems doing this at transit centers. If a bus goes down a ~200m path to get to the transit center and the next leg goes back the other direction how will we know which to match. Not sure I know how to fix this, but should we maybe add a comment that this might be an issue?

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -115,6 +114,37 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
         return null;
     }
 
+    /**
+     * Check that the current position is not passed the "next step". This is to prevent an instruction being provided
+     * for a step which is behind the traveler, but is within radius.
+     */
+    private static boolean isPositionPassedStep(TravelerPosition travelerPosition, Step nextStep) {
```
miles-grant-ibigroup: We can probably add the compass code here to resolve this? Also possible typo `past`?

miles-grant-ibigroup (approved):
A few potential issues. Can't test but code looks clean

src/main/java/org/opentripplanner/middleware/triptracker/ManageLegTraversal.java
binh-dam-ibigroup: This PR focuses on walking, but you are raising a good point for when we work on transit legs.

br648: There is an attempt to get the leg that contains the current position first (`getContainingLeg`) and then to fallback on this method (`getNearestLeg`) if unsuccessful. But, (having written all that) you are right that the wrong leg could be selected. I'll add a comment for future PRs to consider and potentially address.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: Addressed typo. Thanks.

PR #201, opened 2023-12-15
Localized trip times in notifications

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing

### Description

This PR fixes the initial reminder to use the itinerary start time (and not `trip.tripTime`) 
and updates trip notification contents with localized times (e.g. "5:44 PM" vs. "17:44").
There are no other changes to trip notification contents.

PR #233, opened 2024-05-14
Notify Bus Operator

br648:
### Checklist

### Description

Jira: https://ibisolutions.atlassian.net/browse/OTP-1187

If requirements are met, notify a bus operator of an upcoming passenger with mobility needs.

BLOCKED by this PR:

https://github.com/ibi-group/otp-middleware/pull/219 (Changes will have to be made to work with OTP2 e.g. new itinerary structure)


src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+public class BusOpNotificationMessage {
+
+    public BusOpNotificationMessage() {
+        // Required for JSON deserialization.
+    }
+
+    private static final DateTimeFormatter BUS_OPERATOR_NOTIFIER_API_DATE_FORMAT = DateTimeFormatter
+        .ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
+        .withZone(ZoneId.systemDefault());
```
binh-dam-ibigroup: Use `getOtpZoneId` (2 instances)

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+        codes.put("Blind", 7);
+        codes.put("Device-LowVision", 8);
+        codes.put("Mscooter-LowVision", 9);
+        codes.put("WChairE-LowVision", 10);
+        codes.put("WChairM-LowVision", 11);
+        codes.put("Some-LowVision", 12);
+        codes.put("Device-Blind", 13);
+        codes.put("Mscooter-Blind", 14);
+        codes.put("WchairE-Blind", 15);
+        codes.put("WchairM-Blind", 16);
```
binh-dam-ibigroup: Check spelling compared to the [`MobilityProfile` enum](https://github.com/ibi-group/OpenTripPlanner/blob/6df682fa34fe401bcd9e4bbab85c799d9d3a2a96/src/ext/java/org/opentripplanner/ext/mobilityprofile/MobilityProfile.java).

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -52,7 +53,9 @@ private static TrackingResponse doUpdateTracking(Request request, TripTrackingDa
 
             TravelerPosition travelerPosition = new TravelerPosition(
                 trackedJourney,
-                tripData.trip.journeyState.matchingItinerary
+                tripData.trip.journeyState.matchingItinerary,
+                Auth0Connection.getUserFromRequest(request).otpUser
```
binh-dam-ibigroup: Would it be better to get the user from the trip's user id? You'd avoid decoding the Auth0 token as that was done elsewhere already.

configurations/default/env.yml.tmp
```diff
@@ -89,4 +89,8 @@ TRIP_TRACKING_RAIL_ON_TRACK_RADIUS: 200
 # The radius in meters under which an immediate instruction is given.
 TRIP_INSTRUCTION_IMMEDIATE_RADIUS: 2
 # The radius in meters under which an upcoming instruction is given.
-TRIP_INSTRUCTION_UPCOMING_RADIUS: 10
\ No newline at end of file
+TRIP_INSTRUCTION_UPCOMING_RADIUS: 10
+
+BUS_OPERATOR_NOTIFIER_API_URL: https://bus.notifier.example.com
+BUS_OPERATOR_NOTIFIER_API_SUBSCRIPTION_KEY: your-subscription-key
+BUS_OPERATOR_NOTIFIER_QUALIFYING_ROUTES: agency_id:route_id
```
binh-dam-ibigroup: Nit: The URL for the notifier API is probably agency specific, and eventually multiple agencies could support the feature, so could be placed in a YML or JSON file.

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Class containing the expected notify parameters. These will be converted to JSON and make up the body content of
+ * the request.
+ * <p>
+ * 'To' fields omitted as they are not needed for requests for single transit legs.
+ */
+public class BusOpNotificationMessage {
```
binh-dam-ibigroup: I like the separation between the message and the code that performs the request.

binh-dam-ibigroup: This class is specific to a given agency, so the class name should say so, e.g. `UsRideGwinnettBusOpNotificationMessage`?

Also, in #235, I created a package `org.opentripplanner.middleware.triptracker.interactions`, maybe we can put this class in that package.

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+    public Integer msg_type;
+    public List<Integer> mobility_codes;
+    public boolean trusted_companion;
+
+    public BusOpNotificationMessage(Instant timestamp, TravelerPosition travelerPosition) {
+        this.timestamp = BUS_OPERATOR_NOTIFIER_API_DATE_FORMAT.format(timestamp);
+        this.agency_id = removeAgencyPrefix(travelerPosition.nextLeg.agency.id);
+        this.from_route_id = removeAgencyPrefix(travelerPosition.nextLeg.route.id);
+        this.from_trip_id = travelerPosition.nextLeg.trip.id;
+        this.from_stop_id = travelerPosition.nextLeg.from.stop.id;
```
binh-dam-ibigroup: Extract variable for `...nextLeg`.

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Get the second element from value by removing the OTP agency prefix.
+     * E.g. GwinnettCountyTransit:GCT will return just GCT.
+     */
+    private String removeAgencyPrefix(String value) {
+        return (value != null) ? value.split(":")[1] : null;
+    }
+
+    /**
+     * Get the mobility code that matches the mobility mode. Although the API can accept multiple codes, OTP middleware
```
binh-dam-ibigroup: Multiple codes probably refer to being able to notify about multiple users, each of different profiles, at the same stop.

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -132,9 +135,17 @@ public static EndTrackingResponse forciblyEndTracking(Request request) {
...
-     * Complete a journey by defining the ending type, time and condition.
+     * Complete a journey by defining the ending type, time and condition. Also cancel possible upcoming bus
+     * notification.
      */
-    private static EndTrackingResponse completeJourney(TrackedJourney trackedJourney, boolean isForciblyEnded) {
+    private static EndTrackingResponse completeJourney(Request request, TripTrackingData tripData, boolean isForciblyEnded) {
+        TravelerPosition travelerPosition = new TravelerPosition(
+            tripData.journey,
+            tripData.trip.journeyState.matchingItinerary,
+            Auth0Connection.getUserFromRequest(request).otpUser
```
binh-dam-ibigroup: Same thing here, it would be probably better to get the user from the trip id.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.opentripplanner.middleware.triptracker.TripStatus.getSegmentStartTime;
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
+
+/**
+ * If conditions are correct notify a bus operator of a traveler joining the service at a given stop.
+ */
+public class NotifyBusOperator {
```
binh-dam-ibigroup: This is an agency-specific class, so it should be named that way, maybe `UsRideGwinnett`

src/test/java/org/opentripplanner/middleware/triptracker/NotifyBusOperatorTest.java
```diff
@@ -0,0 +1,166 @@
...
+            Arguments.of(true, TripStatus.ON_SCHEDULE, null, "Traveler is on schedule, notification can be sent."),
+            Arguments.of(false, TripStatus.BEHIND_SCHEDULE, null, "Traveler is behind schedule, notification can not be sent."),
+            Arguments.of(true, TripStatus.AHEAD_OF_SCHEDULE, travelerPosition1, "Traveler is ahead of schedule, but within the notify window."),
+            Arguments.of(false, TripStatus.AHEAD_OF_SCHEDULE, travelerPosition2, "Too far ahead of schedule to notify bus operator.")
+        );
+    }
+
+    private static OtpUser createOtpUser() {
+        MobilityProfile mobilityProfile = new MobilityProfile();
+        mobilityProfile.mobilityMode = "WchairE";
```
binh-dam-ibigroup: Check spelling of the mode.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public NotifyBusOperator() {}
+
+    public static boolean IS_TEST = false;
+
+    private static final Logger LOG = LoggerFactory.getLogger(NotifyBusOperator.class);
+
+    private static final String BUS_OPERATOR_NOTIFIER_API_URL
```
binh-dam-ibigroup: The config variables should probably include the agency, e.g. `US_RIDEGWINNETT_BUS_OPERATOR_NOTIFIER_URL` (multiple instances)

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+        = getConfigPropertyAsText("BUS_OPERATOR_NOTIFIER_API_URL", "not-provided");
+
+    private static final String BUS_OPERATOR_NOTIFIER_API_SUBSCRIPTION_KEY
```
binh-dam-ibigroup: Rename to just `...KEY` instead of `...SUBSCRIPTION_KEY`

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+     * <p>
+     * GwinnettCountyTransit:360,GwinnettCountyTransit:40,GwinnettCountyTransit:25
+     */
+    private static List<String> getBusOperatorNotifierQualifyingRoutes() {
+        String busOperatorNotifierQualifyingRoutes
+            = getConfigPropertyAsText("BUS_OPERATOR_NOTIFIER_QUALIFYING_ROUTES");
+        if (busOperatorNotifierQualifyingRoutes != null) {
+            return Arrays.asList(busOperatorNotifierQualifyingRoutes.split(","));
+        }
+        return new ArrayList<>();
```
binh-dam-ibigroup: If the parameter is not provided, should the default be all routes of the particular agency?

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Stage notification to bus operator by making sure all required conditions are met.
+     */
+    public static void sendNotification(TripStatus tripStatus, TravelerPosition travelerPosition) {
+        try {
+            if (
+                isBusLeg(travelerPosition.nextLeg) &&
+                isWithinOperationalNotifyWindow(tripStatus, travelerPosition) &&
+                hasNotPreviouslyNotifiedBusDriverForRoute(travelerPosition.trackedJourney, travelerPosition.nextLeg.route.id) &&
```
binh-dam-ibigroup: Extract var for route id.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+    public static String createPostBody(TravelerPosition travelerPosition) {
+        return JsonUtils.toJson(new BusOpNotificationMessage(travelerPosition.currentTime, travelerPosition));
+    }
+
+    /**
+     * Make sure the leg in question is a bus transit leg.
+     */
+    public static boolean isBusLeg(Leg leg) {
+        return leg != null && leg.mode.equalsIgnoreCase("BUS") && leg.transitLeg;
+    }
```
binh-dam-ibigroup: Move this to `ItineraryUtils` or a new utils class.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+    public static boolean supportsBusOperatorNotification(String gtfsId) {
+        return QUALIFYING_BUS_NOTIFIER_ROUTES.contains(gtfsId);
+    }
+
+    /**
+     * Has the bus driver already been notified for this journey. The driver must only be notified once.
+     */
+    public static boolean hasNotPreviouslyNotifiedBusDriverForRoute(TrackedJourney trackedJourney, String routeId) {
+        for (String notifiedRouteId : trackedJourney.busNotificationMessages.keySet()) {
+            if (notifiedRouteId.equalsIgnoreCase(routeId)) {
```
binh-dam-ibigroup: I think there should be a check on the type of notification (actual notification vs. cancelation).

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+     * Get how far ahead in minutes the traveler is from the expected schedule.
+     */
+    public static long getMinutesAheadOfSchedule(TravelerPosition travelerPosition) {
+        Instant segmentStartTime = travelerPosition
+            .expectedLeg
+            .startTime
+            .toInstant()
+            .plusSeconds((long) getSegmentStartTime(travelerPosition.legSegmentFromPosition));
+        return Duration.between(segmentStartTime, travelerPosition.currentTime).toMinutes();
+    }
```
binh-dam-ibigroup: This has been previously calculated to obtain `tripStatus`. Could that be stored there?

src/main/java/org/opentripplanner/middleware/triptracker/TravelerPosition.java
```diff
@@ -24,18 +27,34 @@ public class TravelerPosition {
...
+        if (otpUser != null && otpUser.mobilityProfile != null) {
+            mobilityMode = otpUser.mobilityProfile.mobilityMode;
+        }
     }
 
     /** Used for unit testing. */
     public TravelerPosition(Leg expectedLeg, Coordinates currentPosition) {
-        this.expectedLeg = expectedLeg;
         this.currentPosition = currentPosition;
+        this.expectedLeg = expectedLeg;
```
binh-dam-ibigroup: Revert this change?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -37,17 +45,26 @@ public class TripInstruction {
...
-    private boolean tripOnTrack;
+    /** Provided if the next leg for the traveler will be a bus transit leg. */
+    public Leg busLeg;
+
+    /** The time provided by the traveler */
+    public Instant currentTime;
 
-    /** If the traveler is within the upcoming radius an instruction will be provided. */
-    public boolean hasInstruction;
+    /** The type of instruction to be provided to te traveler. */
```
binh-dam-ibigroup: Typo "*the* traveler"

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+        this.trusted_companion = false;
+    }
+
+    /**
+     * Get the second element from value by removing the OTP agency prefix.
+     * E.g. GwinnettCountyTransit:GCT will return just GCT.
+     */
+    private String removeAgencyPrefix(String value) {
+        return (value != null) ? value.split(":")[1] : null;
+    }
```
binh-dam-ibigroup: Move to `ItineraryUtils` or a new utils class.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -121,4 +151,35 @@ private String buildOnTrackInstruction() {
         }
         return NO_INSTRUCTION;
     }
+
+    /**
+     * Build wait for bus instruction.
+     */
+    private String buildWaitForBusInstruction() {
+        String routeId = busLeg.route.id.split(":")[1];
```
binh-dam-ibigroup: Use `removeAgencyPrefix` (see comment for that function).

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -121,4 +151,35 @@ private String buildOnTrackInstruction() {
...
+                .between(currentTime.atZone(DateTimeUtils.getOtpZoneId()), busLeg.getScheduledStartTime())
+                .toMinutes();
+            return String.format("Wait%s for bus %s", getReadableMinutes(waitInMinutes), routeId);
+        } else {
+            return String.format(
+                "Wait for bus %s scheduled to arrive at %s",
+                routeId,
+                BUS_OPERATOR_NOTIFIER_API_TIME_FORMAT.format(busLeg.getScheduledStartTime())
+            );
+        }
```
binh-dam-ibigroup: Can we combine these two instructions into one such as:
"Wait n minutes for your bus, route %s, scheduled at %s [if available: ', on time/now xxx minutes early/late']".

The route number is not the `routeId`, instead it should be the route short name (although they are the same for some agencies such as RideGwinnett).

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -121,4 +151,35 @@ private String buildOnTrackInstruction() {
         }
         return NO_INSTRUCTION;
     }
+
+    /**
+     * Build wait for bus instruction.
+     */
+    private String buildWaitForBusInstruction() {
+        String routeId = busLeg.route.id.split(":")[1];
+        if (busLeg.departureDelay > 0) {
```
binh-dam-ibigroup: Handle cases where the bus is early. Use the same on-time threshold as https://github.com/ibi-group/otp-middleware/blob/7d1f8160cc545294df94816aa6048d3ca47074fc/src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java#L55.

binh-dam-ibigroup: Also, check for real-time data availability (in the `realTime` field, I think).

src/test/java/org/opentripplanner/middleware/models/MobilityProfileTest.java
```diff
@@ -36,6 +36,7 @@ public void testModes(Set<String> devices, String mode) {
         var prof = new MobilityProfile();
         prof.mobilityDevices = devices;
         prof.updateMobilityMode();
+        System.out.println(prof.mobilityMode);
```
binh-dam-ibigroup: Eventually, remove the calls to `System.out.println`.

src/main/java/org/opentripplanner/middleware/otp/response/Leg.java
```diff
@@ -27,7 +27,7 @@ public class Leg implements Cloneable {
     public Double distance;
     public Boolean pathway;
     public String mode;
-    public String route;
+    public Route route;
```
binh-dam-ibigroup: I see. This one is a breaking change from #219, and when that change takes effect, we'll have to run some code to migrate this field.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -100,8 +108,14 @@ public static TripInstruction alignTravelerToTrip(TravelerPosition travelerPosit
                 : null;
         }
 
-        if (isApproachingDestination(travelerPosition)) {
-            return new TripInstruction(getDistanceToDestination(travelerPosition), travelerPosition.expectedLeg.to.name);
+        if (isApproachingEndOfLeg(travelerPosition)) {
+            if (NotifyBusOperator.isBusLeg(travelerPosition.nextLeg)) {
+                // The preceding leg is a bus leg.
```
binh-dam-ibigroup: Did you mean "The upcoming leg"?

src/main/java/org/opentripplanner/middleware/otp/response/Leg.java
binh-dam-ibigroup: Meanwhile, I think we can just use the existing `routeId` field and avoid the breaking change.

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
```diff
@@ -0,0 +1,104 @@
...
+    public String from_arrival_time;
+    public Integer msg_type;
+    public List<Integer> mobility_codes;
+    public boolean trusted_companion;
+
+    public BusOpNotificationMessage(Instant timestamp, TravelerPosition travelerPosition) {
+        this.timestamp = BUS_OPERATOR_NOTIFIER_API_DATE_FORMAT.format(timestamp);
+        this.agency_id = removeAgencyPrefix(travelerPosition.nextLeg.agency.id);
+        this.from_route_id = removeAgencyPrefix(travelerPosition.nextLeg.route.id);
+        this.from_trip_id = travelerPosition.nextLeg.trip.id;
```
binh-dam-ibigroup: Use existing `tripId` field for now until #219 is merged (remove the prefix).

src/main/java/org/opentripplanner/middleware/triptracker/BusOpNotificationMessage.java
binh-dam-ibigroup: Use existing field `from > stopId` for now until #219 is merged (remove prefix).

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -121,4 +151,35 @@ private String buildOnTrackInstruction() {
...
+            long waitInMinutes = Duration
+                .between(currentTime.atZone(DateTimeUtils.getOtpZoneId()), busLeg.getScheduledStartTime())
+                .toMinutes();
+            return String.format("Wait%s for bus %s", getReadableMinutes(waitInMinutes), routeId);
+        } else {
+            return String.format(
+                "Wait for bus %s scheduled to arrive at %s",
+                routeId,
+                BUS_OPERATOR_NOTIFIER_API_TIME_FORMAT.format(busLeg.getScheduledStartTime())
```
binh-dam-ibigroup: For the instruction, use the existing time formatting code from #201 that supports i18n and doesn't show seconds.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+    public static void sendNotification(TripStatus tripStatus, TravelerPosition travelerPosition) {
+        try {
+            if (
+                isBusLeg(travelerPosition.nextLeg) &&
+                isWithinOperationalNotifyWindow(tripStatus, travelerPosition) &&
+                hasNotPreviouslyNotifiedBusDriverForRoute(travelerPosition.trackedJourney, travelerPosition.nextLeg.route.id) &&
+                supportsBusOperatorNotification(travelerPosition.nextLeg.route.id)
+            ) {
+                var body = createPostBody(travelerPosition);
+                var httpStatus = doPost(body);
```
binh-dam-ibigroup: We'll have to add a "call pending" state if the call does not complete right away, so that updates that hit while this call is pending don't initiate another one.

binh-dam-ibigroup (changes requested):
I like that it works with the right things in place. For now let's keep using the OTP1 fields until #219 is merged, and see the various changed I suggested. Note that there is a pending fix with #234 that will correctly set `isStartOfTrip` in various places.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+        var httpResponse = HttpUtils.httpRequestRawResponse(
+            URI.create(BUS_OPERATOR_NOTIFIER_API_URL),
+            1000,
+            HttpMethod.POST,
+            BUS_OPERATOR_NOTIFIER_API_HEADERS,
+            body
+        );
+        return httpResponse.status;
+    }
```
binh-dam-ibigroup: We do already have the common method `HttpUtils.httpRequestRawResponse`, so would it make sense to create a wrapper interface around it and create (i) a generalized class that performs the call and (ii) a generalized class for making mock calls?

configurations/default/env.yml.tmp
br648: @binh-dam-ibigroup are you saying that I should move these two params to a file now? I will at the very least change these params to include `UsRideGwinnett`

binh-dam-ibigroup: The keys and secrets should remain in the `env.yml` file, but there could be some mapping between agencies and the corresponding Java class such as in https://github.com/ibi-group/otp-middleware/pull/235/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5.

br648: OK, I have mirrored what you have done.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
br648: @binh-dam-ibigroup I'll make the change because it would be less to manage. But if none are defined it will increase the number of failed API calls.

binh-dam-ibigroup: > But if none are defined it will increase the number of failed API calls.

What do you mean?

br648: So, if none are defined and only 2 out of 10 routes can be notified, 8 calls to the API are pointless (failed maybe a bit strong, but the response might not be status OK).

binh-dam-ibigroup: I'm good with that.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
br648: @binh-dam-ibigroup I'm not sure I understand. Are you staying that if a notification has previously been cancelled allow a second notification to be sent?

binh-dam-ibigroup: @br648 Are both notifications and cancelations recorded in `trackedJourney.busNotificationMessages`?

br648: @binh-dam-ibigroup No, just the notification. The only difference would be the `msg_type` (1 = Notify, 0 = Cancel). When cancelling, the saved notification is retrived, the `msg_type` is changed to 0 and the request is sent. The saved notification is not updated.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: I have updated, but it will need a second review to confirm I've understood correctly.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
binh-dam-ibigroup: So there is no record of whether a cancel request was sent?

br648: That is correct. It is an easy fix to update the saved message with the updated cancelled `msg_type`. Shall I implement this?

Also, if I do implement and the notification has been previously cancelled, should I allow the notification to be sent again?

binh-dam-ibigroup: @br648 Yes go ahead and update the notification, so that the cancel notification is sent once only. I can bring this up with the bus software folks to see if they support repeat requests.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
br648: @binh-dam-ibigroup Could you expand a bit on this please? How would the generalized class that performs the call be swapped out for making mock calls in unit tests?

binh-dam-ibigroup: @br648 That was a thought. I attempted to do it in #235, but [in this file](https://github.com/ibi-group/otp-middleware/blob/7cd15b5e226e8ddd45b6ae00ca367131f71b79a8/src/main/java/org/opentripplanner/middleware/triptracker/interactions/UsGdotGwinnettTrafficSignalNotifier.java#L71), ended up doing an if block for testing conditions.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -37,40 +47,65 @@ public class TripInstruction {
     /** Name of final destination or street. */
     public String locationName;
 
-    /** Instruction is for a trip that is on track. */
-    private boolean tripOnTrack;
+    /** Provided if the next leg for the traveler will be a bus transit leg. */
+    public Leg busLeg;
```
binh-dam-ibigroup, line 51: Note for another PR: Eventually, we could do an abstract instructions class and have situation-specific classes (while walking, waiting, riding...) derive from it.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
+import static org.opentripplanner.middleware.utils.ItineraryUtils.getRouteIdFromLeg;
+import static org.opentripplanner.middleware.utils.ItineraryUtils.isBusLeg;
+
+/**
+ * If conditions are correct notify a bus operator of a traveler joining the service at a given stop.
```
binh-dam-ibigroup:
```suggestion
 * If conditions are correct, notify a bus operator of a traveler waiting to board a given stop.
```

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Has a previous notification already been cancelled.
+     */
+    public static boolean hasNotCancelledNotificationForRoute(TrackedJourney trackedJourney, String routeId) throws JsonProcessingException {
+        String messageBody = trackedJourney.busNotificationMessages.get(routeId);
+        if (messageBody == null) {
+            // It should not be possible to get here because a notification must exist before it can be cancelled.
+            return false;
```
binh-dam-ibigroup: Should this block throw an exception instead?

src/test/java/org/opentripplanner/middleware/models/MobilityProfileTest.java
```diff
@@ -55,6 +55,7 @@ public void testModesVision(MobilityProfile.VisionLimitation limitation, Set<Str
         prof.mobilityDevices = devices;
         prof.visionLimitation = limitation;
         prof.updateMobilityMode();
+        System.out.println(prof.mobilityMode);
```
binh-dam-ibigroup: Is this console print needed?

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
```diff
@@ -121,4 +159,39 @@ private String buildOnTrackInstruction() {
...
+        long absoluteMinutes = Math.abs(delayInMinutes);
+        long waitInMinutes = Duration
+            .between(currentTime.atZone(DateTimeUtils.getOtpZoneId()), busLeg.getScheduledStartTime())
+            .toMinutes();
+        String delayInfo = (delayInMinutes > 0) ? "late" : "early";
+        String arrivalInfo = (absoluteMinutes <= 1)
+            ? ", on time"
+            : String.format("now %s %s", getReadableMinutes(delayInMinutes), delayInfo);
+        return String.format(
+            "Wait %s for your bus, route %s, scheduled at %s %s",
```
binh-dam-ibigroup: Remove space between the last two `%s`. Right now, if you are late, it prints
```
"Wait  for your bus, route 40, scheduled at 4:14 PM , on time"
```

binh-dam-ibigroup: Nit: Transfer the space after "Wait" to before the number of minutes, or make it conditional to minutes being shown.

binh-dam-ibigroup (approved):
Changes look sound and seem to work! I am going to approve it with a dependency on #234 and a few tweaks as indicated.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: Agree. It's a little messy as is.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
br648: Yes, have updated.

src/test/java/org/opentripplanner/middleware/models/MobilityProfileTest.java
br648: No, removed.

src/main/java/org/opentripplanner/middleware/triptracker/TripInstruction.java
br648: I have addressed the incorrect spaces.

br648:
> Changes look sound and seem to work! I am going to approve it with a dependency on #234 and a few tweaks as indicated.

@binh-dam-ibigroup have you been able to test this fully i.e. using the actual API and getting a 200 response in a resonable time frame? I never had the key in order to do this.

src/main/java/org/opentripplanner/middleware/triptracker/NotifyBusOperator.java
br648: If I get time I'll go back to looking at mocking static methods. It looks like it is possible and would remove testing code from the actual code (which I don't really like).

josh-willis-arcadis (approved):
lots of changes. everything looks clean and readable!

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
```diff
@@ -0,0 +1,210 @@
...
+                isBusLeg(travelerPosition.nextLeg) &&
+                isWithinOperationalNotifyWindow(tripStatus, travelerPosition) &&
+                hasNotSentNotificationForRoute(travelerPosition.trackedJourney, routeId) &&
+                supportsBusOperatorNotification(routeId)
+            ) {
+                // Immediately set the notification state to pending, so that subsequent calls don't initiate another
+                // request before this one completes.
+                travelerPosition.trackedJourney.updateNotificationMessage(routeId, "pending");
+                var body = createPostBody(travelerPosition);
+                var httpStatus = doPost(body);
```
binh-dam-ibigroup, line 79: I was able to perform the call to the API, and it returned 200.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * Stage notification to bus operator by making sure all required conditions are met.
+     */
+    public void sendNotification(TripStatus tripStatus, TravelerPosition travelerPosition) {
+        var routeId = getRouteIdFromLeg(travelerPosition.nextLeg);
+        try {
+            if (
+                isBusLeg(travelerPosition.nextLeg) &&
+                isWithinOperationalNotifyWindow(tripStatus, travelerPosition) &&
```
binh-dam-ibigroup: I have to give this one a thought. This condition seems more restrictive geographically than `isApproachingEndOfLeg(travelerPosition)` that must be met before executing this code. But then, to compute whether someone is on-time or ahead of schedule, we'd have to use the real-time bus arrival time vs. the current time. What are your thoughts on that, @br648?

br648: I'm not 100% sure and won't implement a change until we are in agreement, but I think this method should be updated to consider the bus leg start time (assuming this is altered accordingly by `realTime=true`?). It currently calculates how far the traveler is ahead on the walk leg. I think this then needs to be compared to the bus leg start time to get a truer picture. Thoughts?

binh-dam-ibigroup: That sounds like `isApproachingEndOfLeg(travelerPosition)` should become `isApproachingDestinationOrStartOfNextLeg(travelerPosition)`?

br648: @binh-dam-ibigroup In the interest of time, if you would like to provide psuedo code/logic I could implement something for review as part of our meeting tomorrow?

Also, is my assumption above about real-time correct?

src/main/java/org/opentripplanner/middleware/triptracker/ManageTripTracking.java
```diff
@@ -52,7 +55,8 @@ private static TrackingResponse doUpdateTracking(Request request, TripTrackingDa
```
binh-dam-ibigroup, line 55: Bad merge at line 40: should be `return doUpdateTracking(request, tripData, create);`

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettNotifyBusOperator.java
br648: @binh-dam-ibigroup Refactor in: https://github.com/ibi-group/otp-middleware/pull/233/commits/56e6c1553f1c9d0af609ad45a2890aff2d341aaf

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettBusOpNotificationMessage.java
```diff
@@ -0,0 +1,106 @@
...
+        codes.put("WChairM-Blind", 16);
+        codes.put("Some-Blind", 17);
+        return codes;
+    }
+
+    public String timestamp;
+    public String agency_id;
+    public String from_route_id;
+    public String from_trip_id;
+    public String from_stop_id;
```
binh-dam-ibigroup, line 72: Could you add the `to_stop_id` field too?

br648: Fixed in: https://github.com/ibi-group/otp-middleware/pull/233/commits/d9bb37ea180aeff2823e366d55717829e0ef53ce

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettBusOpNotificationMessage.java
```diff
@@ -0,0 +1,108 @@
...
+    public String to_stop_id;
+    public String from_arrival_time;
+    public Integer msg_type;
+    public List<Integer> mobility_codes;
+    public boolean trusted_companion;
+
+    public UsRideGwinnettBusOpNotificationMessage(Instant timestamp, TravelerPosition travelerPosition) {
+        var nextLeg = travelerPosition.nextLeg;
+        this.timestamp = BUS_OPERATOR_NOTIFIER_API_DATE_FORMAT.format(timestamp);
```
binh-dam-ibigroup: This is incorrect. This will format a local time correctly but show that in UTC. You need to convert the local time to UTC. Here is the sample for the call I made at around 10:43 am EDT:
```
"{"timestamp":"2024-06-06T10:43:45.000Z"
```

binh-dam-ibigroup (changes requested):
Could you fix the timestamps please?

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettBusOpNotificationMessage.java
br648: I think you mean convert UTC to local time? Change in: https://github.com/ibi-group/otp-middleware/pull/233/commits/bbfa26f154d08da4f36f887232a3418e70da467d

binh-dam-ibigroup: I'd have to try out, but the timestamp should be the UTC equivalent of the time the location or request was emitted.

br648: I've based the update on the approach taken to get the `Leg.getScheduledStartTime()` (https://github.com/ibi-group/otp-middleware/blob/f3a0cfbdcae8d81b31cfe7239e6031b6c7f3588a/src/main/java/org/opentripplanner/middleware/otp/response/Leg.java#L76). This is used in `TripInstruction` and `UsRideGwinnettBusOpNotificationMessage` for the `from_arrival_time`.

binh-dam-ibigroup: @br648  In the below example, with bus arrival time at 4:34 pm EDT (local time), even after your latest changes, the timestamp should have been 2024-06-05T20:20:24.000Z:

```
"{\"timestamp\":\"2024-06-05T16:20:24.000Z\",\"agency_id\":\"GCT\",\"from_route_id\":\"40\",\"from_trip_id\":\"t60E-b12F-sl4\",\"from_stop_id\":\"539\",\"to_stop_id\":\"351\",\"from_arrival_time\":\"16:34:56\",\"msg_type\":1,\"mobility_codes\":[],\"trusted_companion\":false}"
```

binh-dam-ibigroup (commented):
See https://github.com/ibi-group/otp-middleware/pull/233#discussion_r1631631418.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/busnotifiers/UsRideGwinnettBusOpNotificationMessage.java
br648: @binh-dam-ibigroup My bad, I didn't realize the spec requires the timestamp in UTC. I think this update should now work: https://github.com/ibi-group/otp-middleware/pull/233/commits/10a32a718ea66488902d3808799ca2ff40ab440b

binh-dam-ibigroup (approved):
Thanks for the changes, the notification timestamp looks good.

PR #235, opened 2024-05-20
Trigger traffic signals [OTP-1125]

binh-dam-ibigroup:
### Checklist

### Description

This PR adds configurations for triggering actions when someone approaches a particular segment defined by coordinates.
The trigger included in this PR is for the API that controls select traffic signals in the Atlanta region. Note that such requests can take a few seconds to complete, so I added a mechanism to discard a call to the signal API if another one has not returned yet.


src/main/java/org/opentripplanner/middleware/triptracker/interactions/UsGdotGwinnettTrafficSignalNotifier.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public String getUrl(String signalId, String crossingId, boolean extended) {
+        return host + String.format(path, signalId, crossingId) + (extended ? "?extended=true" : "");
+    }
+
+    public Map<String, String> getHeaders() {
+        return Map.of("X-API-KEY", key);
+    }
+
+    public String getBody() {
```
br648: So no body is required? Seems like this method could be removed?

src/main/java/org/opentripplanner/middleware/triptracker/interactions/UsGdotGwinnettTrafficSignalNotifier.java
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    String pathAndQuery = getUrl(signalId, crossingId, extended);
+                    var httpResponse = HttpUtils.httpRequestRawResponse(
+                        URI.create(pathAndQuery),
+                        30,
+                        HttpMethod.POST,
+                        getHeaders(),
+                        getBody()
+                    );
+                    if (httpResponse.status == 200) {
```
br648: Could replace with HttpStatus.OK_200.

br648 (approved):
Couple of minors, but looks good to me. We could probably combined the trip actions and bus notifier actions in a later PR.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/UsGdotGwinnettTrafficSignalNotifier.java
binh-dam-ibigroup: Updated in https://github.com/ibi-group/otp-middleware/pull/235/commits/176b5a9acdf5fe2ce09546b71e3673b77470860a.

src/main/java/org/opentripplanner/middleware/triptracker/interactions/UsGdotGwinnettTrafficSignalNotifier.java
binh-dam-ibigroup: Yeah I'll remove for now, see https://github.com/ibi-group/otp-middleware/pull/235/commits/176b5a9acdf5fe2ce09546b71e3673b77470860a

amy-corson-ibigroup (approved):
Sorry for how long this took me but I think it looks good!

PR #266, opened 2024-11-06
Get dependent's mobility profile

br648:
### Checklist

### Description

New API end point to allow a trusted companion to retrieve a dependent's mobility profile if the right conditions are met.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
@@ -26,8 +26,7 @@ import java.util.regex.Pattern;
 
 import static io.github.manusant.ss.descriptor.MethodDescriptor.path;
 import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.ACCEPT_KEY;
-import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.EMAIL_PARAM;
-import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.GET_DEPENDENT_MOBILITY_PROFILE_PATH;
+import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.DEPENDENT_USER_ID;
 import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.USER_LOCALE;
 import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.ensureRelatedUserIntegrity;
 import static org.opentripplanner.middleware.tripmonitor.TrustedCompanion.manageAcceptDependentEmail;
@@ -103,13 +102,12 @@ public class OtpUserController extends AbstractUserController<OtpUser> {
                     .withResponseType(OtpUser.class),
                 TrustedCompanion::acceptDependent
             )
-            .get(path(ROOT_ROUTE + String.format(GET_DEPENDENT_MOBILITY_PROFILE_PATH, ID_PARAM))
+            .get(path(ROOT_ROUTE + "/getdependentmobilityprofile")
                     .withDescription("Retrieve a dependent's mobility profile.")
                     .withResponses(SwaggerUtils.createStandardResponses(MobilityProfile.class))
-                    .withPathParam().withName(ID_PARAM).withRequired(true).withDescription("The id of the related user.").and()
-                    .withPathParam().withName(EMAIL_PARAM).withRequired(true).withDescription("The dependent's email address.").and()
+                    .withPathParam().withName(DEPENDENT_USER_ID).withRequired(true).withDescription("A dependent's user id.").and()
                     .withResponseType(MobilityProfile.class),
-                this::getDependentMobilityProfile, JsonUtils::toJson
+                TrustedCompanion::getDependentMobilityProfile, JsonUtils::toJson
             )
             .get(path(ROOT_ROUTE + String.format(VERIFY_ROUTE_TEMPLATE, ID_PARAM, VERIFY_PATH, PHONE_PARAM))
                     .withDescription("Request an SMS verification to be sent to an OtpUser's phone number.")
@@ -137,14 +135,6 @@ public class OtpUserController extends AbstractUserController<OtpUser> {
             NotificationUtils.updatePushDevices(otpUser);
         }
         return otpUser;
-    }
-
-    /**
-     * Extract the related user from the request and return a dependent's mobility profile if the right conditions
-     * are met.
-     */
-    private MobilityProfile getDependentMobilityProfile(Request request, Response response) {
-        return TrustedCompanion.getDependentMobilityProfile(request, response, getEntityForId(request, response));
     }
 
     /**
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
@@ -7,7 +7,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.auth.Auth0Connection;
 import org.opentripplanner.middleware.auth.RequestingUser;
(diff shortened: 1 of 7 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -100,6 +103,14 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
                     .withResponseType(OtpUser.class),
                 TrustedCompanion::acceptDependent
             )
+            .get(path(ROOT_ROUTE + String.format(GET_DEPENDENT_MOBILITY_PROFILE_PATH, ID_PARAM))
+                    .withDescription("Retrieve a dependent's mobility profile.")
+                    .withResponses(SwaggerUtils.createStandardResponses(MobilityProfile.class))
+                    .withPathParam().withName(ID_PARAM).withRequired(true).withDescription("The id of the related user.").and()
```
binh-dam-ibigroup: I think the caller only has access to emails and does not have access to the user Mongo IDs... should they?

binh-dam-ibigroup: If you prefer using the id over the email as parameter for this endpoint, I think we could add a `userId` field to `RelatedUser` and populate that on confirmation/deletion from the companion user.

binh-dam-ibigroup: Disregard, we are using a list of dependents which is stored as a list of ids.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
```diff
@@ -100,6 +103,14 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
...
+                    .withPathParam().withName(EMAIL_PARAM).withRequired(true).withDescription("The dependent's email address.").and()
```
binh-dam-ibigroup: Okay, after discussion, we don't need the email parameter here.

src/main/java/org/opentripplanner/middleware/tripmonitor/TrustedCompanion.java
```diff
@@ -42,10 +45,11 @@ private TrustedCompanion() {
         ConfigUtils.getConfigPropertyAsText("TRUSTED_COMPANION_CONFIRMATION_PAGE_URL");
     public static final String ACCEPT_KEY = "acceptKey";
     public static final String USER_LOCALE = "userLocale";
-    public static final String EMAIL_FIELD_NAME = "email";
+    public static final String EMAIL_PARAM = "email";
 
     /** Note: This path is excluded from security checks, see {@link OtpMiddlewareMain#initializeHttpEndpoints()}. */
     public static final String ACCEPT_DEPENDENT_PATH = "api/secure/user/acceptdependent";
+    public static final String GET_DEPENDENT_MOBILITY_PROFILE_PATH = "/:%s/getdependentmobilityprofile";
```
binh-dam-ibigroup: I don't think we need to pass the user id here, it is being obtained from the token (I was by mistake leaving the placeholder as is, and the endpoint call proceeded normally).

binh-dam-ibigroup (changes requested):
Good start, please remove the email as a required search parameter, and I don't think we need to pass the userId as a path parameter.

src/main/java/org/opentripplanner/middleware/controllers/api/OtpUserController.java
br648: @binh-dam-ibigroup I think we do. If a related user has many dependents, how do we know which mobility profile to return without the email address?

binh-dam-ibigroup: Yeah, but the email of a dependent is unknown to the companion user without an additional lookup, is it?

br648: My bad, getting confused between email and id. Updated here: https://github.com/ibi-group/otp-middleware/pull/266/commits/3f5f7349fb1354b33c321ec510585e7fe9db65d5

src/main/java/org/opentripplanner/middleware/tripmonitor/TrustedCompanion.java
br648: Updated: https://github.com/ibi-group/otp-middleware/pull/266/commits/3f5f7349fb1354b33c321ec510585e7fe9db65d5

binh-dam-ibigroup (approved):
Much better, thanks!

src/main/java/org/opentripplanner/middleware/tripmonitor/TrustedCompanion.java
```diff
@@ -235,4 +243,38 @@ public static void removeDependent(OtpUser dependent, RelatedUser relatedUser) {
...
+
+        if (isNotEmpty(relatedUser) && isNotEmpty(dependentUser)) {
+            if (!relatedUser.dependents.contains(dependentUser.id)) {
+                logMessageAndHalt(
+                    request,
+                    HttpStatus.FORBIDDEN_403,
+                    String.format("Related user is not a trusted companion for dependent with user id: %s!", dependentUserId)
+                );
+            } else {
+                return dependentUser.mobilityProfile;
```
binh-dam-ibigroup: @br648: @josh-willis-arcadis points out that there is a need to display that user name/nickname/email in the selector, so can we include that in the return? It might make sense to create a separate data structure that has:
- id
- mobility profile string (e.g. "WChairE")
- email
- name or nickname

binh-dam-ibigroup (changes requested):
Sorry, we need more info for the UI to show useful dependent info (see comment).

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
```diff
@@ -90,7 +90,7 @@ public enum Notification {
     /** Companions and observers of this user. */
     public List<RelatedUser> relatedUsers = new ArrayList<>();
 
-    /** Users that are dependent on this user. */
+    /** A list of users (their ids only) that are dependent on this user. */
     public List<String> dependents = new ArrayList<>();
```
binh-dam-ibigroup, line 100: From @josh-willis-arcadis: "Another suggestion is to query a list of userIDs. that way if there are multiple dependents the frontend won't have to make separate calls"

src/main/java/org/opentripplanner/middleware/tripmonitor/TrustedCompanion.java
br648: Updated: https://github.com/ibi-group/otp-middleware/pull/266/commits/deb32264726a15f90fe4a569c41c43f1113d74c5

src/main/java/org/opentripplanner/middleware/models/OtpUser.java
br648: Updated: https://github.com/ibi-group/otp-middleware/pull/266/commits/deb32264726a15f90fe4a569c41c43f1113d74c5

JymDyerIBI (approved):
👍  Okay, this looks good.

src/main/java/org/opentripplanner/middleware/models/MobilityProfileLite.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Objects;
+
+import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
+
+public class MobilityProfileLite {
+    public String userId;
+    public String mobilityMode;
+    public String email;
+    public String name;
```
binh-dam-ibigroup, line 11: We will have to figure out how to handle the name, or if providing a nickname will be needed.

binh-dam-ibigroup (approved):
Looks good and works good!

binh-dam-ibigroup:
Coordinate with @josh-willis-arcadis before merging.

josh-willis-arcadis:
Everything looks good on my end. TY for the changes!

PR #268, opened 2024-11-13
OTP-1458 Provide traveler with continue instruction

br648:
### Checklist

### Description

Provide a "continue on street" reassurance instruction when the traveler is on track, but no immediate instruction is available.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
+++ b/src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
@@ -30,7 +30,6 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static org.opentripplanner.middleware.triptracker.instruction.TripInstruction.NO_INSTRUCTION;
 import static org.opentripplanner.middleware.triptracker.instruction.TripInstruction.TRIP_INSTRUCTION_IMMEDIATE_RADIUS;
 import static org.opentripplanner.middleware.triptracker.instruction.TripInstruction.TRIP_INSTRUCTION_UPCOMING_RADIUS;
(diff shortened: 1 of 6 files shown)


src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
```diff
@@ -294,88 +301,105 @@ private static Stream<Arguments> createInstructionAndStatusCases() {
...
+        Leg multiItinBusLeg = multiLegItinerary.legs.get(multiLegItinerary.legs.size() - 2);
+        Coordinates multiItinFirstLegDestCoords = new Coordinates(multiItinFirstLeg.to);
         Coordinates multiItinLastLegDestCoords = new Coordinates(multiItinLastLeg.to);
+        String ansleyMallPetShopDestinationName = multiItinLastLeg.to.name;
 
         return Stream.of(
             Arguments.of(
                 monitoredTrip,
                 createPoint(firstStepCoords, 1, NORTH_EAST_BEARING),
-                "IMMEDIATE: Head WEST on Adair Avenue Northeast",
```
binh-dam-ibigroup, line 312: I think we'll need the plain text to keep track of the kinds of instructions expected. Or, we can further break down the instructions, this one could be a "directional" instruction for instance.

src/main/java/org/opentripplanner/middleware/triptracker/instruction/ContinueInstruction.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class ContinueInstruction extends SelfLegInstruction {
+    public ContinueInstruction(Step legStep, Locale locale) {
+        this.legStep = legStep;
+        this.locale = locale;
+    }
+
+    @Override
+    public String build() {
+        if (isNotEmpty(legStep) && isNotEmpty(legStep.streetName)) {
```
binh-dam-ibigroup: For objects, can we use `!= null` and for strings, `Strings.isBlank` (or what is the rationale for pulling in a new library)?

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -182,16 +186,55 @@ public static TripInstruction alignTravelerToTrip(
...
+     * instruction providing they are on a walk leg. This will be based on the current or previous step depending on the
+     * traveler's relative position to the next leg.
+     */
+    private static ContinueInstruction getContinueInstruction(
+        TravelerPosition travelerPosition,
+        Step nextStep,
+        Locale locale
+    ) {
+        if (!travelerPosition.expectedLeg.transitLeg && nextStep != null) {
+            Step currentStep = isPositionPastStep(travelerPosition, nextStep)
```
binh-dam-ibigroup: At times, this piece produces the incorrect street name. I'll  send an example where the behavior is not what you/d expect.

binh-dam-ibigroup (changes requested):
Can you address the wrong street name, and put back literals in the tests, please.

src/main/java/org/opentripplanner/middleware/triptracker/instruction/ContinueInstruction.java
br648: Updated: https://github.com/ibi-group/otp-middleware/pull/268/commits/6a1a94ab4b6f8ac32448c66db5b572cbeaf942c0

src/test/java/org/opentripplanner/middleware/controllers/api/TrackedTripControllerTest.java
br648: It would be cleaner to further break down the instruction. Also, this way eliminates time zone issues. But I think that is out of scope. Let me know.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: I've reworked this. Seems to work ok now with the example you provided.

binh-dam-ibigroup (approved):
Thanks for the fixes!

JymDyerIBI (approved):
This looks good.

PR #271, opened 2024-11-20
Trip sharing

binh-dam-ibigroup:
### Checklist

### Description

Provides a new endpoint to query a given user's monitored trips, including trips on which a user is a companion and or an observer. The field type for `primary` user is changed to `MobilityProfileLite` as a result from introducing the endpoint in #266 for querying the dependent profile.

Out of scope: email notifications to users added as companions/observers/primary travelers will be addressed in a separate PR.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
@@ -10,6 +10,7 @@ import org.opentripplanner.middleware.auth.RequestingUser;
 import org.opentripplanner.middleware.controllers.response.ResponseList;
 import org.opentripplanner.middleware.models.ItineraryExistence;
 import org.opentripplanner.middleware.models.MonitoredTrip;
+import org.opentripplanner.middleware.models.OtpUser;
 import org.opentripplanner.middleware.persistence.Persistence;
 import org.opentripplanner.middleware.tripmonitor.jobs.CheckMonitoredTrip;
 import org.opentripplanner.middleware.tripmonitor.jobs.MonitoredTripLocks;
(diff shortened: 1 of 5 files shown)


josh-willis-arcadis (approved):
LGTM!

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
```diff
@@ -176,7 +176,7 @@ public class MonitoredTrip extends Model {
      */
     public boolean notifyAtLeadingInterval = true;
 
-    public RelatedUser primary;
+    public MobilityProfileLite primary;
     public RelatedUser companion;
     public List<RelatedUser> observers = new ArrayList<>();
```
br648, line 181: Probably out of scope, but should any updates to related users be reflected in monitored trips? For instance, if a related user is removed?

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -53,6 +54,16 @@ protected void buildEndpoint(ApiEndpoint baseEndpoint) {
         super.buildEndpoint(modifiedEndpoint);
     }
 
+    @Override
+    protected Bson getEntityFilter(OtpUser user) {
+        return Filters.or(
+            Filters.eq(USER_ID_PARAM, user.id),
+            Filters.eq("primary.userId", user.id),
+            Filters.eq("companion.email", user.email),
+            Filters.eq("observers.email", user.email)
```
br648, line 63: I wasn't sure if this syntax would work given that it is an array. Did a quick test and it does!

br648 (commented):
Just one comment related to the relationship between related users and monitored trips. Also, I was expecting a new end point, but I guess that wasn't needed in the end?

binh-dam-ibigroup:
> Also, I was expecting a new end point, but I guess that wasn't needed in the end?

Indeed!

src/main/java/org/opentripplanner/middleware/models/MonitoredTrip.java
br648: @binh-dam-ibigroup just this to potentially address if you think it is in scope/important?

br648: > @binh-dam-ibigroup just this to potentially address if you think it is in scope/important?

@binh-dam-ibigroup I just need your thoughts on this and then I can approve if it is out of scope.

binh-dam-ibigroup: @br648 If you are ok with this, we can do delete propagation in another PR.

br648: @binh-dam-ibigroup I think it is out of scope. Will approve.

br648 (approved):
Approving but delete propagation needs to be covered in another PR.

PR #275, opened 2024-12-03
OTP-1382 GMAP Push notifications for leg transition

br648:
### Checklist

### Description

Check for upcoming leg transitions (arrive, departure & mode change) and notify observers.


src/main/java/org/opentripplanner/middleware/i18n/Message.java
```diff
@@ -19,7 +19,10 @@ public enum Message {
     ACCEPT_DEPENDENT_EMAIL_SUBJECT,
     ACCEPT_DEPENDENT_EMAIL_MANAGE,
     ACCEPT_DEPENDENT_ERROR,
+    DESTINATION_CHANGE_NOTIFICATION,
```
binh-dam-ibigroup: Rename to something like `ARRIVED_NOTIFICATION` or `AT_DESTINATION_NOTIFICATION`?

src/main/java/org/opentripplanner/middleware/i18n/Message.java
```diff
@@ -19,7 +19,10 @@ public enum Message {
...
     LABEL_AND_CONTENT,
+    MODE_CHANGE_NOTIFICATION,
+    ORIGIN_CHANGE_NOTIFICATION,
```
binh-dam-ibigroup: Rename to something like `DEPARTED_NOTIFICATION` or `LEAVE_ORIGIN_NOTIFICATION`?

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
```diff
@@ -110,4 +111,46 @@ public static TripMonitorNotification createInitialReminderNotification(
...
+                changedLeg.mode,
+                expectedLeg.from.name
+            )
+        );
+    }
+
+    /**
+     * Creates a notification that the origin or destination has changed.
+     */
+    public static TripMonitorNotification createStopChangeNotification(
```
binh-dam-ibigroup: Could you clarify what you had in mind for these notification types? I thought we would need a notification that the person traveling has left their origin or arrived at the destination.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -498,6 +549,95 @@ public TripMonitorNotification checkTripForDelay(NotificationType delayType) {
         return null;
     }
 
+    /**
+     * Check for major upcoming change and notify the traveler.
+     */
+    @Nullable
+    public TripMonitorNotification checkTripForMajorChange(
+        NotificationType delayType,
```
binh-dam-ibigroup: Rename this argument to `notificationType`.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -498,6 +549,95 @@ public TripMonitorNotification checkTripForDelay(NotificationType delayType) {
...
+     */
+    private List<NotificationType> getAllMajorChangeNotificationTypes() {
+        return List.of(NotificationType.MODE_CHANGE, NotificationType.ORIGIN_CHANGE, NotificationType.DESTINATION_CHANGE);
+    }
+
+    /**
+     * Attempt to match on just a single exception, else use the wider exception criteria.
+     */
+    @Nullable
+    private Itinerary getMatchingItinerary(NotificationType delayType, List<Itinerary> itineraries) {
```
binh-dam-ibigroup: Rename `delayType` to `notificationType`

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -498,6 +549,95 @@ public TripMonitorNotification checkTripForDelay(NotificationType delayType) {
...
+            return !ItineraryUtils.stopsMatch(trip.itinerary.legs.get(index).to, itineraryToCheck.legs.get(index).to);
+        }
+        return false;
+    }
+
+    /**
+     * A list of all notification types that are classed as major changes to an itinerary.
+     */
+    private List<NotificationType> getAllMajorChangeNotificationTypes() {
+        return List.of(NotificationType.MODE_CHANGE, NotificationType.ORIGIN_CHANGE, NotificationType.DESTINATION_CHANGE);
```
binh-dam-ibigroup: Could this be a constant?

binh-dam-ibigroup (commented):
Could you clarify the intent behind origin/destination change? I think what we want to do is to notify observers that someone has departed or arrived on their trip.

br648:
> Could you clarify the intent behind origin/destination change? I think what we want to do is to notify observers that someone has departed or arrived on their trip.

Oh, really! I thought I was to check the OTP response for any changes to the traveler's trip and notify them of this. Could I have a couple of use cases just to clarify what is expected please?

binh-dam-ibigroup:
> Oh, really! I thought I was to check the OTP response for any changes to the traveler's trip and notify them of this. Could I have a couple of use cases just to clarify what is expected please?

It is simpler than you think it is. Basically, a notification when the traveler leaves the origin or arrives at destination, and when the traveler changes modes (arrives at a bus stop, gets off the bus, gets on/off the bike if there is a bike involved).

binh-dam-ibigroup:
> Basically, a notification when the traveler leaves the origin

Notifications are for observers who are not traveling with the traveler and their companion.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -86,6 +91,12 @@ public static String getInstruction(
                 }
             }
         }
+
+        try {
+            new CheckMonitoredTrip(travelerPosition.trip).checkForMajorTripChanges(travelerPosition.nextLeg);
+        } catch (Exception e) {
+            LOG.error("Error encountered while checking for major trip changes.", e);
+        }
```
binh-dam-ibigroup: When a traveler approaches a bus leg from a walk leg, this block doesn't get called. In the case of a walk->bus transition, the block at line 67 is executed instead. The code in this block needs to execute with the logic when arriving at the end of a leg where the next leg is on a different mode.

binh-dam-ibigroup (changes requested):
Could you make these changes please:
- convert the departure/arrival change notification into leaving the origin/arriving at destination notifications, and send the notification when the traveler gets to the corresponding location.
- execute mode change notification when approaching the end of a leg where the next leg has a different mode.

src/main/java/org/opentripplanner/middleware/i18n/Message.java
br648: Fixed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/i18n/Message.java
br648: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/models/TripMonitorNotification.java
br648: Refactored and addressed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Refactored and addressed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Refactored and addressed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Refactored and addressed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: Refactored and addressed in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Get the traveler's name if available, if not provide a generic traveler name.
+     */
+    private String getTravelerName() {
+        if (travelerName != null) {
+            return travelerName;
+        } else {
+            return Message.TRIP_TRAVELER_GENERIC_NAME.get(observerLocale);
```
binh-dam-ibigroup: Fallback on email if the name is not available (extract/reuse that code), similar to with notification emails about companion requests and shared trips.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -247,6 +250,31 @@ private boolean isTrackingOngoing() {
...
+        OtpUser otpUser = getOtpUser();
+        if (otpUser != null) {
+            otpUser.relatedUsers.forEach(relatedUser -> {
+                if (relatedUser.isConfirmed()) {
+                    OtpUser observer = Persistence.otpUsers.getOneFiltered(eq("email", relatedUser.email));
+                    if (observer != null) {
+                        enqueueNotification(
+                            LegTransitionNotification.createLegTransitionNotifications(
+                                legTransitionTypes,
+                                otpUser.name,
```
binh-dam-ibigroup: Fallback on the user email (UI is not populating names at this moment.)

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -525,9 +560,12 @@ private void sendNotifications() {
             return;
         }
 
+        Locale locale = getOtpUserLocale();
```
binh-dam-ibigroup: Replace this call to `getOtpUserLocale` with `I18nUtils.getOtpUserLocale(otpUser)` to use the locale of the desired user. Would it make sense to delete method `getOtpUserLocale` because so many calls are user-specific?

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Create {@link TripMonitorNotification} for leg transition based on notification type.
+     */
+    @Nullable
+    public TripMonitorNotification createTripMonitorNotification(NotificationType notificationType) {
+        String body;
+        switch (notificationType) {
+            case MODE_CHANGE_NOTIFICATION:
+                body = String.format(
+                    Message.MODE_CHANGE_NOTIFICATION.get(travelerPosition.locale),
```
binh-dam-ibigroup: `observerLocale` should be used in all notifications.

src/main/java/org/opentripplanner/middleware/utils/ItineraryUtils.java
```diff
@@ -371,4 +371,4 @@ public static Leg getFirstLeg(Itinerary itinerary) {
             .map(legs -> legs.get(0))
             .orElse(null);
     }
-}
+}
```
binh-dam-ibigroup: Revert this change?

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -214,6 +220,65 @@ private static Stream<Arguments> createDelayNotificationTestCases() {
...
+    @MethodSource("createLegTransitionNotificationTestCases")
+    void testLegTransitionNotifications(
+        NotificationType legTransitionType,
+        String travelerName,
+        TravelerPosition travelerPosition,
+        Locale locale,
+        String message
+    ) {
+        TripMonitorNotification[] notification = LegTransitionNotification.createLegTransitionNotifications(
+            List.of(legTransitionType),
```
binh-dam-ibigroup: I didn't initially get why you were using a list, but I see what that does on transitions... I still think that only one transition type should be dealt with as a time as a person arrives at the end of a leg and begins the next one.

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -0,0 +1,103 @@
...
+                travelerName,
+                legTransitionType,
+                travelerPosition,
+                observerLocale
+            );
+            if (legTransitionNotification.tripMonitorNotification != null) {
+                tripMonitorNotifications.add(legTransitionNotification.tripMonitorNotification);
+            }
+        }
+        return tripMonitorNotifications.toArray(new TripMonitorNotification[0]);
```
binh-dam-ibigroup: Change (and rename) `legTransitionTypes` to be a single `NotificationType`, so we can remove the for loop and still return the corresponding notification.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -247,6 +250,31 @@ private boolean isTrackingOngoing() {
         return trip.journeyState.tripStatus == TripStatus.TRIP_ACTIVE && TripTrackingData.getOngoingTrackedJourney(trip.id) != null;
     }
 
+    /**
+     * Process leg transition notifications.
+     */
+    public void processLegTransition(List<NotificationType> legTransitionTypes, TravelerPosition travelerPosition) {
+        OtpUser otpUser = getOtpUser();
+        if (otpUser != null) {
+            otpUser.relatedUsers.forEach(relatedUser -> {
```
binh-dam-ibigroup: The notifications should be sent to the observers on a trip (not to all companions of a user), and to either the primary traveler or the companion on a trip, if they are not the user who started live tracking.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -247,6 +250,31 @@ private boolean isTrackingOngoing() {
...
+                if (relatedUser.isConfirmed()) {
+                    OtpUser observer = Persistence.otpUsers.getOneFiltered(eq("email", relatedUser.email));
+                    if (observer != null) {
+                        enqueueNotification(
```
binh-dam-ibigroup: I think we need to send the notification once, similar to the bus driver notifications. Otherwise, if the traveler stays at an origin or a transition spot for a while, we will be sending the same notification email every few seconds...

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -91,6 +101,46 @@ public static String getInstruction(
...
+
+    private static List<NotificationType> getLegTransitionTypes(TravelerPosition travelerPosition) {
+        List<NotificationType> notificationTypes = new ArrayList<>();
+        if (isAtStartOfLeg(travelerPosition)) {
+            notificationTypes.add(DEPARTED_NOTIFICATION);
+        }
+        if (isAtEndOfLeg(travelerPosition)) {
+            notificationTypes.add(ARRIVED_NOTIFICATION);
+        }
+        if (hasModeChanged(travelerPosition)) {
```
binh-dam-ibigroup: Can the mode change notification include some of the text of the arrival notification too (so you don't have to make two notifications for that and don't have to pass a list of transition types)?

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
```diff
@@ -214,6 +220,65 @@ private static Stream<Arguments> createDelayNotificationTestCases() {
...
+        Coordinates expectedLegDestinationCoords = new Coordinates(expectedLeg.to);
+        Leg nextLeg = itinerary.legs.get(2);
+        Coordinates nextLegDepartureCoords = new Coordinates(nextLeg.from);
+        return Stream.of(
+            Arguments.of(
+                NotificationType.MODE_CHANGE_NOTIFICATION,
+                travelerName,
+                new TravelerPosition(expectedLeg, nextLeg, expectedLegDestinationCoords),
+                locale,
+                "Obi-Wan has changed transit from TRAM to WALK."
```
binh-dam-ibigroup: Refine the text to say that "[traveler] has arrived at transit stop [stop name]", and omit the specific modes.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
```diff
@@ -91,6 +101,46 @@ public static String getInstruction(
...
+        }
+
+    }
+
+    private static List<NotificationType> getLegTransitionTypes(TravelerPosition travelerPosition) {
+        List<NotificationType> notificationTypes = new ArrayList<>();
+        if (isAtStartOfLeg(travelerPosition)) {
+            notificationTypes.add(DEPARTED_NOTIFICATION);
+        }
+        if (isAtEndOfLeg(travelerPosition)) {
```
binh-dam-ibigroup: If arriving from a bus leg with the bus shape in the middle of the road, and the stop is away from the curb, I think the threshold might not be met. Not sure if we want to use the "upcoming" radius instead of the "immediate" radius.

binh-dam-ibigroup (changes requested):
I think the logic is getting there. I don't think we need to identify/process a list of notification types (one should be fine). Notifications are for the observers and non-trip-initiators, and there are language tweaks.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: I've left `getOtpUserLocale()` in place for cases where the notification is for the user related to the trip i.e. `trip.userId`.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: The reason for using `CheckMonitoredTrip` is bc in `sendNotifications(OtpUser otpUser)` there is a check to not send notifications if already sent `if (previousJourneyState.lastNotifications.containsAll(notifications) && !hasInitialReminder)`. Then I noticed that `lastNotifications` is never set! So, I have made an update in this area. The advantage of this approach is that the notification will be sent again if not successful the first time.

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/test/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTripTest.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/triptracker/TravelerLocator.java
br648: Updated in: https://github.com/ibi-group/otp-middleware/pull/275/commits/91d66190cb611283993b8e745148641efd695c5e

src/main/java/org/opentripplanner/middleware/i18n/Message.java
```diff
@@ -19,10 +19,10 @@ public enum Message {
     ACCEPT_DEPENDENT_EMAIL_SUBJECT,
     ACCEPT_DEPENDENT_EMAIL_MANAGE,
     ACCEPT_DEPENDENT_ERROR,
+    ARRIVED_AND_MODE_CHANGE_NOTIFICATION,
```
binh-dam-ibigroup: I'd say just keep the old name `MODE_CHANGE_NOTIFICATION`.

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -66,38 +82,69 @@ public TripMonitorNotification createTripMonitorNotification(NotificationType no
...
+        Set<OtpUser> notifyUsers = new HashSet<>();
+
+        if (trip.ownedByPrimary() && trip.companion != null) {
+            notifyUsers.add(Persistence.otpUsers.getOneFiltered(eq("email", trip.companion.email)));
+        } else if (trip.ownedByCompanion() && trip.primary != null) {
+            notifyUsers.add(Persistence.otpUsers.getById(trip.primary.userId));
         }
+
+        trip.observers.forEach(observer -> {
+            if (observer.isConfirmed()) {
```
binh-dam-ibigroup, line 97: Good thought about using confirmed users only, I forgot to do check that when notifying users when sharing trips!

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -66,38 +82,69 @@ public TripMonitorNotification createTripMonitorNotification(NotificationType no
...
+                notifyUsers.add(Persistence.otpUsers.getOneFiltered(eq("email", observer.email)));
+            }
+        });
+
+        return notifyUsers;
     }
 
     /**
-     * Create locale specific notifications.
+     * If a traveler is on schedule and on either a walk or transit leg check for possible leg transition notification.
```
binh-dam-ibigroup: I think you meant to say "If a traveler is on the route (not deviated)"...

binh-dam-ibigroup: Insert comma before "check".

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -66,38 +82,69 @@ public TripMonitorNotification createTripMonitorNotification(NotificationType no
...
+                    LOG.error("Error encountered while checking leg transition.", e);
+                }
             }
         }
-        return tripMonitorNotifications.toArray(new TripMonitorNotification[0]);
     }
-}
+
+    /**
+     * Depending on the traveler's proximity to the start/end of a leg return the appropriate notification type.
```
binh-dam-ibigroup: Insert comma before "return".

src/test/java/org/opentripplanner/middleware/models/LegTransitionNotificationTest.java
```diff
@@ -0,0 +1,136 @@
...
+            travelerPosition,
+            locale
+        ).tripMonitorNotification;
+        assertNotNull(notification);
+        assertEquals(message, notification.body);
+    }
+
+    private static Stream<Arguments> createLegTransitionNotificationTestCases() throws Exception {
+        String travelerName = "Obi-Wan";
+        Locale locale = Locale.US;
```
binh-dam-ibigroup: Maybe remove `locale` as a test parameter (set that in the test itself).

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -559,6 +592,9 @@ private void sendNotifications() {
         // TODO: better handle below when one of the following fails
         if (successEmail || successPush || successSms) {
             notificationTimestampMillis = DateTimeUtils.currentTimeMillis();
+            // Prevent repeated notifications by saving successfully sent notifications.
```
binh-dam-ibigroup: This check is at the wrong spot, I see duplicate notifications being sent and recorded (see screenshot)

image

binh-dam-ibigroup (changes requested):
Could you address duplicate notifications still being sent, please?

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+    }
+
+    /**
+     * Depending on the traveler's proximity to the start/end of a leg, return the appropriate notification type.
+     */
+    private static NotificationType getLegTransitionNotificationType(TravelerPosition travelerPosition) {
+        if (isAtStartOfLeg(travelerPosition)) {
+            return DEPARTED_NOTIFICATION;
```
binh-dam-ibigroup, line 133: Nitpick: Change the condition so that departing notifications are sent at the beginning of the trip only.

I found one case where I set two coordinates, one close to the end of a leg, and the next one close to the beginning of the next leg, and I got two companion notifications, one for arriving at the end of the previous leg, and one for departing the next leg, which is redundant with the previous one and is thus unneeded noise. What do you think?

binh-dam-ibigroup (approved):
Overall, the changes are working and code looks good. The nice-to-have tweak is not blocking.

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
br648: I think for consistency we leave it as is. The observer would get a notification that the traveler has arrived but not one to tell them they have departed. The next notification would be the traveler arriving at the end of the next leg. Correct?

src/main/resources/Message_fr.properties
```diff
@@ -4,7 +4,10 @@ ACCEPT_DEPENDENT_EMAIL_LINK_TEXT = Accepter la demande
 ACCEPT_DEPENDENT_EMAIL_SUBJECT = Demande d'accompagnateur
 ACCEPT_DEPENDENT_EMAIL_MANAGE = G�rez vos pr�f�rences
 ACCEPT_DEPENDENT_ERROR = La demande d'accompagnateur n'a pas �t� re�ue.
+ARRIVED_NOTIFICATION = TODO %s.
```
br648: @binh-dam-ibigroup can you provide the french messages please!

src/main/java/org/opentripplanner/middleware/models/LegTransitionNotification.java
binh-dam-ibigroup: That's what I observed.

src/main/resources/Message_fr.properties
binh-dam-ibigroup: Oui ! https://github.com/ibi-group/otp-middleware/pull/275/commits/36b26afa7d532112d72346b040408576db40a705

JymDyerIBI (approved):
LGTM

PR #318, opened 2025-03-28
Topic based logging

binh-dam-ibigroup:
### Checklist

- [na] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented

### Description

Note: installing new dependencies of this PR might prevent you from running older code later...

Looking for feedback on other things that could be logged separately!
This PR improves logback.xml so that select log messages for notifications and GMAP and are sent to specific log files, so they are easier to retrieve later.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
+++ b/src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
@@ -2,7 +2,9 @@ package org.opentripplanner.middleware.utils;
 
 import com.google.gson.Gson;
 import com.sparkpost.Client;
+import com.sparkpost.exception.SparkPostException;
 import com.sparkpost.model.responses.Response;
+import com.sparkpost.model.responses.ServerErrorResponse;
 import com.twilio.Twilio;
 import com.twilio.rest.api.v2010.account.Message;
 import com.twilio.rest.verify.v2.service.Verification;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
```diff
@@ -167,6 +169,7 @@ static String sendPush(
                 jsonBody
             );
             if (httpResponse.status == 200) {
+                LOG.info("Push notification sent");
```
daniel-heppner-ibigroup: Would it be good to add some more dynamic details to this log message? Maybe notification ID + user?

daniel-heppner-ibigroup: similar to the email part

daniel-heppner-ibigroup (approved):
This looks good to me! I learned a lot about logback from this PR.

src/main/java/org/opentripplanner/middleware/utils/NotificationUtils.java
binh-dam-ibigroup: Good suggestion, added in https://github.com/ibi-group/otp-middleware/pull/318/commits/ddb31dd2c1d14e6542fc52c9f67a1ff0de557c83. I suppose we will have to find a way to clean up all these API calls.

src/main/resources/logback.xml
```diff
@@ -1,12 +1,56 @@
...
             <pattern>%highlight(%-5level) %d{yyyy-MM-dd'T'HH:mm:ss.SSS} %yellow([%thread]) %blue(%logger{36}\(%file:%line\)) %yellow(%X{prefix}) %msg%n</pattern>
         </encoder>
     </appender>
+    <appender name="NOTIFICATION-FILE" class="RollingFileAppender">
+        <file>logs/notifications.log</file>
+        <rollingPolicy class="TimeBasedRollingPolicy">
+            <!-- Archive logs to a folder -->
+            <fileNamePattern>${ARCHIVED_LOG_FOLDER}/notifications.log</fileNamePattern>
+            <!-- Keep about two months worth of logs -->
+            <maxHistory>62</maxHistory>
```
br648: Nit. Could you make this a global variable and reuse here and below?

br648 (approved):
All good. Right now, nothing comes to mind that we should be logging separately. But, we now have the means to add additional logging easily if need be.

src/main/resources/logback.xml
binh-dam-ibigroup: Yep - https://github.com/ibi-group/otp-middleware/pull/318/commits/a764076884bc19946a6e204a5272507715320245

PR #322, opened 2025-05-01
Entry points for recurring jobs

br648:
### Checklist

### Description

This PR seeks to allow fine grain control over the jobs within the OTP middleware which should be started. This will then allow various instances of OTP middleware to be started which focus on specific jobs instead of all jobs running under a single instance.

Breaking Change!

**_--endpoints-only_** or **_-E_** must be defined in order to get endpoints.


src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+            this.recurringJobs = recurringJobs;
+            return this;
+        }
+
+        public CommandLineTestCase withLoadBalanced(boolean isLoadBalanced) {
+            this.isLoadBalanced = isLoadBalanced;
+            return this;
+        }
+
+        public CommandLineTestCase withCommand(String[] command) {
```
binh-dam-ibigroup: Is it possible to use `String... command` to simplify the syntax of the test cases?

src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+        public String[] command;
+        public String errorMessage;
+        public String message;
+
+        public CommandLineTestCase withConfigFile(String configFile) {
+            this.configFile = configFile;
+            return this;
+        }
+
+        public CommandLineTestCase withRecurringJob(Set<RecurringJob> recurringJobs) {
```
binh-dam-ibigroup: Rename to `withRecurringJobs`.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+    public void parseArguments(String[] arguments) {
+        if (arguments.length > 0 && arguments[0].contains(".yml")) {
+            configFile = arguments[0];
+        }
+        Map<String, Set<String>> groupedParams = parseCommandLineArguments(arguments);
+        groupedParams.forEach((command, commandValues) -> {
+            if (command.equalsIgnoreCase("-recurringjobs")) {
+                defineRecurringJobs(commandValues);
+            }
+            if (command.equalsIgnoreCase("-loadbalance")) {
```
binh-dam-ibigroup: Trying to find a better name for this option without having to reference load balancers... How about something like `--endpoints-only` or `--api-endpoints` because this serves the API endpoints only, correct? (so that a load balancer can dispatch web requests).

binh-dam-ibigroup: For command line flags, we should probably follow the convention used by many Unix programs - typically  double-dash + kebab-case-command (e.g. `--load-balancer`) and either one dash + one letter/number (e.g. `-L`) for shorthand syntax. (Note: OTP uses pascalCase commands as in `--loadBalancer` and that's ok if we want to adopt that.)

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum RecurringJob {
```
binh-dam-ibigroup, line 18: This is probably okay as first approach. I had in mind that some recurring jobs should eventually be handled by the OS (e.g. cron) so that they can be started/terminated regardless of completion status.

src/main/java/org/opentripplanner/middleware/utils/ConfigUtils.java
```diff
@@ -70,17 +70,17 @@ public static boolean getBooleanEnvVar(String var) {
...
-        } else if (args.length == 0) {
-            LOG.warn("Using default env.yml: {}", DEFAULT_ENV);
-            envConfig = yamlMapper.readTree(new FileInputStream(DEFAULT_ENV));
+        } else if (configFile.equalsIgnoreCase(DEFAULT_ENV)) {
+            LOG.warn("Using default env.yml: {}", configFile);
+            envConfig = yamlMapper.readTree(new FileInputStream(configFile));
         } else {
-            LOG.info("Loading env.yml: {}", args[0]);
-            envConfig = yamlMapper.readTree(new FileInputStream(args[0]));
+            LOG.info("Loading env.yml: {}", configFile);
```
binh-dam-ibigroup: Nit - Could you combine these two blocks?

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -57,58 +54,35 @@ public class OtpMiddlewareMain {
...
-                    tripSurveyUploadJob,
-                    0,
-                    1,
-                    TimeUnit.DAYS
-                );
+            Set<RecurringJob> recurringJobs = processor.getRecurringJobs();
+            if (!recurringJobs.isEmpty()) {
+                recurringJobs.forEach(RecurringJob::schedule);
+                try {
+                    // Block the main thread until the scheduler is terminated to prevent the application from
```
binh-dam-ibigroup, line 79: Maybe this is why I'd prefer recurring jobs to be scheduled by the OS itself?

binh-dam-ibigroup (commented):
First round of comments. I still need to play around with the changes, but I would like to explore using cron for some recurring jobs.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
br648: My thinking was to keep this generic, so anything that could be load balanced would be covered. At the moment, the only item is the API, so will change it to suit this.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
br648: Yes, this isn't ideal. My approach for this first iteration was to not change the existing scheduled services. In the case of `BugsnagJobs` it actually schedules three different jobs with different parameters. Perhaps for a specific job (e.g. --cron-job) that qualifies, I can rearchitecture to have a `Main` entry point? We still need to consider backwards compatibility.

binh-dam-ibigroup: It's fine to take it one step at a time.

src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
```diff
@@ -0,0 +1,122 @@
...
+            this.configFile = configFile;
+            return this;
+        }
+
+        public CommandLineTestCase withRecurringJobs(Set<RecurringJob> recurringJobs) {
+            this.recurringJobs = recurringJobs;
+            return this;
+        }
+
+        public CommandLineTestCase withLoadBalanced(boolean isLoadBalanced) {
```
binh-dam-ibigroup: Rename to `withEndpoints` (rename the parameter accordingly too).

src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
```diff
@@ -0,0 +1,122 @@
...
+                .withCommand(recurringJobFlagShorthand, "unknown-job")
+                .withMessage("Unknown job, throw exception.")
+                .withErrorMessage(CommandLineProcessor.getRecurringJobErrorMessage("unknown-job"))
+        );
+    }
+
+    private static class CommandLineTestCase {
+        public String configFile = DEFAULT_ENV;
+        public Set<RecurringJob> recurringJobs = RecurringJob.getAllRecurringJobs();
+        public boolean isLoadBalanced = true;
```
binh-dam-ibigroup: Rename to `hasEndpoints`.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -57,58 +54,35 @@ public class OtpMiddlewareMain {
...
+                    // immediately stopping.
+                    Scheduler.schedulerService.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+                } catch (InterruptedException e) {
+                    LOG.info("Shutting down.", e);
+                    Thread.currentThread().interrupt();
+                }
             }
         }
+
+        if (processor.hasEndPoints()) {
```
binh-dam-ibigroup: Move this block to before the scheduler stuff. Otherwise, because of the thread blocking above, this block is never executed (until termination), and therefore no endpoints are available.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+                ));
+            }
+        }
+    }
+
+    /**
+     * Define if the end points should be enabled.
+     */
+    private void defineEndPoints(Set<String> endPointArguments) {
+        if (endPointArguments.size() == 1 && endPointArguments.contains("yes")) {
```
binh-dam-ibigroup: In how the parsing code works, is it possible to just pass a `--endpoint-only` flag without the 
"yes" or "no"? (so that `--endpoints-only` without a parameter implies "yes")

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.opentripplanner.middleware.utils.ConfigUtils.DEFAULT_ENV;
+
+/**
+ * Parses the command line to extract which jobs and/or services should be started. The only stipulation, is that if a
+ * config file is defined, it must be the first argument defined. e.g.
+ *
+ * /configurations/test/env.yml --endpoints-only yes --recurring-jobs monitor-all-trips-job
+ */
+public class CommandLineProcessor {
+    public static final List<String> END_POINTS_ONLY_FLAGS = List.of("--endpoints-only", "-E");
```
binh-dam-ibigroup: Sorry, rename to `--endpoints` because the flag is not meant to exclude the other features.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public void parseArguments(String[] arguments) {
+        if (arguments.length > 0 && arguments[0].contains(".yml")) {
+            configFile = arguments[0];
+        }
+        Map<String, Set<String>> groupedParams = parseCommandLineArguments(arguments);
+        groupedParams.forEach((flag, commandValues) -> {
+            if (RECURRING_JOB_FLAGS.contains(flag)) {
+                defineRecurringJobs(commandValues);
+            }
```
binh-dam-ibigroup, line 49: To only enable the endpoints and run no recurring jobs, the command line would read `--recurring-jobs`, and that feels kinda strange. Maybe including a `--recurring-jobs no` or `--recurring-jobs none` in addition to just `--recurring-jobs` would help reduce confusion.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public CommandLineProcessor() {
+        this.recurringJobs = RecurringJob.getAllRecurringJobs();
+        this.hadEndPoints = true;
+        this.configFile = DEFAULT_ENV;
+    }
+
+    /**
+     * Parse the arguments provided on the command line. To maintain backward compatibility, if a config file is
+     * required, it must be the first argument and no command e.g. -config is NOT required.
```
binh-dam-ibigroup:
```suggestion
     * required, it must be the first argument, and no flag (e.g. --config) is used.
```

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
```diff
@@ -0,0 +1,64 @@
...
+        this.commandLineName = commandLineName;
+        this.recurringJobScheduler = recurringJobScheduler;
+    }
+
+    public static RecurringJob getJobFromCommandLineArgument(String commandLineArgument) {
+        return Arrays.stream(RecurringJob.values())
+            .filter(job -> job.commandLineName.equalsIgnoreCase(commandLineArgument))
+            .findFirst()
+            .orElse(null);
+    }
```
binh-dam-ibigroup, line 41: How does this method differ from the native `Enum.valueOf()`?

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public static Set<RecurringJob> getAllRecurringJobs() {
+        return new HashSet<>(Arrays.asList(RecurringJob.values()));
```
binh-dam-ibigroup:
```suggestion
        return Set.of(RecurringJob.values());
```

binh-dam-ibigroup (changes requested):
Moving in the right direction, but endpoints are not working at all as written. See also other suggestions.

src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
br648: Doh! My bad.

src/test/java/org/opentripplanner/middleware/utils/CommandLineProcessorTest.java
br648: Doh! My bad.

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
br648: This will at least allow individual jobs to be stopped/started.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
br648: It can and I have updated, but this is now NOT backward compatible.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
br648: I have updated with `--recurring-jobs none`

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
br648: It is similar, but the IllegalArgumentException I throw contains more relevant information i.e. valid jobs.

br648:
@binh-dam-ibigroup because of this:

BREAKING CHANGE: The command --endpoints (or -E) is required for end points to be initialised.

CI is failing. Is this something you can address?

src/main/java/org/opentripplanner/middleware/recurringjobs/RecurringJob.java
br648: The Set needs to be modifiable.

binh-dam-ibigroup: > The Set needs to be modifiable.

Please add a comment about where this requirement comes from.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -57,58 +54,37 @@ public class OtpMiddlewareMain {
...
+        ConfigUtils.loadConfig(processor.getConfigFile());
         // Connect to MongoDB.
         Persistence.initialize();
 
-        MonitoredComponent.initializeMonitoredComponentsFromConfig();
-
-        initializeHttpEndpoints();
+        if (processor.hasEndPoints()) {
+            // This must take place before the main thread is blocked below.
+            TemplateUtils.initialize();
```
binh-dam-ibigroup: `TemplateUtils.initialize();` is for notification content... Move this to outside the if block. It is needed for endpoint mode (when saving a trip the first time, monitored trip stuff runs and notifications might be sent then) and in job-only mode. Without that, in jobs-only mode, notifications will not work.

src/main/java/org/opentripplanner/middleware/OtpMiddlewareMain.java
```diff
@@ -57,58 +54,37 @@ public class OtpMiddlewareMain {
...
+                    // Block the main thread until the scheduler is terminated to prevent the application from
+                    // immediately stopping.
+                    Scheduler.schedulerService.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+                } catch (InterruptedException e) {
+                    LOG.info("Shutting down.", e);
+                    Thread.currentThread().interrupt();
+                }
             }
         }
+
```
binh-dam-ibigroup: Remove extra blank line

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+    public static String getRecurringJobErrorMessage(String unknownJobName) {
+        return String.format(RECURRING_JOB_ERROR_MESSAGE, unknownJobName, RecurringJob.getAllCommandLineNames());
+    }
+
+    public static String[] getDefaultArguments() {
+        return new String[]{END_POINTS_ONLY_FLAGS.get(0)};
+    }
+
+    public static String[] getDefaultArguments(String configFile) {
+        return new String[]{configFile, END_POINTS_ONLY_FLAGS.get(0)};
```
binh-dam-ibigroup: Insert space after `]` (2 instances).

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
binh-dam-ibigroup: With the latest changes, `--endpoints no` exposes endpoints.
I should have clarified we need the following:
- `--endpoints yes` => expose endpoints
- `--endpoints no` => hide endpoints
- `--endpoints` (for convenience) => same as `--endpoints yes`.

binh-dam-ibigroup (changes requested):
Sorry, a few more chinks:
- Make `--endpoints no` not expose endpoints
- Initialize templates for all execution modes.

src/main/java/org/opentripplanner/middleware/utils/CommandLineProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Define if the end points should be enabled.
+     */
+    private void defineEndPoints(Set<String> endPointArguments) {
+        if (endPointArguments.isEmpty() || (endPointArguments.size() == 1 && endPointArguments.contains("yes"))) {
+            hadEndPoints = true;
+        } else if (endPointArguments.size() == 1 && endPointArguments.contains("no")) {
+            hadEndPoints = false;
+        }
```
binh-dam-ibigroup, line 109: Nice to have: I think we need to throw an `IllegalArgumentException` in other cases.

binh-dam-ibigroup (approved):
One nice-to-have but o/w I think it looks good with all the changes.

JymDyerIBI (approved):
👍  Looks good. I'm adapting one of my tickets to use the `RecurringJob` class.

PR #328, opened 2025-07-02
fix(ConnectedDataManager): Fix job interval unit.

binh-dam-ibigroup:
### Checklist

### Description

Setting `CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_FREQUENCY_IN_MINUTES` to 5 "minutes" is resulting in the trip upload job to run every 5 milliseconds!! This PR fixes that.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
+++ b/src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
@@ -141,7 +141,7 @@ public class ConnectedDataManager implements RecurringJobScheduler {
             Scheduler.scheduleJob(
                 new TripHistoryUploadJob(),
                 initialDelayMillis,
-                CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_FREQUENCY_IN_MINUTES,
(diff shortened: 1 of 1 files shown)


JymDyerIBI (approved):
👍  That should certainly work.

src/main/java/org/opentripplanner/middleware/connecteddataplatform/ConnectedDataManager.java
```diff
@@ -141,7 +141,7 @@ public void scheduleRecurringJob() {
             Scheduler.scheduleJob(
                 new TripHistoryUploadJob(),
                 initialDelayMillis,
-                CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_FREQUENCY_IN_MINUTES,
+                CONNECTED_DATA_PLATFORM_TRIP_HISTORY_UPLOAD_JOB_FREQUENCY_IN_MINUTES * 60000L, // milliseconds
```
br648, line 144: I might be missing something here, but would it not be better to change `TimeUnit.MILLISECONDS` to `TimeUnit.MINUTES` instead?

br648: Wait. That would mess the initial delay up! Coffee brb.

br648 (approved):

PR #337, opened 2025-08-08
Monitor trip day refactor

br648:
### Checklist

### Description

Simplify the process to obtain required trip day.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
+++ b/src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
@@ -364,7 +364,7 @@ public class CheckMonitoredTrip implements Runnable {
                         );
                     }
 
-                    advanceToNextActiveTripDate();
+                    targetZonedDateTime = advanceToNextActiveTripDate(targetZonedDateTime);
                     updateMonitoredTrip();
 
                     if (matchingItinerary.hasEnded()) {
@@ -786,7 +786,7 @@ public class CheckMonitoredTrip implements Runnable {
             matchingItinerary = previousMatchingItinerary.clone();
         }
 
-        setTargetDateTime();
+        targetZonedDateTime = computeTargetDateTime();
 
         if (isPreviousTripOngoingAtLastCheck()) {
             matchingItinerary = previousMatchingItinerary;
@@ -851,9 +851,9 @@ public class CheckMonitoredTrip implements Runnable {
      * initializing the appropriate target time. Start by checking today's date at the earliest in case the user has
      * paused trip monitoring for a while.
      */
-    private void setTargetDateTime() {
+    private ZonedDateTime computeTargetDateTime() {
         // Initialize to the current date with the trip's hours and minutes.
-        targetZonedDateTime = trip.tripZonedDateTime(DateTimeUtils.nowAsLocalDate());
+        ZonedDateTime nextTripDateTime = trip.tripZonedDateTime(DateTimeUtils.nowAsLocalDate());
 
         // Attempt to advance to the next monitored day, except for one-time trips
         // or if tracking is ongoing or if the matching itinerary is still valid.
@@ -868,15 +868,16 @@ public class CheckMonitoredTrip implements Runnable {
                     DateTimeUtils.DEFAULT_DATE_FORMAT_PATTERN
                 );
                 if (
-                    lastDate.getYear() == targetZonedDateTime.getYear() &&
-                    lastDate.getMonthValue() == targetZonedDateTime.getMonthValue() &&
-                    lastDate.getDayOfMonth() == targetZonedDateTime.getDayOfMonth()
+                    lastDate.getYear() == nextTripDateTime.getYear() &&
+                    lastDate.getMonthValue() == nextTripDateTime.getMonthValue() &&
+                    lastDate.getDayOfMonth() == nextTripDateTime.getDayOfMonth()
                 ) {
-                    targetZonedDateTime = targetZonedDateTime.plusDays(1);
+                    nextTripDateTime = nextTripDateTime.plusDays(1);
                 }
             }
-            advanceToNextActiveTripDate();
+            nextTripDateTime = advanceToNextActiveTripDate(nextTripDateTime);
         }
+        return nextTripDateTime;
     }
 
     /**
@@ -991,8 +992,8 @@ public class CheckMonitoredTrip implements Runnable {
      * itinerary and journey state data the appropriate number of days and then recalculate the new scheduled time
      * based on the updated time in the updated itinerary.
      */
-    private void advanceToNextActiveTripDate() {
-        targetZonedDateTime = findNextMonitoredDay(trip, targetZonedDateTime);
+    private ZonedDateTime advanceToNextActiveTripDate(ZonedDateTime targetZonedDateTime) {
+        ZonedDateTime nextTripDateTime = findNextMonitoredDay(trip, targetZonedDateTime);
 
         LOG.info("Next itinerary happening on {}.", targetZonedDateTime);
 
@@ -1006,6 +1007,7 @@ public class CheckMonitoredTrip implements Runnable {
         );
 
         applyDelayOffset(scheduledTime);
+        return nextTripDateTime;
     }
 
     /**
--- a/src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
+++ b/src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
@@ -498,7 +498,13 @@ public class CheckMonitoredTrip implements Runnable {
      * or future
      */
     private void updateTripStatus() {
-        journeyState.tripStatus = matchingItinerary.isActive() ? TripStatus.TRIP_ACTIVE : TripStatus.TRIP_UPCOMING;
+        if (matchingItinerary.isActive()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -854,34 +822,75 @@ public boolean shouldSkipMonitoredTripCheck(boolean persist) throws Exception {
...
+        LOG.debug("Trip criteria not met to check. Skipping.");
         return true;
     }
 
+    /**
+     * Calculate target time for the next trip plan request find the next possible day the trip is active by
+     * initializing the appropriate target time. Start by checking today's date at the earliest in case the user has
+     * paused trip monitoring for a while.
+     */
+    private void setTargetDateTime() {
```
binh-dam-ibigroup: Can you create function that computes and returns the date/time of the next trip?

br648: @binh-dam-ibigroup When you say function are you thinking Supplier?

binh-dam-ibigroup: `public ZonedDateTime computeTargetDateTime() { ... }` - Nothing fancy. Then, in `shouldSkipMonitoredTripCheck`, write:
```
targetZoneDateTime = computeTargetDateTime();
```

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -854,34 +822,75 @@ public boolean shouldSkipMonitoredTripCheck(boolean persist) throws Exception {
...
+                LocalDate lastDate = DateTimeUtils.getDateFromString(
+                    previousJourneyState.targetDate,
+                    DateTimeUtils.DEFAULT_DATE_FORMAT_PATTERN
+                );
+                if (
+                    lastDate.getYear() == targetZonedDateTime.getYear() &&
+                    lastDate.getMonthValue() == targetZonedDateTime.getMonthValue() &&
+                    lastDate.getDayOfMonth() == targetZonedDateTime.getDayOfMonth()
+                ) {
+                    targetZonedDateTime = targetZonedDateTime.plusDays(1);
```
binh-dam-ibigroup: Trying to avoid setting intermediate values to that field. A local variable should be used for these intermediate assignments, here and at L856 above, and inside `advanceToNetActiveTripDate`.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -856,34 +824,79 @@ public boolean shouldSkipMonitoredTripCheck(boolean persist) throws Exception {
...
+    /**
+     * Define the number of minutes until the start of a trip. If dealing with a one time trip, the matching itinerary
+     * is unlikely to be defined in which case use the original trip start time instead.
+     */
+    private long getMinutesUntilNextTrip() {
+        // Get current time and trip time (with the time offset to today) for comparison.
+        ZonedDateTime now = DateTimeUtils.nowAsZonedDateTime();
+
+        Instant tripStartInstant = !trip.isOneTime() && !isPreviousTripOngoingAtLastCheck()
+            ? findEarliestTargetDate(trip, now).toInstant()
```
br648: @binh-dam-ibigroup This is now the same as `getMinutesUntilTrip()`. I don't see how we can get all the unit tests to pass without using the `findEarliestTargetDate` method.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -366,7 +366,7 @@ private boolean makeOTPRequestAndUpdateMatchingItineraryInternal() {
                         );
                     }
 
-                    advanceToNextActiveTripDate();
+                    targetZonedDateTime = advanceToNextActiveTripDate(targetZonedDateTime);
```
binh-dam-ibigroup: Can we reuse `computeTargetDateTime()` here too?

binh-dam-ibigroup: > Can we reuse `computeTargetDateTime()` here too?

Or can `computeTargetDateTime()` also include the logic between lines 356-366?

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
binh-dam-ibigroup: `findEarliestTargetDate` is the way to go.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
binh-dam-ibigroup: `findEarliestTargetDate` is the way to go. I did side effects trying to compute `targetZoneDateTime` using `findEarliestTargetDate`:
- Test `shouldReportRecurringTripInstanceInPastAsUpcoming` fails because previously, a call to `updateTripStatus` was made during `shouldSkipMonitoredTripCheck`. The trip status should be updated too.
- One of the tests in `canHandleChangingMonitoredDays` fails because a target date is computed without accounting for the fact that a trip is still ongoing.

binh-dam-ibigroup: See if the suggestion from #339 makes sense (more refactoring after that is possible).

br648:
@binh-dam-ibigroup I've reinstated `computeTargetZonedDateTime()` so `targetZonedDateTime` is now only initialized here. Which I think is a step forward. Unfortunately `targetZonedDateTime` is referenced all over the place, so returning the value isn't possible (as previously covered). If you think it is benefical, I'll update to pass `targetZonedDateTime` around and look to remove it as a global variable?

br648:
BTW, I hope the unit tests cover all use cases!

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -920,7 +905,7 @@ public static ZonedDateTime findEarliestTargetDate(MonitoredTrip trip, ZonedDate
             trip.itinerary.startTime.toInstant()
         );
 
-        return findNextMonitoredDay(trip, nextStartDay);
+        return findNextMonitoredDay (trip, nextStartDay);
```
binh-dam-ibigroup: Revert this change.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -856,34 +832,43 @@ public boolean shouldSkipMonitoredTripCheck(boolean persist) throws Exception {
...
+        LOG.debug("Trip criteria not met to check. Skipping.");
         return true;
     }
 
+    /**
+     * Calculate target time for the next trip plan request find the next possible day the trip is active by
+     * initializing the appropriate target time.
+     */
+    private void computeTargetZonedDateTime() {
+        targetZonedDateTime = findEarliestTargetDate(trip, DateTimeUtils.nowAsZonedDateTime());
```
binh-dam-ibigroup: Move this line to above line 868.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
```diff
@@ -856,34 +832,43 @@ public boolean shouldSkipMonitoredTripCheck(boolean persist) throws Exception {
...
+        if (matchingItinerary.isActive()) {
+            targetZonedDateTime = DateTimeUtils.makeOtpZonedDateTime(matchingItinerary.startTime);
+        } else if (journeyState.tripStatus == TripStatus.TRIP_UPCOMING) {
+            targetZonedDateTime = advanceToNextActiveTripDate(targetZonedDateTime);
```
binh-dam-ibigroup: We probably don't need to call `advanceToNextActiveTripDate` and we could delete that method, because it is redundant with `findEarliestTargetDate` - unless there is something you spotted? Tests about `CheckMonitoredTrip` still seem to pass without this call.

binh-dam-ibigroup (commented):
See if we can delete some other redundant code.

binh-dam-ibigroup:
> If you think it is benefical, I'll update to pass `targetZonedDateTime` around and look to remove it as a global variable?

We can deal with that in a separate PR.

src/main/java/org/opentripplanner/middleware/tripmonitor/jobs/CheckMonitoredTrip.java
br648: I've simplified this method even further. Tests still pass.

br648:
@binh-dam-ibigroup I think this is pretty close to being done now.

binh-dam-ibigroup (approved):
Much cleaner and everything still seems to work, nice work!

JymDyerIBI (approved):
👍  Looks good.

PR #345, opened 2025-09-12
Fix bus to walk transition

binh-dam-ibigroup:
### Checklist

### Description

This PR addresses an issue with live tracking where a traveler would be deemed arrived at their destination when instead they would be between a bus and a walk leg (see screenshot). With this change, the traveler is directed to the first segment of the next leg.

image


JymDyerIBI (approved):
👍  Looks good.

src/main/java/org/opentripplanner/middleware/triptracker/TravelerPosition.java
```diff
@@ -93,20 +93,27 @@ public TravelerPosition(TrackedJourney trackedJourney, Itinerary itinerary, OtpU
...
+
+    /**
+     * Returns the closest leg in 'upcoming' radius.
+     */
+    public Leg getLegWithClosestUpcomingOrigin(boolean isTransit) {
+        double distanceToExpectedLeg = distanceToLegOrigin(currentPosition, expectedLeg, isTransit);
+        double distanceToNextLeg = distanceToLegOrigin(currentPosition, nextLeg, isTransit);
 
-        if (distancetoExpectedLeg <= TRIP_INSTRUCTION_UPCOMING_RADIUS && distancetoExpectedLeg < distanceToNextLeg) {
+        if (distanceToExpectedLeg <= TRIP_INSTRUCTION_UPCOMING_RADIUS && distanceToExpectedLeg < distanceToNextLeg) {
```
br648: Out of interest, I thought I'd run this through CoPilot. It could be 'simplified' to the following, but I'm not sure if it's easier to read. Anyway, you can change it, if you like!

`if (distanceToExpectedLeg <= TRIP_INSTRUCTION_UPCOMING_RADIUS || distanceToNextLeg <= TRIP_INSTRUCTION_UPCOMING_RADIUS) {
            return (distanceToExpectedLeg <= distanceToNextLeg) ? expectedLeg : nextLeg;
        }
        return null;`

br648 (approved):
All good and the test passes. Just one minor (slightly silly) comment!

src/main/java/org/opentripplanner/middleware/triptracker/TravelerPosition.java
binh-dam-ibigroup: Good suggestion, updated in https://github.com/ibi-group/otp-middleware/pull/345/commits/50739c5b871442eeb64f9b144b8d3223cc0569c6

PR #346, opened 2025-09-20
Update User Preferred Language in Auth0 Metadata

amy-corson-ibigroup:
### Checklist

- [x] The description lists all relevant PRs included in this release _(remove this if not merging to master)_

### Description
Creates a new method `updateAuth0Language` that takes the user's new `preferredLanguage` as set in the application, and updates it in the Auth0 user metadata. This is in order to send account emails (such as password reset emails) to users in their preferred language (currently all emails are sent in English)

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -11,6 +11,7 @@ import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.HttpStatus;
 import org.opentripplanner.middleware.bugsnag.BugsnagReporter;
 import org.opentripplanner.middleware.models.AbstractUser;
+import org.opentripplanner.middleware.models.OtpUser;
 import org.opentripplanner.middleware.persistence.TypedPersistence;
 import org.opentripplanner.middleware.utils.HttpResponseValues;
 import org.opentripplanner.middleware.utils.HttpUtils;
@@ -23,6 +24,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.HashMap;
+import java.util.Objects;
 
 import static com.mongodb.client.model.Filters.eq;
 import static org.opentripplanner.middleware.utils.ConfigUtils.getConfigPropertyAsText;
@@ -135,6 +138,25 @@ public class Auth0Users {
         }
         return null;
     }
+
+    public static User updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
+        if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
+            try {
+                Map<String, Object> userMetadata = new HashMap<>();
+                userMetadata.put("lang", user.preferredLocale);
+                User auth0user = new User();
+                auth0user.setUserMetadata(userMetadata);
+                return getManagementAPI()
+                        .users()
+                        .update(user.auth0UserId, auth0user)
+                        .execute();
+            } catch (Auth0Exception e) {
+                BugsnagReporter.reportErrorToBugsnag("Could not update metadata", e);
+                return null;
+            }
+        }
+        return null;
+    };
 
     /**
      * Method to trigger an Auth0 job to resend a verification email. Returns an Auth0 {@link Job} which can be used to
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -150,6 +150,8 @@ public class Auth0Users {
             try {
                 Map<String, Object> userMetadata = new HashMap<>();
                 userMetadata.put("lang", user.preferredLocale);
+                // Passing the entire user here will return a 403, because Auth0 sees it as you overwriting
+                // read-only fields, so create a blank user and give it only the data we want to update.
                 User auth0user = new User();
                 auth0user.setUserMetadata(userMetadata);
                 getManagementAPI()
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -139,7 +139,13 @@ public class Auth0Users {
         return null;
     }
 
+    /**
+     * Method to update a user's Auth0 metadata to include their preferred language as set in OTP-RR. This will
+     * ensure that account emails (such as email verification and reset password emails) are sent in the correct
+     * language.
+     */
     public static User updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
+        // Only call API if the user language has changed
         if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
             try {
                 Map<String, Object> userMetadata = new HashMap<>();
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -148,12 +148,10 @@ public class Auth0Users {
         // Only call API if the user language has changed
         if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
             try {
-                Map<String, Object> userMetadata = new HashMap<>();
-                userMetadata.put("lang", user.preferredLocale);
                 // Passing the entire user here will return a 403, because Auth0 sees it as you overwriting
                 // read-only fields, so create a blank user and give it only the data we want to update.
                 User auth0user = new User();
-                auth0user.setUserMetadata(userMetadata);
+                auth0user.setUserMetadata(Map.of("lang", user.preferredLocale));
                 getManagementAPI()
                         .users()
                         .update(user.auth0UserId, auth0user)
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -144,7 +144,7 @@ public class Auth0Users {
      * ensure that account emails (such as email verification and reset password emails) are sent in the correct
      * language.
      */
-    public static User updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
+    public static void updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
         // Only call API if the user language has changed
         if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
             try {
@@ -152,16 +152,14 @@ public class Auth0Users {
                 userMetadata.put("lang", user.preferredLocale);
                 User auth0user = new User();
                 auth0user.setUserMetadata(userMetadata);
-                return getManagementAPI()
+                getManagementAPI()
                         .users()
                         .update(user.auth0UserId, auth0user)
                         .execute();
             } catch (Auth0Exception e) {
                 BugsnagReporter.reportErrorToBugsnag("Could not update metadata", e);
-                return null;
             }
         }
-        return null;
     };
 
     /**
--- a/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
+++ b/src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
@@ -140,6 +140,26 @@ public class Auth0Users {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/AbstractUserController.java
```diff
@@ -177,6 +178,7 @@ U preCreateHook(U user, Request req) {
     @Override
     U preUpdateHook(U user, U preExistingUser, Request req) {
         Auth0Users.validateExistingUser(user, preExistingUser, req, this.persistence);
+        updateAuth0Language((OtpUser) user, (OtpUser) preExistingUser);
```
br648: I would move this to line 185 and then use `otpUser` and `existingOtpUser` variables.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -136,6 +139,31 @@ public static User getUserByEmail(String email, boolean createIfNotExists) {
         return null;
     }
 
+    /**
+     * Method to update a user's Auth0 metadata to include their preferred language as set in OTP-RR. This will
+     * ensure that account emails (such as email verification and reset password emails) are sent in the correct
+     * language.
+     */
+    public static User updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
```
br648: As it stands, nothing is done with the returned `User` so this method can be updated to be `void`. This will also allow you to remove the `return` statements.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -136,6 +139,31 @@ public static User getUserByEmail(String email, boolean createIfNotExists) {
...
+        // Only call API if the user language has changed
+        if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
+            try {
+                Map<String, Object> userMetadata = new HashMap<>();
+                userMetadata.put("lang", user.preferredLocale);
+                User auth0user = new User();
```
br648, line 173: Instead of creating a new user, use the method above this one to get the Auth0 user, if they exist, and update that.

br648 (changes requested):
This is on the right track. Need to consider testing next, whether that is a unit test in otp-middleware or in the UI. The test `OtpUserControllerTest` -> `canPreserveSmsConsentDate` is an example of a PUT request. This along with the use of `auth0User` created on line 87 would be a good starting point if you decide to go down this route.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
amy-corson-ibigroup: Good point, thank you!

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
amy-corson-ibigroup: I thought this too, but when I tried updating the entire user I got a 403 error from the server, because the API sees it as you trying to overwrite read-only user data (such as when the user was created, etc). This seemed like the cleanest way to touch only the data we wanted to change

br648: Interesting, and this doesn't overwrite existing user values with defaults? I guess if it works then this is fine.

br648: @amy-corson-ibigroup before I forget, can you add a comment explaining this please.

amy-corson-ibigroup: Tested this again just to make absolutely sure, and I think we're safe. Thanks for the reminder, added comment in [460d9d5](https://github.com/ibi-group/otp-middleware/pull/346/commits/460d9d5a1e0083197ca019aa935c2cf9b5198aeb)

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -83,8 +80,9 @@ public static void setUp() throws Exception {
         dependentUserThree = PersistenceTestUtils.createUser(ApiTestUtils.generateEmailAddress("dependent-three"));
 
         relatedUserFour = PersistenceTestUtils.createUser(ApiTestUtils.generateEmailAddress("related-user-four"));
+        relatedUserFour.preferredLocale = "fr";
```
br648: I would set this in the test so it has context.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -83,8 +80,9 @@ public static void setUp() throws Exception {
...
 
-        User auth0User = createAuth0UserForEmail(relatedUserFour.email, TEMP_AUTH0_USER_PASSWORD);
+        auth0User = createAuth0UserForEmail(relatedUserFour.email, TEMP_AUTH0_USER_PASSWORD);
```
br648: There is no need to do this bc the `relatedUserFour` gets the `auth0User` id on the next line. They are then related. See further comments below on this...

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -294,6 +292,20 @@ void canRemoveRelatedUserOnDelete() {
         assertEquals(RelatedUser.RelatedUserStatus.INVALID, updatedTripWithCompanionAndObserver.observers.get(0).status);
     }
 
+    @Test
+    void canPassUserMetadataToAuth0() throws Exception {
+        Map<String, Object> relatedUserFourMetadata = new HashMap<>();
+        relatedUserFourMetadata.put("lang", relatedUserFour.preferredLocale);
+        auth0User.setUserMetadata(relatedUserFourMetadata);
```
br648: This is only updating the `auth0User` obj created in this class and is also doing the work of your new method. Instead, I would assert that the current `auth0UserProfile` (created in the setUp()) does not contain the preferred locale:

`
User auth0UserProfile = getUserByEmail(relatedUserFour.email, false);
Assertions.assertNull(auth0UserProfile.getUserMetadata());
`

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -294,6 +292,20 @@ void canRemoveRelatedUserOnDelete() {
...
+        makeRequest(
+                String.format("api/secure/user/%s", relatedUserFour.id),
+                JsonUtils.toJson(relatedUserFour),
+                getMockHeaders(relatedUserFour),
+                HttpMethod.PUT
+        );
+        assertTrue(auth0User.getUserMetadata().containsValue(relatedUserFour.preferredLocale));
```
br648: Instead, do a similar check to above, but this time confirm the user meta data contains the preffered locale.

br648 (changes requested):
Update to test the auth0 user profile before and after the PUT request is made.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -23,11 +23,7 @@
 import org.opentripplanner.middleware.utils.HttpResponseValues;
 import org.opentripplanner.middleware.utils.JsonUtils;
 
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
br648: Nit. Revert this update to explicitly reference required classes.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -294,6 +291,28 @@ void canRemoveRelatedUserOnDelete() {
...
+                String.format("api/secure/user/%s", relatedUserFour.id),
+                JsonUtils.toJson(relatedUserFour),
+                getMockHeaders(relatedUserFour),
+                HttpMethod.PUT
+        );
+
+        // Confirm that the same user now has language metadata set correctly
+        User updatedAuth0UserProfile = getUserByEmail(relatedUserFour.email, false);
+        assertNotNull(updatedAuth0UserProfile);
+        assertEquals("fr", updatedAuth0UserProfile.getUserMetadata().get("lang"));
```
br648: Nit. You could use `relatedUserFour.preferredLocale` or create a new param `String preferredLocale = "fr"` and use this in both locations.

src/test/java/org/opentripplanner/middleware/controllers/api/OtpUserControllerTest.java
```diff
@@ -294,6 +291,28 @@ void canRemoveRelatedUserOnDelete() {
...
+        // Confirm the user as currently exists in Auth0 does not contain metadata
+        User auth0UserProfile = getUserByEmail(relatedUserFour.email, false);
+        assertNotNull(auth0UserProfile);
+        assertNull(auth0UserProfile.getUserMetadata());
+
+        makeRequest(
+                String.format("api/secure/user/%s", relatedUserFour.id),
+                JsonUtils.toJson(relatedUserFour),
+                getMockHeaders(relatedUserFour),
+                HttpMethod.PUT
```
br648: Nit. Bring this indentation back by one tab.

br648 (approved):
Some minor items, but apart from that all good!

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
```diff
@@ -136,6 +139,31 @@ public static User getUserByEmail(String email, boolean createIfNotExists) {
...
+    public static void updateAuth0Language(OtpUser user, OtpUser preExistingUser) {
+        // Only call API if the user language has changed
+        if (!Objects.equals(user.preferredLocale, preExistingUser.preferredLocale)) {
+            try {
+                Map<String, Object> userMetadata = new HashMap<>();
+                userMetadata.put("lang", user.preferredLocale);
+                // Passing the entire user here will return a 403, because Auth0 sees it as you overwriting
+                // read-only fields, so create a blank user and give it only the data we want to update.
+                User auth0user = new User();
+                auth0user.setUserMetadata(userMetadata);
```
binh-dam-ibigroup: Replace `userMetaData` with `Map.of("lang", user.preferredLocale)`

binh-dam-ibigroup: Auth0 uses slightly different codes for certain languages (e.g. French: `fr` => `fr-FR`, Chinese Simplified/Traditional: `zh-Hans/t` => `zh-CN/TW`), so we will need to map them here, otherwise the messages for some of the languages will revert to English.

binh-dam-ibigroup (changes requested):
Requesting to add mapping between language codes from the UI for Chinese (traditional and simplified), and (France) French to those used by Auth0.

src/main/java/org/opentripplanner/middleware/auth/Auth0Users.java
amy-corson-ibigroup: @binh-dam-ibigroup Thoughts on if this should be a front end change? We pass these same values to Auth0 in the login, and there's a util `convertChineseLanguageCode` in OTP-RR that does this for Chinese languages. We could set when the `preferredLanguage` changes so it's correctly passed to both?

binh-dam-ibigroup: Maybe, so we'd put the Auth0 code in `user.preferredLocale`. It's ok to duplicate here too, because we also have some different mapping for Twilio (because they use specific language codes too).

amy-corson-ibigroup: > Replace `userMetaData` with `Map.of("lang", user.preferredLocale)`
Thanks, fixed in [a954131](https://github.com/ibi-group/otp-middleware/pull/346/commits/a954131009d4f11ac17afe731a5cb512b83f7807)

amy-corson-ibigroup: Added mapping between language codes in [aacffa6](https://github.com/ibi-group/otp-middleware/pull/346/commits/aacffa628cfd0284f3d6455c15bd789ee5eacc2e)

binh-dam-ibigroup (approved):
Nice, thank you for the changes!

PR #365, opened 2026-02-19
Save faretable

binh-dam-ibigroup:
### Checklist

### Description

This PR adds fares to legs so they are persisted in Mongo.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/otp/response/FareProduct.java
+++ b/src/main/java/org/opentripplanner/middleware/otp/response/FareProduct.java
@@ -1,0 +1,23 @@
+package org.opentripplanner.middleware.otp.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import org.bson.codecs.pojo.annotations.BsonIgnore;
+
+import java.util.Collection;
+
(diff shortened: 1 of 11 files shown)


src/main/java/org/opentripplanner/middleware/otp/response/FareProduct.java
```diff
@@ -0,0 +1,20 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class FareProduct implements Cloneable {
+
+    // __typename
+    public /*FeedScopedId*/ String id;
+    public FareMedium medium;
+    public String name;
+    public RiderCategory riderCategory;
+    public Money price;
+    public List<FareDependency> dependencies = new ArrayList<>();
```
JymDyerIBI: I had this as `@JsonIgnore` and `@BsonIgnore` as there may be no dependencies. Is that not necessary?

binh-dam-ibigroup: We can't use `@JsonIgnore` or `@BsonIgnore`because we need to persist `dependencies`, so that we can correctly display transfer pricing. But I updated the field in https://github.com/ibi-group/otp-middleware/pull/365/commits/939c59ab84a1c7b4a6573e52df453a51e60d7653 so that the field is omitted for null values when a not-dependent fare product is returned.

JymDyerIBI (approved):
👍  This looks good.

daniel-heppner-ibigroup (approved):
Looks good to me! Result ended up being quite a nice solution.

PR #376, opened 2026-04-28
Improve Itinerary Checks

binh-dam-ibigroup:
### Checklist

### Description

This PR improves itinerary checks:
- For new monitored trips, where a `/checkitinerary` call is made by a client prior to persisting, the itinerary check result is placed in a cache and consumed when the trip that triggered the check is persisted, with the condition that the client provides the check id in the trip's `itineraryExistence.id` field. This avoids performing the check again upon initial POST. If `itineraryExistence.id` is not provided, the check is performed on POST as before.
- For existing trips, a request to `/checkitinerary` causes the returned `ItineraryExistence` to be also persisted into the existing trip. The PUT operation on an existing trip does not touch existing content in `itineraryExistence`.

Tests are added for the new logic and to solidify the existing logic.

### DIFF
--- a/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
+++ b/src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
@@ -112,14 +112,7 @@ public class MonitoredTripController extends ApiController<MonitoredTrip> {
         // Ensure user has not reached their limit for number of trips.
         verifyBelowMaxNumTrips(monitoredTrip.userId, req);
         preCreateOrUpdateChecks(monitoredTrip, req);
-
-        // FIXME: Pending https://github.com/ibi-group/otp-middleware/pull/219,
-        //   check itinerary existence for recurring trips only for now.
-        //   (Existence should ultimately be checked on all trips.)
-        if (!monitoredTrip.isOneTime()) {
-            checkItineraryExistenceOrReusePriorCheck(monitoredTrip, req);
-        }
-
+        checkItineraryExistenceOrReusePriorCheck(monitoredTrip, req);
         notifyTripCompanionsAndObservers(monitoredTrip, null);
 
         return monitoredTrip;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
```diff
@@ -85,15 +117,35 @@ MonitoredTrip preCreateHook(MonitoredTrip monitoredTrip, Request req) {
...
-            // Check itinerary existence for all days and replace the provided trip's itinerary with a verified,
-            // non-realtime version of it.
-            boolean success = monitoredTrip.checkItineraryExistence(true);
-            if (!success) {
-                logMessageAndHalt(
-                    req,
-                    HttpStatus.BAD_REQUEST_400,
-                    monitoredTrip.itineraryExistence.message
-                );
+            String checkId = monitoredTrip.itineraryExistence != null ? monitoredTrip.itineraryExistence.id : null;
```
br648: I think this block would be cleaner in it's own method. Is the `FIXME` still a thing? The PR has been merged.

br648 (approved):
Just one minor suggestion. Tests all pass. Approving.

src/main/java/org/opentripplanner/middleware/controllers/api/MonitoredTripController.java
binh-dam-ibigroup: I extracted the block in https://github.com/ibi-group/otp-middleware/pull/376/commits/76034a1625d87dd8aba31f9d9a30387f159d2040. Good remark on the TODO, I got rid of it with https://github.com/ibi-group/otp-middleware/pull/376/commits/fc0fe3b5cf0228100bd2d6af1cc75bdfa2ed007f but kept the old behavior to let persist regardless of existence result.

br648 (approved):
Yep, that looks much better. Thanks for updating.

amy-corson-ibigroup (approved):
I think this makes sense! Thank you for the changes

PR #348, opened 2025-10-10
OTP-1890 Fix "Short" Time Formats with AM/PM

JymDyerIBI:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing

### Description

Short time format for locale `en_US` generates a string designating hours and minutes, followed by an ASCII space and then "AM" or "PM". As of JDK 20, this space is replaced with  narrow non-breaking space (NNBSP), so unit tests of strings containing these formatted times break.

An earlier fix used regular expressions for both formats (e.g. `"[\\u202f ]PM"`) but this proved to be cumbersome and new unit tests were introduced with new hardcoded strings that also broke.

This fix introduces a helper method to test formatted strings against hardcoded strings that include a NNBSP, with a backup retry with a string that has a regular ASCII space. Our unit test cases are easier to read and maintain.

PR #380, opened 2026-07-30
Support new JDK time format

binh-dam-ibigroup:
### Checklist

### Description

Supersedes #348.

This PR makes unit tests that involve local US time formatting run in both older and newer JRE versions by applying the correct format to expected time values according to the ambient JRE. In older JREs, with times such as "10:48 AM", a regular space precedes "AM". In newer JREs, that space is actually a narrow non-breaking space (NNBSP - u202f). CI now runs tests with Java 11 and 25.

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -8,6 +8,9 @@ permissions:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        javaVersion: [11, 25]
 
     steps:
       - uses: actions/checkout@v2
@@ -15,10 +18,10 @@ jobs:
         uses: supercharge/mongodb-github-action@1.12.1
         with:
           mongodb-version: 4.2
-      - name: Set up JDK 11
+      - name: Set up JDK ${{ matrix.javaVersion }}
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          java-version: ${{ matrix.javaVersion }}
       - name: Setup Maven Cache
         uses: actions/cache@v4
         id: cache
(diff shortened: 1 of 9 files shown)


.github/workflows/maven-ci.yml
```diff
@@ -98,30 +101,36 @@ jobs:
       # only prepare deploy files on push events to avoid duplicate pushes to s3 when a commit is
       # pushed to a branch with an open pull request
       - name: Prepare files for deployment to s3
-        if: github.event_name == 'push'
+        if: github.event_name == 'push' && matrix.javaVersion == '11'
```
br648, line 106: Why are all these tasks skipped if version 25 is used?

src/test/java/org/opentripplanner/middleware/utils/DateTimeUtilsTest.java
```diff
@@ -100,4 +103,18 @@ void canGetHoursBetween() {
         );
         assertEquals(expectedHours, getHoursBetween(date1, date2));
     }
+
+    @Test
+    void canFormatShortTime() {
```
br648, line 108: This test fails for me using JDK 17. 

```
Expected: a string matching the pattern '\d:06[\u202f ]AM'
     but: was "2:06 PM"
```

br648 (commented):
A couple of items to consider.

.github/workflows/maven-ci.yml
binh-dam-ibigroup: The condition is for uploading the artifacts once only (from the JDK 11 version)

src/test/java/org/opentripplanner/middleware/utils/DateTimeUtilsTest.java
binh-dam-ibigroup: Oops, timezone stuff. Try https://github.com/ibi-group/otp-middleware/pull/380/commits/5caae350d2710b8d5a2e8beea8d2bfbd3d94220c.

br648: Test passes now. Thanks.

br648 (approved):
All good now. Thanks for the update.

src/main/java/org/opentripplanner/middleware/utils/DateTimeUtils.java
```diff
@@ -390,4 +394,11 @@ public static long diffInMinutes(long millis1, long millis2) {
     public static Date dateAsSeconds() {
         return new Date(Instant.now().getEpochSecond());
     }
+
+    /**
+     * Used in tests. Inserts strings such as "5:40 PM" within other text with the correct spacing character.
+     */
+    public static String usTime(int hour, int minute, String message) {
```
daniel-heppner-ibigroup: I think I would like to see this function in the test files so we don't accidentally use the same function in production code and test code (to make sure the tests actually work to detect deviations from expected values)

binh-dam-ibigroup: You are right. Fixed in https://github.com/ibi-group/otp-middleware/pull/380/commits/c376f5e5b38d1e1b90a2d6561aa43a409c1e3a96.

